PR #1, opened 2020-01-06
DO NOT MERGE: Go-based kind deployment

dimaunx:


mpeterson:
As we talked in person the first thing to do is to divide this into multiple patches so it can be reviewed, as ~6300 lines is a bit too much. 

I'll start familiarizing myself with the code though, for when you start sending those new PRs. BTW, make sure that your commit messages and PR messages are descriptive of what they introduce.

pkg/cluster/kubeconfig.go
```diff
@@ -0,0 +1,144 @@
...
+		Name string `yaml:"name"`
+		User struct {
+			ClientCertificateData string `yaml:"client-certificate-data"`
+			ClientKeyData         string `yaml:"client-key-data"`
+		} `yaml:"user"`
+	} `yaml:"users"`
+}
+
+// PrepareKubeConfigs modifies kubconfig file generated by kind and returns the kubeconfig file path
+func PrepareKubeConfigs(clName string, sourceKubeConfigFilePath, masterIP string) error {
```
mpeterson: Why is this necessary? it's not clear to me why the configs that kind generates are not enough. Could you explain and also document it in the doc text?

pkg/cluster/kubeconfig.go
```diff
@@ -0,0 +1,144 @@
...
+		return errors.Wrapf(err, "failed to save kube config %s.", newLocalKubeFilePath)
+	}
+
+	kubeconf.Clusters[0].Cluster.Server = "https://" + masterIP + ":6443"
+	d, err = yaml.Marshal(&kubeconf)
+	if err != nil {
+		return errors.Wrapf(err, "failed to marshal kube config.")
+	}
+
+	err = ioutil.WriteFile(newContainerKubeFilePath, d, 0644)
```
mpeterson: could you elaborate why you need to save the same file twice but with a change on `.Cluster.Server` ?

pkg/cluster/cluster.go
```diff
@@ -0,0 +1,293 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	apiExtClientSet, err := apiextclientset.NewForConfig(kconfig)
+	if err != nil {
+		return nil, err
+	}
+	return apiExtClientSet, nil
+}
```
mpeterson: Practically the same code for both functions, I think you could extract commonalities and make them just wrappers to a clientset getter.

pkg/cluster/cluster.go
```diff
@@ -0,0 +1,293 @@
...
+		err = wait.ForDaemonSetReady(cl.Name, clientSet, "kube-system", "weave-net")
+		if err != nil {
+			return err
+		}
+
+		err = wait.ForDeploymentReady(cl.Name, clientSet, "kube-system", "coredns")
+		if err != nil {
+			return err
+		}
+	}
```
mpeterson: The contents of each switch clause should be moved into functions that are related to cluster creation but each CNI in a different file. You could even do it in a `map[string]function`, remove the switch and allow for registration of CNIs into that `map`. This, together with having a single `--cni` flag instead of multiple would allow for easy extending of support for other CNIs

cmd/armada/create/cluster/createcluster.go
```diff
@@ -0,0 +1,190 @@
...
+			if err != nil {
+				log.Fatal(err)
+			}
+
+			var wg sync.WaitGroup
+			wg.Add(len(targetClusters))
+			for _, cl := range targetClusters {
+				go func(cl *cluster.Config) {
+					err := cluster.Create(cl, provider, box, &wg)
+					if err != nil {
```
mpeterson: Both here and in L94 I think you should abort the processing of other clusters in case of failure. It's not going to be useful to have the rest of the clusters deployed, I think. Do you think differently?

cmd/armada/create/cluster/createcluster.go
```diff
@@ -0,0 +1,190 @@
...
+			wg.Wait()
+			return nil
+		},
+		PersistentPostRun: func(cmd *cobra.Command, args []string) {
+			files, err := ioutil.ReadDir(defaults.KindConfigDir)
+			if err != nil {
+				log.Fatal(err)
+			}
+
+			provider := kind.NewProvider()
```
mpeterson: You are obfuscating the `provider` variable from the function in this closure. Do you want to do that? Shouldn't you already use the `provider` variable from the function?

cmd/armada/create/cluster/createcluster.go
```diff
@@ -0,0 +1,190 @@
...
+					if err != nil {
+						defer wg.Done()
+						log.Fatalf("%s: %s", cl.Name, err)
+					}
+				}(cl)
+			}
+			wg.Wait()
+			return nil
+		},
+		PersistentPostRun: func(cmd *cobra.Command, args []string) {
```
mpeterson: I don't get the purpose of this `PersistentPostRun` could you explain?

cmd/armada/destroy/cluster/destroycluster.go
```diff
@@ -0,0 +1,62 @@
...
+		Long:  "Destroys clusters",
+		RunE: func(cmd *cobra.Command, args []string) error {
+
+			var targetClusters []string
+			if len(flags.Clusters) > 0 {
+				targetClusters = append(targetClusters, flags.Clusters...)
+			} else {
+				configFiles, err := ioutil.ReadDir(defaults.KindConfigDir)
+				if err != nil {
+					log.Fatal(err)
```
mpeterson: You probably want to be a little bit more descriptive than that.

cmd/armada/destroy/cluster/destroycluster.go
```diff
@@ -0,0 +1,62 @@
...
+				}
+				for _, configFile := range configFiles {
+					clName := strings.FieldsFunc(configFile.Name(), func(r rune) bool { return strings.ContainsRune(" -.", r) })[2]
```
mpeterson: This line is a bit cryptic, deserves some comment

cmd/armada/destroy/cluster/destroycluster.go
```diff
@@ -0,0 +1,62 @@
...
+					targetClusters = append(targetClusters, clName)
```
mpeterson: I'm guessing that this is something you do in multiple places and not just here. If that's true then make a function.

cmd/armada/destroy/cluster/destroycluster.go
```diff
@@ -0,0 +1,62 @@
...
+				known, err := cluster.IsKnown(clName, provider)
+				if err != nil {
+					log.Fatalf("%s: %v", clName, err)
+				}
+				if known {
+					err := cluster.Destroy(clName, provider)
+					if err != nil {
+						log.Fatalf("%s: %v", clName, err)
+					}
+				} else {
```
mpeterson: This else is not needed if you do `if !known`

pkg/cluster/cluster.go
```diff
@@ -0,0 +1,293 @@
...
+	usr, err := user.Current()
+	if err != nil {
+		return err
+	}
+
+	_ = os.Remove(filepath.Join(defaults.KindConfigDir, "kind-config-"+clName+".yaml"))
+	_ = os.Remove(filepath.Join(defaults.LocalKubeConfigDir, "kind-config-"+clName))
+	_ = os.Remove(filepath.Join(defaults.ContainerKubeConfigDir, "kind-config-"+clName))
+	_ = os.RemoveAll(filepath.Join(usr.HomeDir, ".kube", strings.Join([]string{"kind-config", clName}, "-")))
+	_ = os.RemoveAll(filepath.Join(defaults.KindLogsDir, clName))
```
mpeterson: It would not be a bad idea to log errors if any.

pkg/cluster/cluster.go
```diff
@@ -0,0 +1,293 @@
...
+}
+
+// Destroy destroys a kind cluster
+func Destroy(clName string, provider *kind.Provider) error {
+	log.Infof("Deleting cluster %q ...\n", clName)
+	if err := provider.Delete(clName, ""); err != nil {
+		return errors.Wrapf(err, "failed to delete cluster %s", clName)
+	}
+
+	usr, err := user.Current()
```
mpeterson: Here you are using the current user home dir but on create you were using current working dir. You need to consolidate to one.

cmd/armada/export/logs/exportlogs.go
```diff
@@ -0,0 +1,57 @@
...
+				targetClusters = append(targetClusters, flags.Clusters...)
+			} else {
+				configFiles, err := ioutil.ReadDir(defaults.KindConfigDir)
+				if err != nil {
+					log.Fatal(err)
+				}
+				for _, configFile := range configFiles {
+					clName := strings.FieldsFunc(configFile.Name(), func(r rune) bool { return strings.ContainsRune(" -.", r) })[2]
+					targetClusters = append(targetClusters, clName)
+				}
```
mpeterson: As expected this logic is duplicated across many places in the code. This is the last comment I put about it but you need to move this into a function.

cmd/armada/export/logs/exportlogs.go
```diff
@@ -0,0 +1,57 @@
...
+	flags := &ExportLogsFlagpole{}
+	cmd := &cobra.Command{
+		Args:  cobra.NoArgs,
+		Use:   "logs",
+		Short: "Export kind cluster logs",
+		Long:  "Export kind cluster logs",
+		RunE: func(cmd *cobra.Command, args []string) error {
+
+			// remove existing before exporting
+			_ = os.RemoveAll(filepath.Join(defaults.KindLogsDir, defaults.KindLogsDir))
```
mpeterson: once again you are not reporting nor handling errors. In this case for sure it needs to handle because as per this line I'm assuming you expect the folder not to exist for the rest of the command,  thus you need to act accordingly.

cmd/armada/export/export.go
```diff
@@ -0,0 +1,18 @@
+package export
+
+import (
+	"github.com/spf13/cobra"
+	"github.com/submariner-io/armada/cmd/armada/export/logs"
+	kind "sigs.k8s.io/kind/pkg/cluster"
+)
+
+// ExportCmd returns a new cobra.Command under root command for armada
+func ExportCmd(provider *kind.Provider) *cobra.Command {
```
mpeterson: I'm not sure about the usefulness of this command. It does not add much to what kind does except executing it for every cluster you pass or have configs for.

I think a much more interesting command that could replace this would be an `execute` or `kind` command that passes the command to kind and executes it for every cluster. You could even import the commands from kind as subcommands to make it more interesting.

mpeterson:
A general comment. There are flags that you repeat a lot across commands. They should be all declared once and added to the commands that make use of it. Too much repetition of code.

pkg/cluster/cluster.go
```diff
@@ -0,0 +1,293 @@
...
+		kind.CreateWithRetain(cl.Retain),
+		kind.CreateWithWaitForReady(cl.WaitForReady),
+		kind.CreateWithDisplayUsage(false),
+		kind.CreateWithDisplaySalutation(false),
+	); err != nil {
+		if errs := kinderrors.Errors(err); errs != nil {
+			for _, problem := range errs {
+				return problem
+			}
+			return errors.New("aborting due to invalid configuration")
```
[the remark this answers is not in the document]
tpantelis: Only is `errs` is empty.

pkg/wait/wait.go
```diff
@@ -0,0 +1,89 @@
...
+			LabelSelector: selector,
+			FieldSelector: "status.phase=Running",
+		})
+		if err == nil && len(podList.Items) == replicas {
+			log.Infof("✔ All pods with label: %q are running for %s.", selector, clName)
+			cancel()
+		} else {
+			log.Debugf("Still waiting for pods. label: %q, namespace: %q, replicas: %v, cluster: %s.", selector, namespace, replicas, clName)
+		}
+	}, 15*time.Second, podsContext.Done())
```
tpantelis: It seems like we should be able to check more frequently then 15 sec, say every 5 sec. Same on L58.

pkg/wait/wait.go
```diff
@@ -0,0 +1,89 @@
...
+	wait.Until(func() {
+		podList, err := c.CoreV1().Pods(namespace).List(metav1.ListOptions{
+			LabelSelector: selector,
+			FieldSelector: "status.phase=Running",
+		})
+		if err == nil && len(podList.Items) == replicas {
+			log.Infof("✔ All pods with label: %q are running for %s.", selector, clName)
+			cancel()
+		} else {
+			log.Debugf("Still waiting for pods. label: %q, namespace: %q, replicas: %v, cluster: %s.", selector, namespace, replicas, clName)
```
tpantelis:
```suggestion
			log.Debugf("Still waiting for pods with label %q, namespace %q, replicas %v in cluster: %q.", selector, namespace, replicas, clName)
```

pkg/wait/wait.go
```diff
@@ -0,0 +1,89 @@
...
+func ForPodsRunning(clName string, c kubernetes.Interface, namespace, selector string, replicas int) error {
+	ctx := context.Background()
+	log.Debugf("Waiting for pods to be running. label: %q, namespace: %q, replicas: %v, duration: %v, *types: %s.", selector, namespace, replicas, defaults.WaitDurationResources, clName)
+	podsContext, cancel := context.WithTimeout(ctx, defaults.WaitDurationResources)
+	wait.Until(func() {
+		podList, err := c.CoreV1().Pods(namespace).List(metav1.ListOptions{
+			LabelSelector: selector,
+			FieldSelector: "status.phase=Running",
+		})
+		if err == nil && len(podList.Items) == replicas {
```
tpantelis: I think we should log the error , ie:

```
if err != nil {
    log.Errorf("Error listing pods for label %q, namespace %q in cluster %q: %v", selector, namespace, clName, err)
    return
}
```

pkg/wait/wait.go
tpantelis: I think this would be useful at Info level to show progress.

pkg/wait/wait.go
```diff
@@ -0,0 +1,89 @@
...
+			log.Infof("✔ All pods with label: %q are running for %s.", selector, clName)
```
tpantelis:
```suggestion
			log.Infof("✔ All pods with label %q in namespace %q are running in cluster %q.", selector, namespace, clName)
```

pkg/wait/wait.go
```diff
@@ -0,0 +1,89 @@
...
+// ForDeploymentReady waits for deployment roll out
+func ForDeploymentReady(clName string, c kubernetes.Interface, namespace, deploymentName string) error {
+	ctx := context.Background()
+	log.Debugf("Waiting up to %v for %s deployment roll out %s ...", defaults.WaitDurationResources, deploymentName, clName)
+	deploymentContext, cancel := context.WithTimeout(ctx, defaults.WaitDurationResources)
+	wait.Until(func() {
+		deployment, err := c.AppsV1().Deployments(namespace).Get(deploymentName, metav1.GetOptions{})
+		if err == nil {
+			if deployment.Status.ReadyReplicas == *deployment.Spec.Replicas {
+				log.Infof("✔ %s successfully deployed to %s, ready replicas: %v", deploymentName, clName, deployment.Status.ReadyReplicas)
```
tpantelis:
```suggestion
				log.Infof("✔ %q successfully deployed in cluster %q with %v ready replicas", deploymentName, clName, deployment.Status.ReadyReplicas)
```

pkg/wait/wait.go
```diff
@@ -0,0 +1,89 @@
...
+				cancel()
+			} else {
+				log.Infof("Still waiting for %s deployment for %s, ready replicas: %v", deploymentName, clName, deployment.Status.ReadyReplicas)
```
tpantelis:
```suggestion
				log.Infof("Still waiting for %q deployment in cluster %q, %v out of %v replicas ready", deploymentName, clName, deployment.Status.ReadyReplicas, *deployment.Spec.Replicas)
```

pkg/wait/wait.go
```diff
@@ -0,0 +1,89 @@
...
+			}
+		} else {
+			log.Debugf("Still waiting for %s deployment roll out %s ...", deploymentName, clName)
```
tpantelis:
```suggestion
			log.Infof("Still waiting for %q deployment roll out in cluster %q", deploymentName, clName)
```

pkg/wait/wait.go
```diff
@@ -0,0 +1,89 @@
...
+	wait.Until(func() {
+		deployment, err := c.AppsV1().Deployments(namespace).Get(deploymentName, metav1.GetOptions{})
+		if err == nil {
+			if deployment.Status.ReadyReplicas == *deployment.Spec.Replicas {
+				log.Infof("✔ %s successfully deployed to %s, ready replicas: %v", deploymentName, clName, deployment.Status.ReadyReplicas)
+				cancel()
+			} else {
+				log.Infof("Still waiting for %s deployment for %s, ready replicas: %v", deploymentName, clName, deployment.Status.ReadyReplicas)
+			}
+		} else {
```
tpantelis: I think we should check the specific error and log if necessary, ie:

```
import "k8s.io/apimachinery/pkg/api/errors"

if err == nil {
    ...
} else if errors.IsNotFound(err) {
    log.Infof(...)
} else {
    log.Errorf("Error getting deployment %q in cluster %q: %v", deploymentName, clName, err)
}
```

pkg/wait/wait.go
```diff
@@ -0,0 +1,89 @@
...
+// ForDaemonSetReady waits for daemon set roll out
+func ForDaemonSetReady(clName string, c kubernetes.Interface, namespace, daemonSetName string) error {
+	ctx := context.Background()
+	log.Debugf("Waiting up to %v for %s daemon set roll out %s ...", defaults.WaitDurationResources, daemonSetName, clName)
+	deploymentContext, cancel := context.WithTimeout(ctx, defaults.WaitDurationResources)
+	wait.Until(func() {
+		daemonSet, err := c.AppsV1().DaemonSets(namespace).Get(daemonSetName, metav1.GetOptions{})
+		if err == nil && daemonSet.Status.CurrentNumberScheduled > 0 {
+			if daemonSet.Status.NumberReady == daemonSet.Status.DesiredNumberScheduled {
+				log.Infof("✔ %s successfully rolled out to %s, ready replicas: %v", daemonSetName, clName, daemonSet.Status.NumberReady)
```
tpantelis:
```suggestion
				log.Infof("✔ Daemon set %q successfully rolled out %v replicas in cluster %q", daemonSetName, clName, daemonSet.Status.NumberReady)
```

pkg/wait/wait.go
```diff
@@ -0,0 +1,89 @@
...
+				cancel()
+			} else {
+				log.Infof("Still waiting for %s daemon set roll out for %s, ready replicas: %v", daemonSetName, clName, daemonSet.Status.NumberReady)
```
tpantelis:
```suggestion
				log.Infof("Still waiting for daemon set %q roll out in cluster %q, %v out of %v replicas ready", daemonSetName, clName, daemonSet.Status.NumberReady, daemonSet.Status.DesiredNumberScheduled)
```

pkg/wait/wait.go
```diff
@@ -0,0 +1,89 @@
...
+			}
+		} else {
+			log.Debugf("Still waiting for %s daemon set roll out %s ...", daemonSetName, clName)
```
tpantelis:
```suggestion
			log.Debugf("Still waiting for daemon set %q roll out in cluster %q", daemonSetName, clName)
```

pkg/wait/wait.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+// ForDaemonSetReady waits for daemon set roll out
+func ForDaemonSetReady(clName string, c kubernetes.Interface, namespace, daemonSetName string) error {
+	ctx := context.Background()
+	log.Debugf("Waiting up to %v for %s daemon set roll out %s ...", defaults.WaitDurationResources, daemonSetName, clName)
+	deploymentContext, cancel := context.WithTimeout(ctx, defaults.WaitDurationResources)
+	wait.Until(func() {
+		daemonSet, err := c.AppsV1().DaemonSets(namespace).Get(daemonSetName, metav1.GetOptions{})
+		if err == nil && daemonSet.Status.CurrentNumberScheduled > 0 {
```
tpantelis: Is it necessary to check `CurrentNumberScheduled`?

pkg/wait/wait.go
```diff
@@ -0,0 +1,89 @@
...
+			if daemonSet.Status.NumberReady == daemonSet.Status.DesiredNumberScheduled {
+				log.Infof("✔ %s successfully rolled out to %s, ready replicas: %v", daemonSetName, clName, daemonSet.Status.NumberReady)
+				cancel()
+			} else {
+				log.Infof("Still waiting for %s daemon set roll out for %s, ready replicas: %v", daemonSetName, clName, daemonSet.Status.NumberReady)
+			}
+		} else {
```
tpantelis: Same as commented in `ForDeploymentReady` re: checking specific error.

pkg/wait/wait.go
tpantelis: Done w/ https://github.com/submariner-io/armada/pull/1

test/e2e/e2e_test.go
```diff
@@ -0,0 +1,445 @@
...
+	"github.com/submariner-io/armada/pkg/image"
+	"github.com/submariner-io/armada/pkg/wait"
+	"sigs.k8s.io/kind/pkg/cluster/nodes"
+
+	dockertypes "github.com/docker/docker/api/types"
+	"github.com/docker/docker/api/types/filters"
+	dockerclient "github.com/docker/docker/client"
+	"github.com/gobuffalo/packr/v2"
+	log "github.com/sirupsen/logrus"
+	createclustercmd "github.com/submariner-io/armada/cmd/armada/create/cluster"
```
mpeterson: I'll put a couple of comments on the structure of the project here since you have the majority of the imports here but applies in general..

1. RE: `armada/cmd/armada/create/cluster` and `armada/cmd/armada/destroy/cluster`
    * A general comment, why repeating `armada` twice (before/after `cmd`)?
    * Why `destroy` and `create` are they in different modules? It makes more sense to have them in a `cluster` module.

2. RE: `armada/pkg/cluster` and `armada/pkg/defaults`
    * Why are they in different modules separate from the `cmd` hierarchy?

3. In general the most of the modules within `cmd` should not there. For example, the `cluster` module should exist independent from `cmd` and should provide the `create` and `deploy` functions and `cmd` then should have all the wrapping to call those functions in the context of a CLI.

4. RE: too specific commands:
    * the `export/logs` command should be an `execute` command instead that passes that execution to each `kind` separately (since anyway internally you are doing just that but calling just one specific `kind` command)
    * the `deploy/nginx` command should instead be an `apply` that is passed to `kind` and thus can be generalized. As a corollary of this, the YAMLs specific to doing these deployments should not be here.

tpantelis: > 1. RE: `armada/cmd/armada/create/cluster` and `armada/cmd/armada/destroy/cluster`
>    
>    * A general comment, why repeating `armada` twice (before/after `cmd`)?
>    * Why `destroy` and `create` are they in different modules? It makes more sense to have them in a `cluster` module.
> 3. In general the most of the modules within `cmd` should not there. For example, the `cluster` module should exist independent from `cmd` and should provide the `create` and `deploy` functions and `cmd` then should have all the wrapping to call those functions in the context of a CLI.

Done with https://github.com/submariner-io/armada/pull/4

pkg/cluster/kubeconfig.go
```diff
@@ -0,0 +1,144 @@
...
+	} `yaml:"users"`
+}
+
+// PrepareKubeConfigs modifies kubconfig file generated by kind and returns the kubeconfig file path
+func PrepareKubeConfigs(clName string, sourceKubeConfigFilePath, masterIP string) error {
+	var kubeconf kubeConfig
+	currentDir, _ := os.Getwd()
+
+	_ = os.MkdirAll(defaults.LocalKubeConfigDir, os.ModePerm)
+	_ = os.MkdirAll(defaults.ContainerKubeConfigDir, os.ModePerm)
```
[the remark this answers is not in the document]
tpantelis: Done w/ https://github.com/submariner-io/armada/pull/5..

pkg/cluster/kubeconfig.go
```diff
@@ -0,0 +1,144 @@
...
+	newContainerKubeFilePath := filepath.Join(currentDir, defaults.ContainerKubeConfigDir, kindKubeFileName)
+
+	sourceKubeFile, err := ioutil.ReadFile(sourceKubeConfigFilePath)
+	if err != nil {
+		return errors.Wrapf(err, "failed to read kube config %s.", sourceKubeConfigFilePath)
+	}
+
+	err = yaml.Unmarshal(sourceKubeFile, &kubeconf)
+	if err != nil {
+		return errors.Wrapf(err, "failed to read kube config %s.", sourceKubeConfigFilePath)
```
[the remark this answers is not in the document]
tpantelis: Done w/ https://github.com/submariner-io/armada/pull/5..

pkg/cluster/cluster.go
tpantelis: `GetCrdClientSet` has been removed

pkg/cluster/cluster.go
```diff
@@ -0,0 +1,293 @@
...
+		kind.CreateWithNodeImage(cl.NodeImageName),
+		kind.CreateWithKubeconfigPath(cl.KubeConfigFilePath),
+		kind.CreateWithRetain(cl.Retain),
+		kind.CreateWithWaitForReady(cl.WaitForReady),
+		kind.CreateWithDisplayUsage(false),
+		kind.CreateWithDisplaySalutation(false),
+	); err != nil {
+		if errs := kinderrors.Errors(err); errs != nil {
+			for _, problem := range errs {
+				return problem
```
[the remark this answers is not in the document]
tpantelis: It's done this way to return the first error if any.

pkg/cluster/cluster.go
```diff
@@ -0,0 +1,293 @@
...
+			}
+			return errors.New("aborting due to invalid configuration")
+		}
+		return errors.Wrap(err, "failed to create cluster")
+	}
```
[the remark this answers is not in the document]
tpantelis: The code was simplified by removing L64-69 and just wrap `err` as is.

mangelajo:
Shall we close this one ? it will always be pull#1 in our hearts ;) , easy to find

PR #6, opened 2020-01-24
Modifications to config & test

tpantelis:
- delete file on error in GenerateKindConfig
- improved error handling
- improved error messages
- restructured config tests, added missing tests, moved GenerateKindConfig
  from cluster_test.go
- other minor changes


pkg/cluster/config_test.go
```diff
@@ -1,244 +1,184 @@
...
+			config.NumWorkers = 5
+		})
+
+		It("should generate the correct kind config file", func() {
+			verifyKindConfigFile(configDir, configPath, "custom_five_workers.golden")
 		})
+	})
+}
+
+func verifyKindConfigFile(configDir string, actualConfigFilePath string, expConfigFileName string) {
```
mkolesnik: nit: why `expConfigFileName` and not `expectedConfigFileName`?

pkg/cluster/config_test.go
```diff
@@ -1,244 +1,184 @@
...
+
+	When("the CNI is flannel", func() {
+		It("should set WaitForReady to 0", func() {
+			config := executePopulateConfig("", "flannel", false)
+			Expect(config.Cni).To(Equal("flannel"))
+			Expect(config.WaitForReady).To(Equal(time.Duration(0)))
+		})
+	})
+
+	When("overlapping IPs are desired", func() {
```
mkolesnik: Whilst the original test checked creating configuration of 2 clusters and verifying the subnets, this new one is more limited in scope and only tests a signle cluster, which leaves more room for error in the tested code.

pkg/cluster/config_test.go
tpantelis: A bit shorter but I'm fine either way.

pkg/cluster/config.go
```diff
@@ -57,55 +57,71 @@ type Config struct {
...
 	if err != nil {
-		return "", err
+		os.Remove(kindConfigFilePath)
+		return "", errors.Wrapf(err, "failed to generated kind config file %q", kindConfigFilePath)
 	}
 
-	if err := f.Close(); err != nil {
-		return "", err
+	if err := kindConfigFile.Close(); err != nil {
+		os.Remove(kindConfigFilePath)
```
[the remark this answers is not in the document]
tpantelis: > Not sure it makes sense to delete the file, from the docs it seems that close will return an error if the file is already closed in which case I'm not sure you'd want to delete it.

I'm not sure I follow. We just successfully created the file on L86 so it wouldn't be already closed when we hit L97. 

> 
> Also, shouldn't we have the close with defer right after the file is opened (at create)?

We could but then we wouldn't be able to return an error if `Close` fails.

pkg/cluster/config_test.go
tpantelis: The original test didn't really add any value that I could see. It externally looped and called `PopulateConfig` but each call is isolated and wouldn't have any side effect to potentially cause the next call to fail.

pkg/cluster/config.go
```diff
@@ -57,55 +57,71 @@ type Config struct {
...
 		Cni:                 cni,
 		NumWorkers:          defaults.NumWorkers,
-		DNSDomain:           defaults.ClusterNameBase + strconv.Itoa(i) + ".local",
+		DNSDomain:           name + ".local",
 		KubeAdminAPIVersion: defaults.KubeAdminAPIVersion,
 		Retain:              retain,
 		Tiller:              tiller,
 		WaitForReady:        wait,
-		KubeConfigFilePath:  filepath.Join(usr.HomeDir, ".kube", strings.Join([]string{"kind-config", defaults.ClusterNameBase + strconv.Itoa(i)}, "-")),
+		KubeConfigFilePath:  filepath.Join(user.HomeDir, ".kube", strings.Join([]string{"kind-config", name}, "-")),
```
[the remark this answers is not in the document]
tpantelis: Done

pkg/cluster/config_test.go
tpantelis: Done

pkg/cluster/config_test.go
tpantelis: That said, I think this would be useful if it tested non-overlapping and verified different CIDRs across 2 clusters. I made that change with https://github.com/submariner-io/armada/pull/6/commits/16b0405e4f5f89ea440502bf997e05f2f737017a.

pkg/cluster/config.go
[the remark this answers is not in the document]
mkolesnik: > 
> > Also, shouldn't we have the close with defer right after the file is opened (at create)?
> 
> We could but then we wouldn't be able to return an error if `Close` fails.

Right, that a good point. We could do it in a way that also catches that error if it happens [1].
Anyway this is not a major point just a thought on how should we be making sure the code is more resilient by using go's defer mechanism. This way you wouldn't need to remove the file manually on L93 since it would be handled by the defer.

[1] https://blog.learngoprogramming.com/5-gotchas-of-defer-in-go-golang-part-iii-36a1ab3d6ef1

pkg/cluster/config_test.go
mkolesnik: Cool, that one's nice :)

pkg/cluster/config.go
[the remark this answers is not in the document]
tpantelis: OK I changed it to use defer and named params. TBH, I'm not really crazy about it but I can live with it.

PR #11, opened 2020-02-06
Moved out cluster names logic to common utils

mkolesnik:
As part of overhauling the create command, I've noticed that the cluster name extraction logic is copied all over the place.
As such I've moved it into a common utils package and reused it where needed

### DIFF
--- a/pkg/utils/cluster_names_test.go
+++ b/pkg/utils/cluster_names_test.go
@@ -21,9 +21,18 @@ func TestUtils(t *testing.T) {
 }
 
 var _ = Describe("Utils tests", func() {
+	Context("ClusterName", testClusterName)
 	Context("ClusterNamesFromFiles", testClusterNamesFromFiles)
 	Context("ClusterNamesOrAll", testClusterNamesOrAll)
 })
+
+func testClusterName() {
+	When("sent a number", func() {
+		It("should return a cluster name", func() {
+			Expect("cluster42").To(Equal(ClusterName(42)))
+		})
+	})
+}
 
 func stubDirectory() {
 	origDir = defaults.KindConfigDir
--- a/pkg/utils/cluster_names_test.go
+++ b/pkg/utils/cluster_names_test.go
@@ -1,0 +1,168 @@
+package utils
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/submariner-io/armada/pkg/defaults"
+)
+
+const stubConfigDir = "utils.test.dir"
+
+var origDir string
+
+func TestUtils(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Utils test suite")
+}
+
+var _ = Describe("Utils tests", func() {
+	Context("ClusterNamesFromFiles", testClusterNamesFromFiles)
+	Context("ClusterNamesOrAll", testClusterNamesOrAll)
+})
+
+func stubDirectory() {
+	origDir = defaults.KindConfigDir
+	defaults.KindConfigDir = stubConfigDir
+}
+
+func restoreDirectory() {
+	defaults.KindConfigDir = origDir
+}
+
+func createStubDirectory() {
+	Expect(os.Mkdir(stubConfigDir, 0755)).To(Succeed())
+}
+
+func deleteStubDirectory() {
+	Expect(os.RemoveAll(stubConfigDir)).To(Succeed())
+}
+
+func generateClusterFiles(expectedClusters []string) {
+	for _, cluster := range expectedClusters {
+		fileName := fmt.Sprintf("kind-config-%s.yaml", cluster)
+		_, err := os.Create(filepath.Join(stubConfigDir, fileName))
+		Expect(err).To(Succeed())
+	}
+}
+
+func testClusterNamesFromFiles() {
+	var (
+		clusters []string
+		err      error
+	)
+
+	BeforeEach(stubDirectory)
+
+	AfterEach(restoreDirectory)
+
+	JustBeforeEach(func() {
+		clusters, err = ClusterNamesFromFiles()
+	})
+
+	When("directory doesn't exist", func() {
+		It("should return an error", func() {
+			Expect(err).To(HaveOccurred())
+		})
+	})
+
+	When("directory exists", func() {
+		BeforeEach(createStubDirectory)
+
+		AfterEach(deleteStubDirectory)
+
+		When("the directory is empty", func() {
+			It("should return an empty slice", func() {
+				Expect(len(clusters)).To(Equal(0))
+			})
+		})
+
+		When("the directory has files", func() {
+			var expectedClusters = []string{"cluster1", "cluster2", "cluster42"}
+			BeforeEach(func() {
+				generateClusterFiles(expectedClusters)
+			})
+
+			It("should return the cluster names", func() {
+				Expect(clusters).To(ConsistOf(expectedClusters))
+			})
+		})
+	})
+}
+
+func testClusterNamesOrAll() {
+	var (
+		clusters     []string
+		sentClusters []string
+	)
+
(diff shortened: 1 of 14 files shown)


pkg/utils/utils.go
```diff
@@ -0,0 +1,42 @@
...
+// In case the slice is empty, it will read the cluster names from the existing kind files.
+// Should the read fail, it will fatally log the error (causing the process to abort).
+func ClusterNamesOrAll(clusters []string) []string {
+	if len(clusters) > 0 {
+		return clusters
+	}
+
+	clusters, err := ClusterNamesFromFiles()
+	if err != nil {
+		log.Fatal(err)
```
tpantelis: I think this should propagate the error for consistency. There's a mix of propagation and use of `log.Fatal` - we should clean that up - I prefer error propagation to the topmost level.

pkg/utils/utils.go
```diff
@@ -0,0 +1,42 @@
...
+		clusters = append(clusters, clusterName)
+	}
+
+	return clusters, nil
+}
+
+// ClusterNamesOrAll will return the cluster names sent to it, if there are any.
+// In case the slice is empty, it will read the cluster names from the existing kind files.
+// Should the read fail, it will fatally log the error (causing the process to abort).
+func ClusterNamesOrAll(clusters []string) []string {
```
tpantelis:
```suggestion
func DetermineClusterNames(clusters []string) []string {
```

pkg/utils/cluster_names_test.go
```diff
@@ -0,0 +1,168 @@
...
+		clusters, err = ClusterNamesFromFiles()
+	})
+
+	When("directory doesn't exist", func() {
+		It("should return an error", func() {
+			Expect(err).To(HaveOccurred())
+		})
+	})
+
+	When("directory exists", func() {
```
tpantelis:
```suggestion
	When("the directory exists", func() {
```

pkg/utils/cluster_names_test.go
```diff
@@ -0,0 +1,168 @@
...
+		BeforeEach(createStubDirectory)
+
+		AfterEach(deleteStubDirectory)
+
+		When("the directory is empty", func() {
```
tpantelis:
```suggestion
		Context("and is empty", func() {
```

This way the output reads "when the directory exists and is empty" which I think is cleaner.

pkg/utils/cluster_names_test.go
```diff
@@ -0,0 +1,168 @@
...
+			It("should return an empty slice", func() {
+				Expect(len(clusters)).To(Equal(0))
+			})
+		})
+
+		When("the directory has files", func() {
```
tpantelis:
```suggestion
		Context("and contains files", func() {
```

pkg/utils/cluster_names_test.go
```diff
@@ -0,0 +1,168 @@
...
+		sentClusters = nil
+	})
+
+	AfterEach(restoreDirectory)
+
+	JustBeforeEach(func() {
+		clusters = ClusterNamesOrAll(sentClusters)
+	})
+
+	When("cluster names are sent", func() {
```
tpantelis:
```suggestion
	When("cluster names are provided", func() {
```

pkg/utils/cluster_names_test.go
```diff
@@ -0,0 +1,168 @@
...
+		BeforeEach(func() {
+			sentClusters = []string{"cluster13", "cluster76"}
+		})
+		It("should return the sent clusters", func() {
```
tpantelis:
```suggestion
		It("should return the provided clusters", func() {
```

pkg/utils/cluster_names_test.go
```diff
@@ -0,0 +1,168 @@
...
+		AfterEach(func() {
+			logFatal = origLogFatal
+		})
+
+		It("should call log fatal", func() {
+			Expect(logFatalCalled).To(Equal(true))
+		})
+	})
+
+	When("directory exists", func() {
```
tpantelis:
```suggestion
	When("the directory exists", func() {
```

pkg/utils/cluster_names_test.go
```diff
@@ -0,0 +1,168 @@
...
+		BeforeEach(createStubDirectory)
+
+		AfterEach(deleteStubDirectory)
+
+		When("the directory is empty", func() {
```
tpantelis:
```suggestion
		Context("and is empty", func() {
```

pkg/utils/cluster_names_test.go
```diff
@@ -0,0 +1,168 @@
...
+			It("should return an empty slice", func() {
+				Expect(len(clusters)).To(Equal(0))
+			})
+		})
+
+		When("the directory has files", func() {
```
tpantelis:
```suggestion
		Context("and contains files", func() {
```

pkg/utils/cluster_names_test.go
```diff
@@ -0,0 +1,168 @@
...
+	When("cluster names are sent", func() {
+		BeforeEach(func() {
+			sentClusters = []string{"cluster13", "cluster76"}
+		})
+		It("should return the sent clusters", func() {
+			Expect(clusters).To(ConsistOf(sentClusters))
+		})
+	})
+
+	When("directory doesn't exist", func() {
```
tpantelis:
```suggestion
	When("the directory doesn't exist", func() {
```

pkg/utils/cluster_names_test.go
```diff
@@ -0,0 +1,168 @@
...
+
+	BeforeEach(stubDirectory)
+
+	AfterEach(restoreDirectory)
+
+	JustBeforeEach(func() {
+		clusters, err = ClusterNamesFromFiles()
+	})
+
+	When("directory doesn't exist", func() {
```
tpantelis:
```suggestion
	When("the directory doesn't exist", func() {
```

pkg/utils/cluster_names_test.go
```diff
@@ -21,10 +21,19 @@ func TestUtils(t *testing.T) {
...
 
 var _ = Describe("Utils tests", func() {
+	Context("ClusterName", testClusterName)
 	Context("ClusterNamesFromFiles", testClusterNamesFromFiles)
 	Context("ClusterNamesOrAll", testClusterNamesOrAll)
 })
 
+func testClusterName() {
+	When("sent a number", func() {
+		It("should return a cluster name", func() {
```
tpantelis: I don't think we need a `When` clause here. How about:

`It("should return the correct cluster name for the given number", ...`

pkg/utils/utils.go
mkolesnik: I opted to keep the current behavior since it's a refactor to reuse code and not to change the behavior.
If you want to remove log.Fatal usage then we can do it in a separate patch that encompasses the entire project.
Personally I don't like the boilerplate-ness that go error handling and propagation forces upon us, hence I'm not as much against the log.Fatal usage because the alternative is a bunch of "if err -> return err" statements all over the place, and this being a CLI it's not that awful to abort it with exit(1).

tpantelis: That's fine to look into that in general for a later patch.

I understand what you're saying about boilerplate but unfortunately that's the way the language designers  decide to handle errors, rather than exceptions.  So I don't think one should necessarily forego  propagation just b/c they don't like how Go implements it. I find it cumbersome too at times but it is what it is.

Either way I think we should be consistent - there's mixed usage of propagation and ` log.Fatal` usage even within he same functions. 

I think embedding `log.Fatal` usage deeper in the code can be detrimental, especially if the function also returns error. There may be cases where a caller can handle an error and proceed w/o exiting  but, by prematurely assuming it's fatal, it doesn't get that chance to decide. So I usually prefer to propagate errors to let the upper layers decide what to do. Also propagating allows intermediate layers to do clean up if they need to.

Another alternative is to use `panic` - this also allows callers to recover if need be but I think it's cleaner to return errors.

mkolesnik: OK let's discuss this as a more general topic and address this in a PR for the whole code base?

PR #12, opened 2020-02-10
Add more wait unit tests

tpantelis:

### DIFF
--- a/pkg/wait/wait_test.go
+++ b/pkg/wait/wait_test.go
@@ -27,15 +27,30 @@ func TestWait(t *testing.T) {
 	RunSpecs(t, "Wait test suite")
 }
 
+const namespace = "test-namespace"
+
+var origWaitDurationResources time.Duration
+var origWaitRetryPeriod time.Duration
+
+var _ = BeforeSuite(func() {
+	origWaitDurationResources = defaults.WaitDurationResources
+	origWaitRetryPeriod = defaults.WaitRetryPeriod
+})
+
+var _ = AfterSuite(func() {
+	defaults.WaitDurationResources = origWaitDurationResources
+	defaults.WaitRetryPeriod = origWaitRetryPeriod
+})
+
 type testClient struct {
 	client.Client
-	returnError error
+	initialError error
 }
 
 func (c *testClient) Get(ctx context.Context, key client.ObjectKey, obj runtime.Object) error {
-	err := c.returnError
+	err := c.initialError
 	if err != nil {
-		c.returnError = nil
+		c.initialError = nil
 		return err
 	}
 
@@ -43,9 +58,9 @@ func (c *testClient) Get(ctx context.Context, key client.ObjectKey, obj runtime.
 }
 
 func (c *testClient) List(ctx context.Context, list runtime.Object, opts ...client.ListOption) error {
-	err := c.returnError
+	err := c.initialError
 	if err != nil {
-		c.returnError = nil
+		c.initialError = nil
 		return err
 	}
 
@@ -65,8 +80,6 @@ var _ = Describe("Wait tests", func() {
 })
 
 func testForPodsRunning() {
-	namespace := "test-namespace"
-
 	var (
 		initialPods   []*corev1.Pod
 		client        *testClient
@@ -89,25 +102,12 @@ func testForPodsRunning() {
 			initObjs = append(initObjs, p)
 		}
 
-		client = &testClient{Client: fake.NewFakeClientWithScheme(scheme.Scheme, initObjs...), returnError: listError}
+		client = &testClient{Client: fake.NewFakeClientWithScheme(scheme.Scheme, initObjs...), initialError: listError}
 
 		waitCh = runAsync(func() error {
 			return wait.ForPodsRunning("east", client, namespace, labelSelector, numReplicas)
 		})
 	})
-
-	newPod := func(name string, phase corev1.PodPhase) *corev1.Pod {
-		return &corev1.Pod{
-			ObjectMeta: metav1.ObjectMeta{
-				Namespace: namespace,
-				Name:      name,
-				Labels:    map[string]string{"app": "test"},
-			},
-			Status: corev1.PodStatus{
-				Phase: phase,
-			},
-		}
-	}
 
 	When("the Pods initially exist", func() {
 		BeforeEach(func() {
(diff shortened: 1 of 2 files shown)


pkg/wait/wait_test.go
```diff
@@ -1,26 +1,338 @@
...
+		}
+
+		client = &testClient{Client: fake.NewFakeClientWithScheme(scheme.Scheme, initObjs...), returnError: listError}
+
+		waitCh = runAsync(func() error {
+			return wait.ForPodsRunning("east", client, namespace, labelSelector, numReplicas)
+		})
+	})
+
+	newPod := func(name string, phase corev1.PodPhase) *corev1.Pod {
```
mkolesnik: Seems like you only use the namespace which is constant, so why not define it as a constant for all the test and declare this function on the outside (next to runAsync for example?)

pkg/wait/wait_test.go
```diff
@@ -1,26 +1,338 @@
...
+
+		It("should return success", func() {
+			Eventually(waitCh, defaults.WaitDurationResources).Should(BeClosed())
+		})
+	})
+
+	When("the Pod never becomes running", func() {
+		BeforeEach(func() {
+			defaults.WaitDurationResources = 300 * time.Millisecond
+			defaults.WaitRetryPeriod = 50 * time.Millisecond
```
mkolesnik: Won't this override these "defaults" for any test that would run after these tests?

pkg/wait/wait_test.go
tpantelis: It's only used within this function hence I scoped it as such. It doesn't really matter to me either way - if you prefer it outside I can move it.

mkolesnik: Whatever you prefer.
Personally I try not to use functions like this in the tests unless necessary, but if you like this style then keep it.

pkg/wait/wait_test.go
mkolesnik: Yeah what I meant is it will have effect even after these tests are done running since you don't reset them to their "original values" and these are globals, so that might interfere with other tests that use these values.
Whether that's good or bad I can't tell, but not sure it's what you intended to ahieve.

pkg/wait/wait_test.go
```diff
@@ -1,26 +1,338 @@
...
+			initialPods = append(initialPods, newPod("test-Pod", corev1.PodPending))
+		})
+
+		It("should timeout and return an error", func() {
+			Eventually(waitCh, defaults.WaitDurationResources*2).Should(Receive())
+		})
+	})
+
+	When("the invalid label selector is provided", func() {
```
mkolesnik: s/the/an/

pkg/wait/wait_test.go
tpantelis: yeah I thought about that but really every other UT that uses these functions will have to set these values appropriately (I already did the same in others). But I can add BeforeSuite/After Suite to be complete.

pkg/wait/wait_test.go
```diff
@@ -1,26 +1,338 @@
...
 )
 
 func TestWait(t *testing.T) {
 	RegisterFailHandler(Fail)
 	RunSpecs(t, "Wait test suite")
 }
 
+type testClient struct {
+	client.Client
+	returnError error
```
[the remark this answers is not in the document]
tpantelis: Done

pkg/wait/wait_test.go
tpantelis: I moved it.

pkg/wait/wait_test.go
tpantelis: Done

pkg/wait/wait_test.go
```diff
@@ -1,26 +1,338 @@
...
+		obj.(*appsv1.DaemonSet).Status.NumberReady = 1
+	}, wait.ForDaemonSetReady)
+}
+
+func testForResourceReady(resourceType string, newResource func() runtime.Object, updateToReady func(runtime.Object),
+	waitForReady func(string, client.Client, string, string) error) {
+
+	var (
+		resource          runtime.Object
+		initialDeployment runtime.Object
```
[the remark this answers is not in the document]
tpantelis: Done

pkg/wait/wait_test.go
```diff
@@ -1,26 +1,338 @@
...
+		client            *testClient
+		waitCh            chan error
+		getError          error
```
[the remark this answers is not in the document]
tpantelis: Done

pkg/wait/wait_test.go
```diff
@@ -1,26 +1,338 @@
...
+			BeforeEach(func() {
+				updateToReady(resource)
+			})
+
+			It("should return success", func() {
+				Eventually(waitCh, defaults.WaitDurationResources).Should(BeClosed())
+			})
+		})
+
+		Context("and are eventually ready", func() {
```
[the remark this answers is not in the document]
tpantelis: Done

pkg/wait/wait_test.go
mkolesnik: Thanks!

pkg/wait/wait_test.go
```diff
@@ -1,26 +1,338 @@
...
+		})
+	})
+}
+
+func testForDeploymentReady() {
+	testForResourceReady("Deployment", func() runtime.Object {
+		replicas := int32(1)
+		return &appsv1.Deployment{
+			ObjectMeta: metav1.ObjectMeta{
+				Namespace: "test-namespace",
```
mkolesnik: nit: you can use the `namespace` const here

pkg/wait/wait_test.go
```diff
@@ -1,26 +1,338 @@
...
+	}, func(obj runtime.Object) {
+		obj.(*appsv1.Deployment).Status.ReadyReplicas = 1
+	}, wait.ForDeploymentReady)
+}
+
+func testForDaemonSetReady() {
+	testForResourceReady("DaemonSet", func() runtime.Object {
+		return &appsv1.DaemonSet{
+			ObjectMeta: metav1.ObjectMeta{
+				Namespace: "test-namespace",
```
mkolesnik: nit: you can use the `namespace` const here

tpantelis: Done

pkg/wait/wait_test.go
tpantelis: Done

PR #14, opened 2020-02-13
One debug flag

mkolesnik:
* Use a global debug flag
* Set debug log level once in tests that use that


tpantelis:
LGTM - just need to manually rebase.

PR #16, opened 2020-02-18
Simplify wait error handling

mkolesnik:
Let me know what you think of this approach


pkg/utils/errors.go
```diff
@@ -0,0 +1,11 @@
+package utils
+
+import "fmt"
+
+func ErrorIfOccurrs(err error, message string, args ...interface{}) error {
```
tpantelis:
```suggestion
func ErrorIfOccurs(err error, message string, args ...interface{}) error {
```

pkg/utils/errors.go
```diff
@@ -0,0 +1,11 @@
...
+	if err != nil {
+		return fmt.Errorf(message, args...)
+	}
+
+	return nil
+}
```
tpantelis: I'd suggest putting this in an `error` package so call sites would read: `error.IfOccurs`.

mkolesnik: So then a new directory under utils I guess?

pkg/utils/errors.go
mkolesnik: Right, I initially named it ErrorIfOccurred which made sense but renamed it to "If occurs" and forgot the double r :)

pkg/utils/errors.go
mkolesnik: Had to call it "errors" otherwise it conflicts with the error builtin type.
So it'll be errors.IfOccurs if that makes sense :)

Another option is errutil or something but I don't think it'll look nicer..

tpantelis:
The call sites should also output the original error as well. There's actually already an `errors. WithMessage` function for this (or `errors. Wrap` if you want the stack trace too).

mkolesnik:
Closed in favor of PR #17 17

PR #19, opened 2020-02-25
Less commands refactor

mkolesnik:
Trying to tidy up the command tree, let me know if you like it or not


cmd/logs/export.go
```diff
@@ -16,12 +16,12 @@ type exportFlagpole struct {
 	clusters []string
 }
 
-// NewExportCommand returns a new cobra.Command under export command for armada
-func NewExportCommand(provider *kind.Provider) *cobra.Command {
+// NewExportLogsCommand returns a new cobra.Command capable of exporting logs
+func NewExportLogsCommand(provider *kind.Provider) *cobra.Command {
```
tpantelis: nit: `NewExportCommand ` to avoid stutter as `logs` is already part of the import path.

PR #31, opened 2020-03-16
Copy framework code from submariner

tpantelis:
...sans the submariner client set creation.


tpantelis:
#30

tpantelis:
> Looks OK but we need to add the code to actually run the e2e example tests

yeah I plan on that and also possibly running other tests. But we need all the kind e2e scripts etc first 😄

PR #32, opened 2020-03-16
Add reusable Framework features from lighthouse

tpantelis:
- Framework.ExecWithOptions function
- Service-related functions
- Deployment-related functions

### DIFF
--- a/test/e2e/framework/deployments.go
+++ b/test/e2e/framework/deployments.go
@@ -6,7 +6,6 @@ import (
 	. "github.com/onsi/gomega"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
-	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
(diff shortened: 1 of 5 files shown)


test/e2e/framework/deployments.go
```diff
@@ -5,6 +5,8 @@ import (
 
 	. "github.com/onsi/gomega"
 	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
mkolesnik: This makes no sense :)

test/e2e/framework/services.go
```diff
@@ -3,6 +3,8 @@ package framework
 import (
 	"fmt"
 
+	"github.com/onsi/ginkgo"
+	corev1 "k8s.io/api/core/v1"
```
mkolesnik: Here as well

tpantelis:
> There's some double imports of core/v1, sometimes as v1 and sometimes as corev1, and I think we need to consistently name it corev1 (since there's also apiv1 imports)

yeah that was unintentional resulting from the copy - thanks for noticing it.

test/e2e/framework/deployments.go
tpantelis: done

test/e2e/framework/services.go
tpantelis: done

PR #33, opened 2020-03-16
Refactor creation of rest Configs and k8s clientsets

tpantelis:
Added a static RestConfigs field to simplify creation of other clientset types for Framework users. This is built during BeforeSuite.
    
Removed the TestContext.ClusterClients field and the redundant Framework.ClusterClients field i lieu of a static KubeClients field. This shortens the referenced name in usage sites.
    
Added a BeforeSuite hook for users to perform their own static initialization (eg create submariner clientsets, perform globalnet check etc)

References: #30

### DIFF
--- a/test/e2e/example/example.go
+++ b/test/e2e/example/example.go
@@ -15,17 +15,19 @@ import (
 )
 
 var _ = Describe("[example] Basic example to demonstrate how to write tests using the framework", func() {
-	f := framework.NewDefaultFramework("basic-example")
+	f := framework.NewFramework("basic-example")
+
 	It("Should be able to list existing nodes on the cluster", func() {
-		testListingNodes(f)
+		testListingNodes()
 	})
+
 	It("Should be able to create a pod using the provided client", func() {
 		testCreatingAPod(f)
 	})
 })
 
-func testListingNodes(f *framework.Framework) {
-	for _, cs := range f.ClusterClients {
+func testListingNodes() {
+	for _, cs := range framework.KubeClients {
 		testListingNodesFromCluster(cs)
 	}
 }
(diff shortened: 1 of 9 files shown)


test/e2e/framework/framework.go
```diff
@@ -87,6 +84,13 @@ type Framework struct {
 	cleanupHandle CleanupActionHandle
 }
 
+var (
+	beforeSuiteFuncs []func()
```
mkolesnik: I guess the function signature might change in the future based on the usage, but right now it looks fine.

tpantelis: The current uses don't need to pass or return anything and I can't imagine a case that would need to.

test/e2e/example/example.go
```diff
@@ -25,7 +25,7 @@ var _ = Describe("[example] Basic example to demonstrate how to write tests usin
 })
 
 func testListingNodes(f *framework.Framework) {
```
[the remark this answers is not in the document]
tpantelis: it is - removed.

mangelajo:
The kubeclients as globals are itching, I understand the benefit of cleaner code though, and we could always roll this back if we needed (concurrent testing?, but that today happens in several processes -ginkgo-, so theoretically should be safe)

tpantelis:
> The kubeclients as globals are itching, I understand the benefit of cleaner code though, and we could always roll this back if we needed (concurrent testing?, but that today happens in several processes -ginkgo-, so theoretically should be safe)

What potential issue do you see? The parameters used to create the rest configs and client sets are static so I think it makes sense to also make the client sets static instead of re-creating them every time for each test. Also this enables the global net check to be static which @sridhargaddam preferred. If you're concerned about concurrency, the creation code is triggered via a `SynchronizedBeforeSuite` function which guarantees it's safely executed only once.

PR #34, opened 2020-03-16
Add TCP connectivity test functions

tpantelis:
...from the submariner project with changes:
    

- wrapped the long list of function params in a ConnectivityTestParams struct
- renamed the NetworkType field in the NetworkPodConfig to Networking and defined a NetworkingType type for the field and the constants for additional clarity
    
References: https://github.com/submariner-io/shipyard/issues/30

### DIFF
--- a/test/e2e/tcp/connectivity.go
+++ b/test/e2e/tcp/connectivity.go
@@ -1,0 +1,116 @@
+package tcp
+
+import (
+	"fmt"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/submariner-io/shipyard/test/e2e/framework"
+)
+
+type ConnectivityTestParams struct {
+	Framework             *framework.Framework
+	UseService            bool
+	Networking            framework.NetworkingType
+	ConnectionTimeout     uint
+	ConnectionAttempts    uint
+	FromCluster           framework.ClusterIndex
+	FromClusterScheduling framework.NetworkPodScheduling
+	ToCluster             framework.ClusterIndex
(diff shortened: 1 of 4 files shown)


test/e2e/tcp/connectivity.go
```diff
@@ -0,0 +1,116 @@
...
+		Expect(listenerPod.TerminationMessage).To(ContainSubstring(connectorPod.Pod.Status.PodIP))
+	}
+
+	// Return the pods in case further verification is needed
+	return listenerPod, connectorPod
+}
+
+func RunNoConnectivityTest(p ConnectivityTestParams) (*framework.NetworkPod, *framework.NetworkPod) {
+	if p.ConnectionTimeout == 0 {
+		p.ConnectionTimeout = 5
```
mangelajo: I guess this is the code we had, but while we are at refactors I don't want us to forget about it :)

Such timeout is too low:
1) we create listener with 5 sec timeout,
2) we go to 2nd cluster, and create connector, with the same timeout

but chances are that the creation time will be much higher than the connection attempts.

This leads to false positives, and we should probably bump this to at least (arbitrary) 30 seconds. I know it will make this test slower, but it will have less unlike false possitives.

We could refine our testing images in the future for better handling of timeouts, etc.

tpantelis: yeah this may be too low but this PR is to simply port the code to shipyard - I'd prefer such a change to be done in a separate PR.

mangelajo: Makes sense, I converted it to an issue: https://github.com/submariner-io/shipyard/issues/40

PR #37, opened 2020-03-18
Dapper base image

mkolesnik:
Copying logic to build and deploy the base dapper image to shipyard

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ git:
 
 jobs:
   include:
-  - env: CMD="make validate"
+  - env: CMD="make validate" DEPLOY=true
 
 install:
   - sudo apt-get install moreutils # make ts available
@@ -21,3 +21,10 @@ services:
 script:
   - set -o pipefail;
     $CMD 2>&1 | ts '[%H:%M:%.S]' -s
+
+deploy:
+  - provider: script
+    script: bash scripts/release
+    on:
+      branch: master
+      condition: $DEPLOY = true
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 
-TARGETS := $(shell ls scripts)
+TARGETS := $(shell ls scripts | grep -v dapper-image)
 
 .dapper:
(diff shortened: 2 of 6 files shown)


Makefile
```diff
@@ -1,5 +1,5 @@
 
-TARGETS := $(shell ls scripts)
+TARGETS := $(shell ls scripts | grep -v dapper-image)
```
mangelajo: In this case you can remove the grep -b dapper-image and L11:L13 right?
```suggestion
TARGETS := $(shell ls scripts)
```

Makefile
```diff
@@ -8,6 +8,12 @@ TARGETS := $(shell ls scripts)
 	@./.dapper.tmp -v
 	@mv .dapper.tmp .dapper
 
+dapper-image: .dapper
+	./.dapper -m bind dapper-image
+
```
mangelajo:
```suggestion
```

.travis.yml
```diff
@@ -10,7 +10,7 @@ git:
 
 jobs:
   include:
-  - env: CMD="make validate"
+  - env: CMD="make validate" DEPLOY=true
```
mangelajo:
```suggestion
  - env: CMD="make validate dapper-image" DEPLOY=true
```

? or do we call the dapper-image part from validate? I can't remember, may be it was part of "ci"

mkolesnik: you're right, I forgot to add it to the Makefile, thanks

Makefile
mkolesnik: I'd keep it for now for consistency with subm since we alrady have a validate target and will also have e2e and maybe some other ones

mkolesnik: Actually makes sense.. I'll try it

skitt: But Make knows how to deal with that on its own, without the `grep -v` and the extra rule — it breaks circular dependencies properly...

mkolesnik: Yeah it seems that it doe, cool, it's gone then :)

PR #41, opened 2020-03-23
Migrate cluster creation & deployment logic to shipyard

mkolesnik:
Extracted the cluster creation and submariner deployment logic to shipyard so it can be reused by consuming projects.


scripts/shared/clusters.sh
```diff
@@ -0,0 +1,116 @@
...
+# Run a local registry to avoid loading images manually to kind
+if registry_running; then
+    echo Local registry $KIND_REGISTRY already running.
+else
+    echo Deploying local registry $KIND_REGISTRY to serve images centrally.
+    docker run -d -p 5000:5000 --restart=always --name $KIND_REGISTRY registry:2
+fi
+
+# This IP is consumed by kind to point the registry mirror correctly to the local registry
+registry_ip="$(docker inspect -f '{{.NetworkSettings.IPAddress}}' "$KIND_REGISTRY")"
```
dfarrell07: We may want to extract this to a function in a different patch.

scripts/shared/deploy.sh
```diff
@@ -0,0 +1,156 @@
...
+set -em
+
+source ${SCRIPTS_DIR}/lib/debug_functions
+source ${SCRIPTS_DIR}/lib/version
+source ${SCRIPTS_DIR}/lib/utils
+
+### Functions ###
+
+function import_images() {
+    docker tag quay.io/submariner/submariner:$VERSION localhost:5000/submariner:local
```
mangelajo: where do we set this $VERSION for the runs in shipyard?

scripts/shared/cleanup.sh
```diff
@@ -0,0 +1,41 @@
+
+#!/usr/bin/env bash
+set -em
+
+source ${SCRIPTS_DIR}/lib/debug_functions
+source ${SCRIPTS_DIR}/lib/utils
```
tpantelis: Shouldn't the path be _${SCRIPTS_DIR}/shared/lib_/?

scripts/shared/clusters.sh
mkolesnik: Thanks, I'm doing this in another PR on top of this with some other cleanup work :)

scripts/shared/deploy.sh
mkolesnik: It's set by importing `version` script. Mind you this code is copied as is from submariner :)

scripts/shared/cleanup.sh
mkolesnik: No since the `shared` library is what we copy into the image. This structuring makes it easier to distinguish the scripts that we share and the local ones that pertain only to this repo.

dfarrell07:
Seems like @mkolesnik addressed all the feedback. Going ahead and merging so we can push on.

PR #49, opened 2020-03-25
Use 'make release' to release

mkolesnik:
To be consistent with the rest of the process, use 'make release' for
the release step in travis.


mangelajo:
As talking over slack, let's merge it to see if we fix the release part and unblock other stuff.

PR #50, opened 2020-03-25
Remove anti deploy check from release

mkolesnik:
This check is already done in travis, no need for it.
Also added QUAY vars to DAPPER_ENV so theyre passed down to the dapper
container.

### DIFF
--- a/Dockerfile.dapper
+++ b/Dockerfile.dapper
@@ -1,6 +1,6 @@
 FROM quay.io/submariner/shipyard-dapper-base
 
-ENV DAPPER_ENV=REPO DAPPER_ENV=TAG \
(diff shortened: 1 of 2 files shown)


Dockerfile.dapper
```diff
@@ -1,6 +1,6 @@
 FROM quay.io/submariner/shipyard-dapper-base
 
-ENV DAPPER_ENV=REPO DAPPER_ENV=TAG \
+ENV DAPPER_ENV="REPO TAG QUAY_USERNAME QUAY_PASSWORD" \
```
mangelajo: also TRAVIS_COMMIT

mangelajo:
```suggestion
ENV DAPPER_ENV="REPO TAG TRAVIS_COMMIT QUAY_USERNAME QUAY_PASSWORD" \
```

mkolesnik: Done!

PR #51, opened 2020-03-25
Validate the tests we ship

skitt:
Since the point of shipyard is to provide a test framework, we should
validate that framework. Doing so reveals a dependency on submariner
which this patch also removes.

### DIFF
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -4,7 +4,7 @@ import (
 	"testing"
 
 	_ "github.com/submariner-io/shipyard/test/e2e/example"
-	_ "github.com/submariner-io/submariner/test/e2e/framework"
+	_ "github.com/submariner-io/shipyard/test/e2e/framework"
 )
 
 func TestE2E(t *testing.T) {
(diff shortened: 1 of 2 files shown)


test/e2e/e2e_test.go
```diff
@@ -4,7 +4,7 @@ import (
 	"testing"
 
 	_ "github.com/submariner-io/shipyard/test/e2e/example"
-	_ "github.com/submariner-io/submariner/test/e2e/framework"
+	_ "github.com/submariner-io/shipyard/test/e2e/framework"
```
tpantelis: This was an oversight - I did not intend any code dependency on submariner.  Can you also remove github.com/submariner-io/submariner from go.mod?

skitt: I thought of that too, but `go.mod` doesn’t mention submariner.

tpantelis: yeah - it was in mine local go.mod - that's wierd.

tpantelis: ahh - it gets added when you build due to this import 😄

PR #52, opened 2020-03-25
Use specific target for release

mkolesnik:
Release expects one or no arguments, so use a specific target for it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@ k8s_version ?= 1.14.6
 globalnet ?= false
 deploytool ?= operator
 
-TARGETS := $(shell ls -p scripts | grep -v /)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -20,6 +20,9 @@ deploy: .dapper dapper-image clusters
 $(TARGETS): .dapper dapper-image
 	./.dapper -m bind $@ --k8s_version $(k8s_version) --globalnet $(globalnet) --deploytool $(deploytool)
 
+release: .dapper dapper-image
```
mangelajo: It doesn't let me comment up for suggestion, but do this:

```bash
TARGETS := $(shell ls -p scripts | grep -v / | grep -v release)
```
otherwise make will complain (just a warning)

PR #54, opened 2020-03-25
Validate shell scripts

skitt:
Since we’re focusing on shell scripts rather than Go code for our
infrastructure, it seems useful to add some validation of shell
scripts using ShellCheck.

This adds scripts/shared/lib/* for now.

### DIFF
--- a/scripts/shared/lib/utils
+++ b/scripts/shared/lib/utils
@@ -1,9 +1,13 @@
-. ${BASH_SOURCE%/*}/source_only
+# shellcheck shell=bash
+# shellcheck source=scripts/shared/lib/source_only
+. "${BASH_SOURCE%/*}"/source_only
 
 ### Constants ###
 
 readonly KIND_REGISTRY=kind-registry
+# shellcheck disable=SC2034 # this variable is used elsewhere
 readonly RESOURCES_DIR=${SCRIPTS_DIR}/resources
+# shellcheck disable=SC2034 # this variable is used elsewhere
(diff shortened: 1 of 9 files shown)


scripts/shared/lib/utils
```diff
@@ -34,15 +38,16 @@ function with_context() {
 # 2nd argument is the command to execute, which will have the $cluster variable set.
 # 3rd argument and so forth get passed to the command.
 function run_parallel() {
-    local clusters=$(eval echo "$1")
-    local cmnd=$2
+    local clusters cmnd
+    cluster=$(eval echo "$1")
```
mkolesnik: Should be clusters

skitt: Thanks, fixed!

skitt:
> However, maybe I'm missing something but I don't see that you added execution of the tool?

See `scripts/validate`, it’s run there (and fails the build if an error creeps in).

scripts/shared/lib/utils
```diff
@@ -70,5 +76,6 @@ function declare_cidrs() {
 }
 
 function declare_kubeconfig() {
-    export KUBECONFIG=$(echo ${KUBECONFIGS_DIR}/kind-config-cluster{1..3} | sed 's/ /:/g')
+    KUBECONFIG=$(echo "${KUBECONFIGS_DIR}"/kind-config-cluster{1..3} | sed 's/ /:/g')
+    export KUBECONFIG
```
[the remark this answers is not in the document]
skitt: Done, thanks!

PR #56, opened 2020-03-26
Share shipyard-controlled Makefile snippets

skitt:
Some recipes are controlled by shipyard, specifically the recipes
implemented by shared scripts; shipyard knows what these scripts’
dependencies are, downstreams shouldn’t have to care about this. This
patch adds a Makefile.inc which should be included by downstreams, and
defines shared recipes there.


mangelajo:
@skitt  Something has gone bananas after the change: 
https://travis-ci.com/github/submariner-io/shipyard/jobs/304812905#L792

But I can't figure out why.

Makefile
```diff
@@ -1,24 +1,8 @@
-k8s_version ?= 1.14.6
-globalnet ?= false
-deploytool ?= operator
+include Makefile.inc
 
 TARGETS := $(shell ls -p scripts | grep -v -e / -e clusters -e deploy)
```
mkolesnik: You should remove clusters and deploy from the exclusion since you deleted the scripts

scripts/deploy
```diff
@@ -1,16 +0,0 @@
...
-source ${SCRIPTS_DIR}/lib/version
-
-function load_image() {
-    docker pull ${1}:latest
-    docker tag ${1}:latest ${1}:$VERSION
-}
-
-load_image quay.io/submariner/submariner
-load_image quay.io/submariner/submariner-route-agent
-load_image quay.io/submariner/submariner-globalnet
```
mkolesnik: You removed this part but it's necessary since shipyard doesn't build these images, that's why the CI fails on this patch.
I'd keep this script or alternatively add this as a "pre_deploy" target which the deploy needs (since you want to share the deploy recipe)

Makefile
```diff
@@ -1,24 +1,8 @@
...
-	@echo Downloading dapper
-	@curl -sL https://releases.rancher.com/dapper/latest/dapper-`uname -s`-`uname -m` > .dapper.tmp
-	@@chmod +x .dapper.tmp
-	@./.dapper.tmp -v
-	@mv .dapper.tmp .dapper
-
-shell:
-	./.dapper -m bind -s
-
-clusters: .dapper dapper-image
```
mkolesnik: Not sure why you removed this, clusters needs the dapper-image rebuild in case something changed in the files, otherwise it'll pull an older image without that changes.

skitt:
> Also - how do you propose this Makefile.inc be consumed by the downstream projects?

Ship it in the base image (I need to take care of that part) and include it in the downstream Makefiles...

Makefile
skitt: But `clusters` doesn’t pull anything based on the Dapper image...

scripts/deploy
skitt: I fixed `deploy.sh` to pull images it can’t find locally instead.

Makefile
skitt: Done!

Makefile
skitt: I thought I’d replied to this already, but I can’t find my reply. `clusters.sh` doesn’t deploy anything based on the Dapper image, does it? That’s why I removed the dependency... (That, and the fact that `Makefile.inc` should be self-contained.)

skitt: Fixed, thanks.

package/Dockerfile.dapper-base
```diff
@@ -55,3 +55,4 @@ RUN dnf -y distrosync --nodocs --setopt=install_weak_deps=False && \
 # Copy CI deployment scripts into image to share with all submariner-io/* projects
 WORKDIR $SCRIPTS_DIR
 COPY scripts/shared/ .
+COPY Makefile.inc ..
```
mkolesnik:
```suggestion
COPY Makefile.inc ${SHIPYARD_DIR}
```

Would be more robust :)

Makefile.inc
```diff
@@ -0,0 +1,25 @@
...
+	./.dapper -m bind -s
+
+cleanup: .dapper
+	./.dapper -m bind $(SCRIPTS_DIR)/cleanup.sh
+
+clusters: .dapper
+	./.dapper -m bind $(SCRIPTS_DIR)/clusters.sh --k8s_version $(k8s_version) --globalnet $(globalnet) --registry_inmemory $(registry_inmemory)
+
+deploy: .dapper clusters
+	DAPPER_ENV="OPERATOR_IMAGE" ./.dapper -m bind $(SCRIPTS_DIR)/deploy.sh --globalnet $(globalnet) --deploytool $(deploytool)
```
mkolesnik:
```suggestion
	./.dapper -m bind $(SCRIPTS_DIR)/deploy.sh --globalnet $(globalnet) --deploytool $(deploytool)
```

I know it's not entirely related, but since you're moving it lets move it clean :)

mangelajo: +1..

PR #61, opened 2020-03-30
Support globalnet deployment via Operator in shipyard

sridhargaddam:
This PR enables deploying Globalnet in Shipyard.

### DIFF
--- a/scripts/shared/lib/deploy_operator
+++ b/scripts/shared/lib/deploy_operator
@@ -33,10 +33,12 @@ function deploytool_prereqs() {
 }
 
 function setup_broker() {
+    local gn
+    [[ $globalnet != true ]] || gn="--globalnet"
     echo "Installing broker..."
     (
         cd "${OUTPUT_DIR}" && \
-        subctl --kubeconfig "${KUBECONFIGS_DIR}/kind-config-$cluster" deploy-broker --no-dataplane
+        subctl --kubeconfig "${KUBECONFIGS_DIR}/kind-config-$cluster" deploy-broker --no-dataplane ${gn}
     )
 }
 
@@ -53,7 +55,20 @@ function subctl_install_subm() {
 }
 
 function install_subm_all_clusters() {
-    run_parallel "{1..3}" subctl_install_subm
+    if [[ $globalnet = "true" ]]; then
+        # When Globalnet is enabled, each cluster gets a non-overlapping globalCIDR from
+        # globalnet-cidr-range and this globalCIDR is stored in the clusterCRD for each
+        # cluster. When globalnet is deployed via subctl/Operator, during join operation,
+        # subctl first queries the existing clusters that are associated with the Broker
+        # and allocates a non-overlapping chunk of globalCIDR to the newly joining cluster.
+        # Due to this, we cannot install submariner globalnet (via subctl) in parallel and
+        # it has to be done sequentially.
+        run_parallel "1" subctl_install_subm
+        run_parallel "2" subctl_install_subm
+        run_parallel "3" subctl_install_subm
+    else
+        run_parallel "{1..3}" subctl_install_subm
+    fi
 }
 
 function deploytool_postreqs() {
--- a/scripts/shared/lib/deploy_operator
+++ b/scripts/shared/lib/deploy_operator
@@ -63,9 +63,7 @@ function install_subm_all_clusters() {
         # and allocates a non-overlapping chunk of globalCIDR to the newly joining cluster.
         # Due to this, we cannot install submariner globalnet (via subctl) in parallel and
         # it has to be done sequentially.
-        run_parallel "1" subctl_install_subm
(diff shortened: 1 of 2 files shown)


scripts/shared/lib/deploy_operator
```diff
@@ -33,10 +33,12 @@ function deploytool_prereqs() {
 }
 
 function setup_broker() {
+    gn=
```
mkolesnik:
```suggestion
    local gn
```

scripts/shared/lib/deploy_operator
```diff
@@ -33,10 +33,12 @@ function deploytool_prereqs() {
...
+    [[ $globalnet = true ]] && gn=--globalnet
```
mkolesnik:
```suggestion
    [[ $globalnet != true ]] || gn="--globalnet"
```

logonos: This is very important, in order to avoid exiting bash scripts (that has set -e) when **evaluating false conditions**. 
We should replace all these conditions: `[[ condition ]] && command`
With:   `[[ ! condition ]] || command`

manosnoam: This is very important, in order to avoid exiting bash scripts (that has set -e) when **evaluating false conditions**.
We should replace all these conditions: `[[ condition ]] && command`
With: `[[ ! condition ]] || command`

scripts/shared/lib/deploy_operator
sridhargaddam: Done

scripts/shared/lib/deploy_operator
sridhargaddam: Done

scripts/shared/lib/deploy_operator
```diff
@@ -53,7 +55,20 @@ function subctl_install_subm() {
...
+        # When Globalnet is enabled, each cluster gets a non-overlapping globalCIDR from
+        # globalnet-cidr-range and this globalCIDR is stored in the clusterCRD for each
+        # cluster. When globalnet is deployed via subctl/Operator, during join operation,
+        # subctl first queries the existing clusters that are associated with the Broker
+        # and allocates a non-overlapping chunk of globalCIDR to the newly joining cluster.
+        # Due to this, we cannot install submariner globalnet (via subctl) in parallel and
+        # it has to be done sequentially.
+        run_parallel "1" subctl_install_subm
+        run_parallel "2" subctl_install_subm
+        run_parallel "3" subctl_install_subm
```
mkolesnik, line 68: You should use `with_context` command here in this case, also I'd stick it in a `for i in {1..3}`.
Alternatively you could add a `run_sequential` command for this in the utils if you wish.

sridhargaddam: Okay, I'll add a new api - run_sequential in utils.
Also, with_context is used inside the run_parallel/sequential api.

scripts/shared/lib/utils
```diff
@@ -52,6 +52,19 @@ function run_parallel() {
...
+# Run cluster commands sequentially.
+# 1st argument is the numbers of the clusters to run for, supports "1 2 3" or "{1..3}" for range
+# 2nd argument is the command to execute, which will have the $cluster variable set.
+# 3rd argument and so forth get passed to the command.
+function run_sequential() {
+    local clusters cmnd
+    clusters=$(eval echo "$1")
+    cmnd=$2
+    for i in ${clusters}; do
+        ( with_context "cluster${i}" "$cmnd" "${@:3}" | sed "s/^/[cluster${i}] /" )
```
mkolesnik, line 64:
```suggestion
        with_context "cluster${i}" "$cmnd" "${@:3}" | sed "s/^/[cluster${i}] /"
```

Subshell isn't really needed here, it was more important for the parallel execution

sridhargaddam: Ack, thanks @mkolesnik

scripts/shared/lib/deploy_operator
```diff
@@ -53,7 +55,18 @@ function subctl_install_subm() {
 }
 
 function install_subm_all_clusters() {
-    run_parallel "{1..3}" subctl_install_subm
+    if [[ $globalnet = "true" ]]; then
+        # When Globalnet is enabled, each cluster gets a non-overlapping globalCIDR from
```
vthapar, line 59: If you *really* want to, can optimize this by starting parallel run on second once first one's global cluster is allocated., so on and so forth. Something to consider if we find deployment times to be too high.

mkolesnik: That would be rather hard to find out since the allocation of global CIDR is part of the subctl workflow.
What would be possible is to either fix subctl (or whatever does the allocation) to handle parallel joins (which makes sense anyhow), or add a pair of commands `allocate/release-global-cidr` which could be run serially and a flag to set the global cidr on join.

PR #62, opened 2020-04-01
Make Makefile.inc really shareable

skitt:
To allow sharing Makefile.inc, we need to rework the Dapper
integration. If we detect a Dapper environment (based on the existence
of a non-empty DAPPER_HOST_ARCH environment variable), we define the
available targets, without caring about Dapper. Otherwise, we define
targets to set Dapper up, and run everything through Dapper.

Downstreams can use this by copying the Makefile template, removing
the Shipyard-specific lines marked by comments, and adding their own
project-specific targets in the “Dapper is running” section.

A nice side-effect of this is that targets with dependencies are
handled entirely within Dapper, saving quite a lot of time in complex
builds.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,12 +1,34 @@
-include Makefile.inc
+ifneq (,$(DAPPER_HOST_ARCH))
+
+# Running in Dapper
+
+include $(SHIPYARD_DIR)/Makefile.inc
 
 TARGETS := $(shell ls -p scripts | grep -v -e /)
 
-$(TARGETS): .dapper dapper-image
-	./.dapper -m bind $@
+# Add any project-specific arguments here
+$(TARGETS):
+	./scripts/$@
 
-# We need the latest image for these targets in shipyard
+.PHONY: $(TARGETS)
+
+# Project-specific targets go here
+
+else
+
+# Not running in Dapper
+
+# Shipyard-specific starts
 clusters: dapper-image
 deploy: dapper-image
 
-.PHONY: $(TARGETS)
+dapper-image:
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -1,14 +1,43 @@
+ifneq (,$(DAPPER_HOST_ARCH))
+
+# Running in Dapper
+
 include Makefile.inc
```
skitt, line 5: As pointed out by @mkolesnik, this should use `$(SHIPYARD_DIR)`.

Makefile
```diff
@@ -1,14 +1,43 @@
...
+	@mv .dapper.tmp .dapper
 
-.PHONY: $(TARGETS)
+%: .dapper dapper-image
+	./.dapper -m bind $(MAKE) $@
+
+shell: .dapper
+	./.dapper -m bind -s
+
+.PHONY: dapper-image shell
```
skitt, line 38: As pointed out by @mkolesnik, this should be pulled out to a separate, included Makefile, to make the overall main Makefile easier to read.

Makefile.dapper
```diff
@@ -0,0 +1,19 @@
...
+
+.dapper:
+	@echo Downloading dapper
+	@curl -sL https://releases.rancher.com/dapper/latest/dapper-`uname -s`-`uname -m` > .dapper.tmp
+	@@chmod +x .dapper.tmp
+	@./.dapper.tmp -v
+	@mv .dapper.tmp .dapper
+
+%: .dapper
+	./.dapper -m bind $(MAKE) $@
```
mangelajo, line 14: Make -> dapper -> make :-oooo :-) ;) nice trick

PR #65, opened 2020-04-01
Use submariner-operator namespace even with helm deployments

sridhargaddam:
Currently, when Submariner is deployed via Helm, its installed
in submariner namespace, whereas when its deployed via Operator,
its installed in submariner-operator. It would be helpful to use
the same namespace irrespective of the type of deployment so that
our docs/test-plans can refer to the common namespace.


tpantelis:
> :+1 to this :)

yeah - I never understood why we used separate namespaces in the first place 😄

PR #68, opened 2020-04-02
Genericize test_connection further

mkolesnik:
Receive the source pod name as well as the target address. In addition,
the variable names and the message are now more generic (in case it's
not netshoot & nginx).

### DIFF
--- a/scripts/shared/lib/deploy_funcs
+++ b/scripts/shared/lib/deploy_funcs
@@ -48,16 +48,23 @@ function get_svc_ip() {
 }
 
 function test_connection() {
-    local nginx_address=$1
-    local netshoot_pod
-    netshoot_pod=$(kubectl get pods -l app=netshoot | awk 'FNR == 2 {print $1}')
+    local source_pod=$1
+    local target_address=$2
 
-    echo "Attempting connectivity between clusters - $netshoot_pod cluster2 --> $nginx_address nginx service cluster3"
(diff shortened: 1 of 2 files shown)


scripts/shared/lib/deploy_funcs
```diff
@@ -48,18 +48,25 @@ function get_svc_ip() {
...
-    local nginx_address=$1
-    local netshoot_pod
-    netshoot_pod=$(kubectl get pods -l app=netshoot | awk 'FNR == 2 {print $1}')
+    local source_pod=$1
+    local target_address=$2
 
-    echo "Attempting connectivity between clusters - $netshoot_pod cluster2 --> $nginx_address nginx service cluster3"
-    if ! kubectl exec "${netshoot_pod}" -- curl --output /dev/null -m 30 --silent --head --fail "${nginx_address}"; then
+    echo "Attempting connectivity between clusters - $source_pod (cluster2) --> $target_address (service on cluster3)"
+    if ! kubectl exec "${source_pod}" -- curl --output /dev/null -m 30 --silent --head --fail "${target_address}"; then
```
mangelajo, line 55: I'd take the chance to add --context cluster2 to the exec. So we don't depend on which cluster the caller was... does that make sense?

mkolesnik: It's already run `with_context` so the caller is responsible for the source cluster. The function itself should be agnostic to this (or it could receive it in params, but for this function it's rather cumbersome)

mangelajo: Makes sense, I didn't see the callers :D so I was guessing about that.

PR #70, opened 2020-04-02
Add ability to specify nodes dynamically

mkolesnik:
Added an ability to specify for each cluster the types of nodes it
should have.
Downstream projects (or just devs) can use this to vary the deployment according to their needs.


mkolesnik:
> Very nice, we need to add it to documentation somewhere :)

There is some in the change but I agree, I'll add some shipyard docs once it's more stable.

PR #73, opened 2020-04-02
Share vendor handling

skitt:
For Go-based projects with a go.mod file, this defines a
modules/vendor.txt target which can be used as a dependency to ensure
that the vendor directory is up-to-date.


Makefile.inc
```diff
@@ -13,3 +13,9 @@ clusters:
 
 deploy: clusters
 	$(SCRIPTS_DIR)/deploy.sh --globalnet $(globalnet) --deploytool $(deploytool) $(DEPLOY_ARGS)
+
+ifeq (go.mod,$(wildcard go.mod))
```
mkolesnik, line 17: Why wildcard here? It's only for expansion of patterns AFAICT whilst `go.mod` isn't one.
Also what's the point of this comparison?

skitt: It’s an existence check: `$(wildcard go.mod)` expands to `go.mod` if the file exists, empty otherwise.

mkolesnik: ah i see a bit confusing but cool :)
maybe a comment would've been nice here ;)

skitt: Done!

PR #75, opened 2020-04-02
Share entry

skitt:
The entry script is a technical detail of using Dapper, so it should
be provided by shipyard and not re-defined in other projects. This
moves the script so that it’s shipped in the base image, and updates
the Dapper image to use it.

It also makes the default Dapper target "sh" rather than "ci" which is
no longer available.


skitt:
This fails because `vendor` isn’t populated, so we need #73 first.

PR #76, opened 2020-04-03
WIP: Ensure we rebuild the base Dapper image if needed

skitt:
This checks the build timestamp of the base Dapper image against files
we know we share, and rebuilds it if necessary.


tpantelis:
Is this still relevant?

skitt:
> Is this still relevant?

I’m working on an update of this which initialises the tracker files based on the images’ build timestamp.

PR #77, opened 2020-04-03
Allow not installing submariner

mkolesnik:
Some projects (including shipyard) don't need submariner installed on cluster1, hence added an option to specify whether it should be installed or not.

Also added the ability to override default so that shipyard can provide sensible default for downstream projects, while running with a smaller footprint for itself.

### DIFF
--- a/Makefile.inc
+++ b/Makefile.inc
@@ -4,12 +4,14 @@ deploytool ?= operator
 registry_inmemory ?= true
 
 SCRIPTS_DIR ?= /opt/shipyard/scripts
+CLUSTERS_ARGS ?= ''
+DEPLOY_ARGS ?= ''
 
 cleanup:
 	$(SCRIPTS_DIR)/cleanup.sh
 
 clusters:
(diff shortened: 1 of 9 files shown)


Makefile.inc
```diff
@@ -4,12 +4,14 @@ deploytool ?= operator
 registry_inmemory ?= true
 
 SCRIPTS_DIR ?= /opt/shipyard/scripts
+CLUSTERS_ARGS ?= ''
+DEPLOY_ARGS ?= ''
```
skitt, line 8: These aren’t needed, although I suppose they play some use in documenting the variables (but we should really document the build template separately...)

mkolesnik: OK I thought its best to declare them rather than use potentially undeclared variables in the calls, but I get your point.
Would you rather I remove them or add comments above them that indicate their use?

skitt: I don’t like clutter, I’d vote for removing them ;-).

PR #83, opened 2020-04-06
Split base image into two parts

mkolesnik:
Since most of the time we're just changing the scripts and no the entire
base image, it's a shame to have to rebuild it.
Instead, split the base image into a `shipyard-dapper-base-deps` image
which contains the OS and all the downloaded & installed tools.

Each image would be rebuilt only when necessary.


mkolesnik:
Thanks to a suggestion from @mangelajo closing this in favor of #84

PR #87, opened 2020-04-06
Release properly

mkolesnik:
* Fixed the release script to only release the image we meant to
* Shared the release script so that downstream projects can reuse it


.travis.yml
```diff
@@ -31,7 +31,7 @@ after_success:
 
 deploy:
   - provider: script
-    script: make release
+    script: make release images="shipyard-dapper-base"
```
mangelajo, line 34: we need to pass down image_tag=$TRAVIS_TAG or something like that I think
https://github.com/submariner-io/submariner/blob/master/.travis.yml#L58

PR #95, opened 2020-04-15
Add functions to force pod reloads from new images

mangelajo:
This can be used in the reload-images make target of any
shipyard consumers.

### DIFF
--- a/scripts/shared/lib/deploy_funcs
+++ b/scripts/shared/lib/deploy_funcs
@@ -122,3 +122,23 @@ function load_deploytool() {
     . "$deploy_lib"
 }
 
+function find_submariner_namespace() {
+    local namespace
+    namespace="$(kubectl get pods --all-namespaces | awk '/submariner/{ print $1 }' | grep -v broker | head -n 1)"
+    if [[ "${namespace}" == "" ]]; then
+        echo "Could not find a submariner deployment namespace" >&2
+        exit 1
+    fi
+    echo "${namespace}"
+}
+
+function reload_pods() {
+    local resource_type=$1 # the resource type can be deployment or daemonset
+    local resource_name=$2 # the name of the resource
+    local namespace
+    namespace="$(find_submariner_namespace)"
+
+    kubectl patch -n "${namespace}" "${resource_type}" "${resource_name}" \
+              --type='json' \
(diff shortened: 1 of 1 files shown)


mangelajo:
Fixed some shellcheck errors.

scripts/shared/lib/deploy_funcs
```diff
@@ -122,3 +122,17 @@ function load_deploytool() {
     . "$deploy_lib"
 }
 
+function find_submariner_namespace() {
+    kubectl get pods -A | awk '/submariner/{ print $1 }' | grep -v broker | head -n 1
```
sridhargaddam, line 126: In my local kubectl, I see that its not recognizing the shortcut "-A". Can you change it to "--all-namespaces"?

[sgaddam@localhost submariner]$ kubectl get pods -A | awk '/submariner/{ print $1 }' | grep -v broker | head -n 1
Error: unknown shorthand flag: 'A' in -A

While we are at it, instead of querying for pods in all namespaces, do you think its better to query for daemonSets (less output to parse)
kubectl get ds --all-namespaces | awk '/submariner/{ print $1 }' | grep -v broker | head -n 1

scripts/shared/lib/deploy_funcs
```diff
@@ -122,3 +122,17 @@ function load_deploytool() {
...
+}
+
+function reload_pods() {
+    local resource_type=$1 # the resource type can be deployment or daemonset
+    local resource_name=$2 # the name of the resource
+    local namespace
+    namespace="$(find_submariner_namespace)"
```
sridhargaddam, line 133: If the reload_pods is run on a non-submariner deployment, the error message might be hard to read.
Can you add a check to ensure that namespace is not empty and log an appropriate error message.

scripts/shared/lib/deploy_funcs
mangelajo: Better to query for pods, I'm just trying to figure out the namespace and I was trying to support both "deployments" and "daemonsets", 

since in some cases we will have just deployments.

scripts/shared/lib/deploy_funcs
mangelajo: This can only be used in our kind deployments mostly, but I'll add it just in case.

mangelajo: Done :-)

mangelajo:
> LGTM

Thanks! :)

PR #101, opened 2020-04-16
Only compress files that are installed

skitt:
To avoid failures in the future from changes to the Fedora packages we
use, this patch changes our upx invocation:

* one invocation compresses files we install explicitly
* another lists files from packages we install, and compresses those

In the second phase, missing packages (from changes to dependencies)
will only result in a “package ... is not installed” message on
standard output, which will be filtered by the grep. We ignore errors
because this approach results in asking upx to compress symlinks,
which it complains about (but everything works anyway).

### DIFF
--- a/package/Dockerfile.dapper-base
+++ b/package/Dockerfile.dapper-base
@@ -51,7 +51,7 @@ RUN dnf -y distrosync --nodocs --setopt=install_weak_deps=False && \
     curl -Lo /usr/bin/kind "https://github.com/kubernetes-sigs/kind/releases/download/${KIND_VERSION}/kind-linux-${ARCH}" && chmod a+x /usr/bin/kind && \
     curl -Lo /go/bin/subctl "https://github.com/submariner-io/submariner-operator/releases/download/${SUBCTL_VERSION}/subctl-${SUBCTL_VERSION}-linux-${ARCH}" && chmod a+x /go/bin/subctl && \
     GOFLAGS="" go get -v github.com/onsi/ginkgo/ginkgo && \
-    upx /go/bin/{ginkgo,golangci-lint} /usr/bin/{containerd*,ctr,docker*,goimports,helm,kind,kube*,runc} /usr/lib/golang/pkg/tool/*/* && \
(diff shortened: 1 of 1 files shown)


package/Dockerfile.dapper-base
```diff
@@ -51,8 +51,9 @@ RUN dnf -y distrosync --nodocs --setopt=install_weak_deps=False && \
     curl -Lo /usr/bin/kind "https://github.com/kubernetes-sigs/kind/releases/download/${KIND_VERSION}/kind-linux-${ARCH}" && chmod a+x /usr/bin/kind && \
     curl -Lo /go/bin/subctl "https://github.com/submariner-io/submariner-operator/releases/download/${SUBCTL_VERSION}/subctl-${SUBCTL_VERSION}-linux-${ARCH}" && chmod a+x /go/bin/subctl && \
     GOFLAGS="" go get -v github.com/onsi/ginkgo/ginkgo && \
-    GOFLAGS="" go get -v golang.org/x/tools/cmd/goimports && \
-    upx /go/bin/{ginkgo,golangci-lint} /usr/bin/{containerd*,ctr,docker*,goimports,helm,kind,kube*,runc} /usr/lib/golang/pkg/tool/*/* && \
+    upx /go/bin/{ginkgo,golangci-lint} /usr/bin/{helm,kind,kubefedctl} && \
+    (rpm -ql containerd golang-bin golang-x-tools-goimports kubernetes-client moby-engine runc | \
+     grep -E '(/usr/bin|/usr/lib/golang/bin/|/usr/lib/golang/pkg/tool/.*/)' | xargs -r upx || :) && \
```
mkolesnik, line 56:
```suggestion
    find /usr/bin /go/bin /usr/lib/golang -type f -executable -size +5M | grep -v -e subctl | xargs upx && \
```

Might I suggest a simpler alternative? ;)

skitt: Indeed, simple is better, I’ve applied that (minus the `grep`).

mkolesnik: I thought subctl is compressed already, hence the exclusion..

PR #102, opened 2020-04-16
Add the cableDriver  option support

roytman:
depends on https://github.com/submariner-io/submariner-charts/pull/18

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,8 +12,8 @@ jobs:
   include:
   - env: CMD="make validate"
   - env: CMD="make deploy deploytool=helm globalnet=true"
-  - env: CMD="make deploy deploytool=operator cable_driver=wireguard"
   - env: CMD="make deploy deploytool=operator" DEPLOY=true
+  - env: CMD="make deploy cable_driver=wireguard"
 
 install:
   - sudo apt-get install curl moreutils # make ts available
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,11 +12,16 @@ jobs:
   include:
   - env: CMD="make validate"
   - env: CMD="make deploy deploytool=helm globalnet=true"
+  - env: CMD="make deploy deploytool=operator cable_driver=wireguard"
   - env: CMD="make deploy deploytool=operator" DEPLOY=true
 
 install:
   - sudo apt-get install curl moreutils # make ts available
   - curl https://codeload.github.com/kward/shflags/tar.gz/v1.2.3 | tar -xzf - && sudo mv shflags-1.2.3 /usr/share/shflags
+  - sudo add-apt-repository -y ppa:wireguard/wireguard # add WireGuard support
+  - sudo apt-get update
+  - sudo apt-get install linux-headers-`uname -r` wireguard -y
+  - sudo modprobe wireguard
 
 services:
   - docker
--- a/scripts/shared/deploy.sh
+++ b/scripts/shared/deploy.sh
@@ -16,11 +16,7 @@ deploytool="${FLAGS_deploytool}"
 cluster_settings="${FLAGS_cluster_settings}"
 cable_driver="${FLAGS_cable_driver}"
 
-if [[ -z "${cable_driver}" ]]; then
-  echo "Running with: globalnet=${globalnet}, deploytool=${deploytool}, cluster_settings=${cluster_settings}"
-else
-  echo "Running with: globalnet=${globalnet}, deploytool=${deploytool}, cluster_settings=${cluster_settings}, cable_driver=${cable_driver}"
-fi
+echo "Running with: globalnet=${globalnet}, deploytool=${deploytool}, cluster_settings=${cluster_settings}, cable_driver=${cable_driver}"
 
 set -em
 
--- a/scripts/shared/deploy.sh
+++ b/scripts/shared/deploy.sh
@@ -6,13 +6,21 @@ source /usr/share/shflags/shflags
 DEFINE_string 'cluster_settings' '' "Settings file to customize cluster deployments"
 DEFINE_string 'deploytool' 'operator' 'Tool to use for deploying (operator/helm)'
 DEFINE_string 'globalnet' 'false' "Deploy with operlapping CIDRs (set to 'true' to enable)"
+DEFINE_string 'cable_driver' '' "Cable driver implementation"
+
 FLAGS "$@" || exit $?
 eval set -- "${FLAGS_ARGV}"
 
 globalnet="${FLAGS_globalnet}"
 deploytool="${FLAGS_deploytool}"
 cluster_settings="${FLAGS_cluster_settings}"
-echo "Running with: globalnet=${globalnet}, deploytool=${deploytool}, cluster_settings=${cluster_settings}"
+cable_driver="${FLAGS_cable_driver}"
+
+if [[ -z "${cable_driver}" ]]; then
+  echo "Running with: globalnet=${globalnet}, deploytool=${deploytool}, cluster_settings=${cluster_settings}"
+else
+  echo "Running with: globalnet=${globalnet}, deploytool=${deploytool}, cluster_settings=${cluster_settings}, cable_driver=${cable_driver}"
+fi
 
 set -em
 
--- a/scripts/shared/lib/deploy_helm
+++ b/scripts/shared/lib/deploy_helm
@@ -54,31 +54,37 @@ function helm_install_subm() {
         return
     fi
 
+    local flags=(   --kube-context "${cluster}" \
+                    --name submariner \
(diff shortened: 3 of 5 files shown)


.travis.yml
```diff
@@ -12,11 +12,17 @@ jobs:
...
+  - env: CMD="make deploy deploytool=operator cable_driver=wireguard"
   - env: CMD="make deploy deploytool=operator" DEPLOY=true
 
 install:
   - sudo apt-get install curl moreutils # make ts available
   - curl https://codeload.github.com/kward/shflags/tar.gz/v1.2.3 | tar -xzf - && sudo mv shflags-1.2.3 /usr/share/shflags
+  - sudo add-apt-repository -y ppa:wireguard/wireguard # add WireGuard support
+  - sudo apt-get update
+  - sudo apt-get install linux-headers-`uname -r` -y
+  - sudo apt-get install wireguard -y
```
mkolesnik, line 24: Can these two be combine into one?

I.E.
```suggestion
  - sudo apt-get install linux-headers-`uname -r` wireguard -y
```

scripts/shared/deploy.sh
```diff
@@ -6,13 +6,21 @@ source /usr/share/shflags/shflags
...
 FLAGS "$@" || exit $?
 eval set -- "${FLAGS_ARGV}"
 
 globalnet="${FLAGS_globalnet}"
 deploytool="${FLAGS_deploytool}"
 cluster_settings="${FLAGS_cluster_settings}"
-echo "Running with: globalnet=${globalnet}, deploytool=${deploytool}, cluster_settings=${cluster_settings}"
+cableDriver="${FLAGS_cable_driver}"
+
+if [[ "${cableDriver}" == "defult" ]]; then
```
mkolesnik, line 19: Why not just print the value? It would be simplest

scripts/shared/deploy.sh
```diff
@@ -6,13 +6,21 @@ source /usr/share/shflags/shflags
...
+DEFINE_string 'cable_driver' 'default' "Cable driver implementation"
+
 FLAGS "$@" || exit $?
 eval set -- "${FLAGS_ARGV}"
 
 globalnet="${FLAGS_globalnet}"
 deploytool="${FLAGS_deploytool}"
 cluster_settings="${FLAGS_cluster_settings}"
-echo "Running with: globalnet=${globalnet}, deploytool=${deploytool}, cluster_settings=${cluster_settings}"
+cableDriver="${FLAGS_cable_driver}"
```
mkolesnik, line 17: Please no camelCase in bash :)

scripts/shared/lib/deploy_helm
```diff
@@ -54,31 +54,37 @@ function helm_install_subm() {
         return
     fi
 
+    flags=(   --kube-context "${cluster}" \
```
mkolesnik, line 57: Should be local

scripts/shared/lib/deploy_helm
```diff
@@ -54,31 +54,37 @@ function helm_install_subm() {
...
+              --set "routeAgent.image.pullPolicy=IfNotPresent" \
+              --set "engine.image.repository=localhost:5000/submariner" \
+              --set "engine.image.tag=local" \
+              --set "engine.image.pullPolicy=IfNotPresent" \
+              --set "globalnet.image.repository=localhost:5000/submariner-globalnet" \
+              --set "globalnet.image.tag=local" \
+              --set "globalnet.image.pullPolicy=IfNotPresent" \
+              --set "crd.create=${crd_create}" )
+
+    if [[ "${cableDriver}" != "default" ]] ; then
```
mkolesnik, line 82: Since `default` is a fake value, why not just send nothing in case it's not specified and here just check if the value is not empty?

scripts/shared/lib/deploy_helm
```diff
@@ -54,31 +54,37 @@ function helm_install_subm() {
...
+              --set "routeAgent.image.repository=localhost:5000/submariner-route-agent" \
+              --set "routeAgent.image.tag=local" \
+              --set "routeAgent.image.pullPolicy=IfNotPresent" \
+              --set "engine.image.repository=localhost:5000/submariner" \
+              --set "engine.image.tag=local" \
+              --set "engine.image.pullPolicy=IfNotPresent" \
+              --set "globalnet.image.repository=localhost:5000/submariner-globalnet" \
+              --set "globalnet.image.tag=local" \
+              --set "globalnet.image.pullPolicy=IfNotPresent" \
+              --set "crd.create=${crd_create}" )
```
mkolesnik, line 80: I think for this small change it's a bit of an overkill to use an array.
You can (and IMO should) just set a local variable for the flag and pass that. You can see it in action here https://github.com/submariner-io/shipyard/blob/master/scripts/shared/lib/deploy_operator#L21

scripts/shared/lib/deploy_operator
```diff
@@ -33,15 +33,21 @@ function subctl_install_subm() {
         return
     fi
 
-    subctl join --kubeconfig "${KUBECONFIGS_DIR}/kind-config-$cluster" \
+    flags=( --kubeconfig "${KUBECONFIGS_DIR}/kind-config-$cluster" \
```
mkolesnik, line 36: Same comment about the array

scripts/shared/deploy.sh
roytman: Fixed.

scripts/shared/deploy.sh
roytman: It will be simplest, but I think that a statement like 
`Running with: globalnet=false, deploytool=operator, cluster_settings=/go/src/github.com/submariner-io/shipyard/scripts/cluster_settings, cableDriver=` 
might mislead users.
What do you think ?

.travis.yml
roytman: Done.

scripts/shared/deploy.sh
roytman: Done

scripts/shared/lib/deploy_helm
roytman: Done

roytman:
Hi @mkolesnik , thanks for the review. 
I updated the code according some of your comments and would like to recheck with you others.
The cable_drive flag is an optional one, and in contrary to `--globalnet` is a string. 
Therefore I wanted to separate cases when user does not provide any value, provides one, and provides an empty string (so I used this fake value)
In addition, I don't want to specify here a default value (as in the  `--globalnet` ) example, so it we will not have to update the default value in the code and here in the scripts. 

What do you think ?

tpantelis:
> Hi @mkolesnik , thanks for the review.
> I updated the code according some of your comments and would like to recheck with you others.
> The cable_drive flag is an optional one, and in contrary to `--globalnet` is a string.
> Therefore I wanted to separate cases when user does not provide any value, provides one, and provides an empty string (so I used this fake value)
> In addition, I don't want to specify here a default value (as in the `--globalnet` ) example, so it we will not have to update the default value in the code and here in the scripts.
> 
> What do you think ?

In the submariner engine, an empty cable driver means use the default so it seems to me there's no difference between the user not providing any value and providing an empty string. I'd say default it to "".

roytman:
Hi guys thanks for the comments, I removed all arrays and all fake `default` values. 
Only https://github.com/submariner-io/shipyard/pull/102#discussion_r410889158  is not fixed yet, would like to get your second view.

scripts/shared/deploy.sh
tpantelis: I'm fine with `cable_driver=`.

roytman: OK, changed the file according the suggestion.

.travis.yml
```diff
@@ -12,11 +12,17 @@ jobs:
   include:
   - env: CMD="make validate"
   - env: CMD="make deploy deploytool=helm globalnet=true"
+  - env: CMD="make deploy deploytool=helm cable_driver=wireguard globalnet=true"
```
mkolesnik, line 15: I wouldn't add this one, there's no value in duplicating the existing jobs each time a new flag is added, otherwise we'll end up with 10's of jobs that do pretty much the same thing.

In fact since the only thing that matters is the flag, i'd add a simple job that tests only that - `make deploy cable_driver=wireguard`

roytman: I completely understand the problem of the matrix test cases. 
Actually, originally I planed to leave the decision about the automatically running test jobs to the Submariner team. 
Currently with Wireguard we have 2 jobs that cover the following 4 use cases: deployment with operator/helm and with and without `globalnet`. (helm and globalnet are new features.)

mkolesnik: Yeah same way we only test globalnet in one variation here.
And already theres another flag being added for lighthouse so i dont want to end up with 8 jobs :)

PR #103, opened 2020-04-20
Add some basic documentation

mkolesnik:
Added some very basic docs how to use shipyard

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,58 @@ The Shipyard project provides tooling for creating K8s clusters with [kind](K8s 
 - [go 1.12] with [$GOPATH configured]
 - [docker]
 
+## Usage
+To use shipyard for your project, it's easiest to use dapper and make.
+To use dapper you'll need a specific Dockerfile that dapper consumes to create a consistent environment based upon shipyard's base image.
+To use make you'll need some commands to enable dapper and also include the targets which ship in the base image.
+
+### Dockerfile.dapper
+
+The dockerfile should build upon `quay.io/submariner/shipyard-dapper-base`.
+For example:
+
+```Dockerfile
+FROM quay.io/submariner/shipyard-dapper-base
+
+ENV DAPPER_ENV="REPO TAG QUAY_USERNAME QUAY_PASSWORD TRAVIS_COMMIT" \
+    DAPPER_SOURCE=/go/src/github.com/submariner-io/submariner DAPPER_DOCKER_SOCKET=true
+ENV DAPPER_OUTPUT=${DAPPER_SOURCE}/output
+
+WORKDIR ${DAPPER_SOURCE}
+
+ENTRYPOINT ["./scripts/entry"]
+CMD ["ci"]
+```
+
+You can also refer to the project's own [Dockerfile.dapper](Dockerfile.dapper) as an example.
+
+### Makefile
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -10,6 +10,58 @@ The Shipyard project provides tooling for creating K8s clusters with [kind](K8s
 - [go 1.12] with [$GOPATH configured]
 - [docker]
 
+## Usage
+To use shipyard for your project, it's easiest to use dapper and make.
```
dfarrell07, line 14:
```suggestion
To use Shipyard for your project, it's easiest to use Dapper and Make.
```

README.md
```diff
@@ -10,6 +10,58 @@ The Shipyard project provides tooling for creating K8s clusters with [kind](K8s
...
+To use dapper you'll need a specific Dockerfile that dapper consumes to create a consistent environment based upon shipyard's base image.
```
dfarrell07, line 15:
```suggestion
To use Dapper, you'll need a specific Dockerfile that Dapper consumes to create a consistent environment based upon Shipyard's base image.
```

README.md
```diff
@@ -10,6 +10,58 @@ The Shipyard project provides tooling for creating K8s clusters with [kind](K8s
...
+To use make you'll need some commands to enable dapper and also include the targets which ship in the base image.
```
dfarrell07, line 16:
```suggestion
To use Make, you'll need some commands to enable Dapper and also include the targets which ship in the base image.
```

README.md
```diff
@@ -10,6 +10,58 @@ The Shipyard project provides tooling for creating K8s clusters with [kind](K8s
...
+
+### Dockerfile.dapper
+
+The dockerfile should build upon `quay.io/submariner/shipyard-dapper-base`.
```
dfarrell07, line 20:
```suggestion
The Dockerfile should build upon `quay.io/submariner/shipyard-dapper-base`.
```

README.md
```diff
@@ -10,6 +10,58 @@ The Shipyard project provides tooling for creating K8s clusters with [kind](K8s
...
+
+ENTRYPOINT ["./scripts/entry"]
+CMD ["ci"]
+```
+
+You can also refer to the project's own [Dockerfile.dapper](Dockerfile.dapper) as an example.
+
+### Makefile
+
+The makefile should include targets to run everything in dapper, they're defined in [Makefile.dapper](Makefile.dapper) and can be copied as is and included. To use shipyard's target simply include the [Makefile.inc](Makefile.inc) file in your own Makefile.
```
dfarrell07, line 40:
```suggestion
The Makefile should include targets to run everything in Dapper. They're defined in [Makefile.dapper](Makefile.dapper) and can be copied as-is and included. To use Shipyard's target, simply include the [Makefile.inc](Makefile.inc) file in your own Makefile.
```

README.md
mangelajo: dapper and make are normally called in low-case from the cmdline, for those may be it's better to use lowcase?

README.md
mangelajo: Dockerfile uses upcase :+1:

README.md
dfarrell07: I think if we want to use literal commands, we should use `make` or `dapper`. If we are talking about the tools themselves, I think they are proper nouns and should be Capitalized.

dfarrell07: I agree that there are a lot of cases of them lowercase though, so definitely a non-blocker. Wikipedia uses it as I described I think. https://en.wikipedia.org/wiki/Make_(software)

mkolesnik:
Thanks Daniel, I've committed your suggestions

PR #105, opened 2020-04-21
Download shflags if necessary

skitt:
This adds a new shared script which takes care of providing shflags,
using either the installed version, or a previously-downloaded
version, or by downloading it.


mangelajo:
@skitt shellcheck complains for a var quoting: https://travis-ci.com/github/submariner-io/shipyard/jobs/321537768#L431

PR #107, opened 2020-04-21
Fix shflags handling when the package is installed

skitt:
If shflags is available in /usr/share/shflags/shflags, we mustn’t try
to load .shflags.

### DIFF
--- a/scripts/shared/lib/shflags
+++ b/scripts/shared/lib/shflags
@@ -2,16 +2,23 @@
 # shellcheck source=scripts/shared/lib/source_only
 . "${BASH_SOURCE%/*}"/source_only
 
-if [ -f /usr/share/shflags/shflags ]; then
-    . /usr/share/shflags/shflags
-elif [ ! -f .shflags ]; then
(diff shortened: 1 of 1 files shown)


scripts/shared/lib/shflags
```diff
@@ -4,14 +4,16 @@
 
 if [ -f /usr/share/shflags/shflags ]; then
     . /usr/share/shflags/shflags
```
mkolesnik, line 6: I think it'll be much cleaner if you just put exit 0 after this line

Too much indentation makes code harder to read and maintain

skitt: I don’t think you’ll like it when the sourced `shflags` exits ;-).

skitt: I’ve moved the logic into a function so it’s clearer, hopefully.

mkolesnik: Ah right i was thinking of `return` style here but i guess you cant `return` if not in a function..

PR #109, opened 2020-04-22
Retry KIND deployment

mkolesnik:
Should it fail, it won't hurt to retry deployment. KIND can be finicky
sometimes and for some reason fail to deploy one of the clusters.

The only caveat is that should the kind deployment break, it'll take
longer to fail the script since it will retry 3 times, but that is a
small price to pay (and good incentive not to break the deployment!)


skitt:
Amazingly enough, it looks like the KIND setup failed in one of the test jobs, and the kubeconfigs got messed up, possible as a result of the retry.

mkolesnik:
> Amazingly enough, it looks like the KIND setup failed in one of the test jobs, and the kubeconfigs got messed up, possible as a result of the retry.

What a lucky coincidence :smile: 
I've fixed it now (and tested locally to be sure)

mkolesnik:
I've restarted the build a couple of times and all passed.

PR #111, opened 2020-04-23
Allow passing extra arguments to deploytool

mkolesnik:
Allows passing extra arguments by consuming projects.

Also had to fix the way extra arguments are treated in makefile so they don`t overwrite the user provided value.


Makefile
```diff
@@ -2,8 +2,9 @@ ifneq (,$(DAPPER_HOST_ARCH))
 
 # Running in Dapper
 
-CLUSTERS_ARGS = --cluster_settings $(DAPPER_SOURCE)/scripts/cluster_settings
-DEPLOY_ARGS = $(CLUSTERS_ARGS)
+CLUSTER_SETTINGS_FLAG = --cluster_settings $(DAPPER_SOURCE)/scripts/cluster_settings
+CLUSTERS_ARGS += $(CLUSTER_SETTINGS_FLAG)
+DEPLOY_ARGS += $(CLUSTER_SETTINGS_FLAG)
```
skitt, line 7: Ah yes, this is much better :-).

PR #113, opened 2020-04-23
Make the number of clusters configurable

tpantelis:
Not all project E2E tests need 3 clusters so make it configurable.

A new indexed "clusters" array was added to the cluster_settings file
to hold the cluster names.

### DIFF
--- a/scripts/shared/cleanup.sh
+++ b/scripts/shared/cleanup.sh
@@ -8,9 +8,7 @@ source ${SCRIPTS_DIR}/lib/utils
 ### Functions ###
 
 function delete_cluster() {
-    if kind get clusters | grep -q ${cluster}; then
-        kind delete cluster --name=${cluster};
-    fi
+    kind delete cluster --name=${cluster};
 }
 
 function stop_local_registry {
@@ -23,7 +21,9 @@ function stop_local_registry {
 
 ### Main ###
 
-run_all_clusters delete_cluster
+clusters=($(kind get clusters))
+run_parallel "${clusters[*]}" delete_cluster
+
 stop_local_registry
 docker system prune --volumes -f
 
--- a/scripts/shared/clusters.sh
+++ b/scripts/shared/clusters.sh
@@ -25,6 +25,13 @@ source ${SCRIPTS_DIR}/lib/utils
 source "${SCRIPTS_DIR}/lib/cluster_settings"
 [[ -z "${cluster_settings}" ]] || source ${cluster_settings}
 
+cat << EOM
+Cluster settings::
+  clusters - ${clusters[*]@Q}
+  nodes per cluster - $(typeset -p cluster_nodes | cut -f 2- -d=)
+  install submariner - $(typeset -p cluster_subm | cut -f 2- -d=)
+EOM
+
 ### Functions ###
 
 function render_template() {
@@ -123,3 +130,4 @@ with_retries 3 run_all_clusters create_kind_cluster
 declare_kubeconfig
 run_subm_clusters deploy_weave_cni
 
+print_clusters_message
--- a/scripts/shared/deploy.sh
+++ b/scripts/shared/deploy.sh
@@ -52,5 +52,9 @@ install_subm_all_clusters
 
 deploytool_postreqs
 
-with_context cluster2 connectivity_tests
+if [ "${#clusters[@]}" -gt 2 ]; then
+    with_context "${clusters[1]}" connectivity_tests
+else
+    echo "Not executing connectivity tests - requires at least 3 clusters"
+fi
 
--- a/scripts/shared/lib/utils
+++ b/scripts/shared/lib/utils
@@ -53,52 +53,48 @@ function with_context() {
 }
 
 # Run cluster commands in parallel.
-# 1st argument is the numbers of the clusters to run for, supports "1 2 3" or "{1..3}" for range
+# 1st argument is the cluster names for which to run.
 # 2nd argument is the command to execute, which will have the $cluster variable set.
 # 3rd argument and so forth get passed to the command.
 function run_parallel() {
-    local clusters cmnd
-    clusters=$(eval echo "$1")
-    cmnd=$2
+    local cmnd=$2
     declare -A pids
-    for i in ${clusters}; do
+    for cluster in $(eval echo "$1"); do
         (
            set -o pipefail
-           with_context "cluster${i}" "$cmnd" "${@:3}" | sed "s/^/[cluster${i}] /"
+           $cmnd "${@:3}" | sed "s/^/[${cluster}] /"
         ) &
-        pids["${i}"]=$!
+        pids["${cluster}"]=$!
     done
 
-    for i in "${!pids[@]}"; do
-        wait ${pids[$i]}
+    for pid in "${pids[@]}"; do
+        wait $pid
     done
 }
 
 # Run the given command (with any arguments) in parallel on all the clusters.
 function run_all_clusters() {
-    run_parallel "{1..3}" "$@"
+    run_parallel "${clusters[*]}" "$@"
 }
 
 # Run the given command (with any arguments) in parallel on the clusters that install submariner.
 function run_subm_clusters() {
     declare -a subm_clusters
-    for i in {1..3}; do
-        [[ "${cluster_subm[cluster$i]}" != "true" ]] || subm_clusters+=( "$i" )
+    for cluster in "${clusters[@]}"; do
+        [[ "${cluster_subm[${cluster}]}" != "true" ]] || subm_clusters+=( "${cluster}" )
     done
 
     run_parallel "${subm_clusters[*]}" "$@"
 }
 
 # Run cluster commands sequentially.
-# 1st argument is the numbers of the clusters to run for, supports "1 2 3" or "{1..3}" for range
+# 1st argument is the numbers of the clusters names for which to run.
 # 2nd argument is the command to execute, which will have the $cluster variable set.
 # 3rd argument and so forth get passed to the command.
 function run_sequential() {
-    local clusters cmnd
-    clusters=$(eval echo "$1")
-    cmnd=$2
-    for i in ${clusters}; do
-        with_context "cluster${i}" "$cmnd" "${@:3}" | sed "s/^/[cluster${i}] /"
+    local cmnd=$2
+    for cluster in $(eval echo "$1"); do
+        $cmnd "${@:3}" | sed "s/^/[${cluster}] /" 
     done
 }
 
@@ -108,8 +104,8 @@ function registry_running() {
 }
 
 function add_cluster_cidrs() {
-    local idx="cluster${1}"
     local val=$1
+    local idx=$2
     [[ $globalnet != "true" ]] || val="0"
     cluster_CIDRs[$idx]="10.24${val}.0.0/16"
     service_CIDRs[$idx]="100.9${val}.0.0/16"
@@ -120,11 +116,25 @@ function declare_cidrs() {
     # shellcheck disable=SC2034 # these variables are used elsewhere
     declare -gA cluster_CIDRs service_CIDRs global_CIDRs
 
-    for i in {1..3}; do
-        add_cluster_cidrs $i
+    i=1
+    for cluster in "${clusters[@]}"; do
+        add_cluster_cidrs "$i" "$cluster"
+        i=$(("$i"+1))
     done
 }
 
 function declare_kubeconfig() {
     source "${SCRIPTS_DIR}"/lib/kubecfg
+}
+
+function print_clusters_message() {
+    cat << EOM
+Your virtual cluster(s) are deployed and working properly and can be accessed with:
+
+export KUBECONFIG=\$(find \$(git rev-parse --show-toplevel)/output/kubeconfigs/ -type f | -printf %p:)
+
+$ kubectl config use-context cluster1 # or cluster2, cluster3..
+
+To clean everthing up, just run: make cleanup
+EOM
 }
--- a/scripts/shared/lib/cluster_settings
+++ b/scripts/shared/lib/cluster_settings
@@ -3,6 +3,9 @@
 # shellcheck source=scripts/shared/lib/source_only
 . "${BASH_SOURCE%/*}"/source_only
 
+# Array of all cluster names that we create and deploy on.
+declare -ga clusters=(cluster{1..3})
+
 # Map of cluster names to a space separated string, representing a list of nodes to deploy.
 # Possible node types are 'control-plane' and 'worker'.
 # e.g. cluster_nodes['multi-master']="control-plane control-plane worker worker worker"
--- a/scripts/shared/lib/deploy_funcs
+++ b/scripts/shared/lib/deploy_funcs
@@ -52,7 +52,7 @@ function test_connection() {
     local source_pod=$1
     local target_address=$2
 
-    echo "Attempting connectivity between clusters - $source_pod (cluster2) --> $target_address (service on cluster3)"
+    echo "Attempting connectivity between clusters - $source_pod (${clusters[1]}) --> $target_address (service on ${clusters[2]})"
     if ! kubectl exec "${source_pod}" -- curl --output /dev/null -m 30 --silent --head --fail "${target_address}"; then
         return 1
     fi
@@ -62,16 +62,16 @@ function test_connection() {
 
 function connectivity_tests() {
     deploy_resource "${RESOURCES_DIR}/netshoot.yaml"
-    with_context cluster3 deploy_resource "${RESOURCES_DIR}/nginx-demo.yaml"
+    with_context "${clusters[2]}" deploy_resource "${RESOURCES_DIR}/nginx-demo.yaml"
 
     local netshoot_pod nginx_svc_ip
     netshoot_pod=$(kubectl get pods -l app=netshoot | awk 'FNR == 2 {print $1}')
-    nginx_svc_ip=$(with_context cluster3 get_svc_ip nginx-demo)
+    nginx_svc_ip=$(with_context "${clusters[2]}" get_svc_ip nginx-demo)
 
     with_retries 5 test_connection "$netshoot_pod" "$nginx_svc_ip"
 
     remove_resource "${RESOURCES_DIR}/netshoot.yaml"
-    with_context cluster3 remove_resource "${RESOURCES_DIR}/nginx-demo.yaml"
+    with_context "${clusters[2]}" remove_resource "${RESOURCES_DIR}/nginx-demo.yaml"
 }
 
 function add_subm_gateway_label() {


scripts/shared/cleanup.sh
```diff
@@ -23,7 +21,11 @@ function stop_local_registry {
 
 ### Main ###
 
-run_parallel "{1..3}" delete_cluster
+n=$(kind get clusters | wc -l)
```
mkolesnik, line 24: This is not good, you need to extract the exact names and delete them since there could be gaps for example..

scripts/shared/clusters.sh
```diff
@@ -23,6 +31,14 @@ source ${SCRIPTS_DIR}/lib/utils
 
 # Always source the shared cluster settings, to set defaults in case something wasn't set in the provided settings
 source "${SCRIPTS_DIR}/lib/cluster_settings"
+
+if [ ${num_clusters} -gt 3 ]; then
```
mkolesnik, line 35: This is awkward.. What if they're actually set it cluster_settings?

scripts/shared/lib/utils
```diff
@@ -105,12 +105,27 @@ function declare_cidrs() {
...
 
 function declare_kubeconfig() {
+    n=$(kind get clusters | wc -l)
     export KUBECONFIG
-    KUBECONFIG=$(echo "${KUBECONFIGS_DIR}"/kind-config-cluster{1..3} | sed 's/ /:/g')
+    KUBECONFIG=$(eval echo "${KUBECONFIGS_DIR}/kind-config-cluster{1..$n}" | sed 's/ /:/g')
+}
+
+function print_clusters_message() {
+    num_clusters=$(kind get clusters | wc -l)
```
mkolesnik, line 121: This should be the actual cluster names

tpantelis: You mean like this `clusters=$(kind get clusters)`? But then how do I get it to expand with a range on line L125?

scripts/shared/lib/utils
```diff
@@ -105,12 +105,27 @@ function declare_cidrs() {
...
+cat << EOM
+Your ${num_clusters} virtual cluster(s) are deployed and working properly and can be accessed with:
+
+export KUBECONFIG=\$(echo \$(git rev-parse --show-toplevel)/output/kubeconfigs/kind-config-cluster{1..${num_clusters}} | sed 's/ /:/g')
```
mkolesnik, line 125:
```suggestion
echo "export KUBECONFIG=\$(echo \$(find \$(git rev-parse --show-toplevel)/output/kubeconfigs/ -type f) | tr ' ' ':')"
```

scripts/shared/cleanup.sh
tpantelis: This is essentially the same as to what the previous code did except I'm getting the actual #  of clusters instead of hard coding to 3.  That said I agree deleting the actual names retrieved from kind is better although we can't use `run_parallel` but we can still delete in parallel.

mkolesnik:
> We already have the CLUSTER_ARGS var which is used to pass other args to the script like `globalnet` and `registry_inmemory` so I'm not really clear on why you indicate the `num_clusters` arg doesn't fit in or integrate well with that mechanism. We could pass it via the _cluster_settings_ file but then one has to create a file as opposed to just simply doing `CLUSTERS_ARGS=--num_clusters 2`. You mention another unified mechanism - I assume you mean migrate all the args/settings to say yaml eg - which would be fine but I think that's a larger architectural change that is way out of scope for this PR.

I understand this looks weird, but the flags we have in the scripts are not cluster level related, e.g. you either deploy all or nothing with globalnet..

For per-cluster settings we have cluster_settings which accepts a file with expected arrays, this could be another expected array such as `clusters=(cluster1 cluster2 cluster3)`. We could then use the `"${clusters[@]}" ` variable anywhere we need all cluster names.
Also consider that defining the array there is coupled with the other settings, making it easier to review and see if there's any discrepancies.

Another nice benefit would be the ability to name the clusters eg `clusters=(broker east west)` or some such. Of course this will need a few changes in Shipyard, but I was aiming at that anyway.

> 
> I'd like to make progress on this PR as it's blocking other work so it would be great if you could elaborate on exactly what you'd prefer on how to pass the `num_clusters` arg.

Again you can currently skip deploying subm entirely by using `cluster_subm[`clusterX`]="false"` in cluster settings, so you can theoretically skip them all.

I'll be happy to discuss this further or propose an alternative myself.

mkolesnik:
Don't want to leave you hanging, please see my alternative  proposal at #121

tpantelis:
> Don't want to leave you hanging, please see my alternative proposal at #121

I incorporated the indexed `clusters` array proposed by #121. The script code has been changed to use the specified names from `clusters` rather then assuming the names are of the form _cluster{1..N}_.

scripts/shared/clusters.sh
```diff
@@ -25,6 +24,8 @@ source ${SCRIPTS_DIR}/lib/utils
 source "${SCRIPTS_DIR}/lib/cluster_settings"
 [[ -z "${cluster_settings}" ]] || source ${cluster_settings}
 
+echo "Running with: clusters="${clusters[*]}", k8s_version=${version}, globalnet=${globalnet}, registry_inmemory=${registry_inmemory}, cluster_settings=${cluster_settings}"
```
mkolesnik, line 27: I don't think you should move it since you'll be conflating flags with other settings.

Instead I recommend you simply print here the clusters (and perhaps other settings that might be relevant for clusters.sh) as it would be nice to see the final values we'll be working with.

scripts/shared/clusters.sh
```diff
@@ -36,7 +37,7 @@ function generate_cluster_yaml() {
     local service_cidr="${service_CIDRs[${cluster}]}"
     local dns_domain="${cluster}.local"
     local disable_cni="true"
-    if [[ "${cluster}" = "cluster1" ]]; then
+    if [[ "${cluster}" = "${clusters[0]}" ]]; then
```
mkolesnik, line 40: Please leave as is, we'll handle broker cluster in a different PR

scripts/shared/lib/utils
```diff
@@ -105,12 +101,33 @@ function declare_cidrs() {
     # shellcheck disable=SC2034 # these variables are used elsewhere
     declare -gA cluster_CIDRs service_CIDRs global_CIDRs
 
-    for i in {1..3}; do
-        add_cluster_cidrs $i
+    i=1
```
mkolesnik, line 104: You should notice this function is also called here: https://github.com/submariner-io/submariner-operator/blob/master/scripts/kind-e2e/lib_operator_verify_subm.sh#L27

I think it's safer is you just handle clusters internally

scripts/shared/lib/utils
```diff
@@ -105,12 +101,33 @@ function declare_cidrs() {
...
+    for cluster in $(eval echo "$1"); do
```
mkolesnik, line 105:
```suggestion
    for cluster in ${clusters[*]}; do
```

scripts/shared/cleanup.sh
```diff
@@ -23,7 +21,9 @@ function stop_local_registry {
 
 ### Main ###
 
-run_parallel "{1..3}" delete_cluster
+clusters=($(kind get clusters))
```
mkolesnik, line 24: This is a bit more destructive than what we had, since you might delete kind clusters you didnt plan on deleting.

I guess it should be fine for now but we should probably have a nicer alternative..

scripts/shared/clusters.sh
```diff
@@ -113,8 +114,11 @@ rm -rf ${KUBECONFIGS_DIR}
...
 
 run_local_registry
-declare_cidrs
-with_retries 3 run_parallel "{1..3}" create_kind_cluster
+declare_cidrs "${clusters[*]}"
+with_retries 3 run_parallel "${clusters[*]}" create_kind_cluster
 declare_kubeconfig
-run_parallel "2 3" deploy_weave_cni
 
+weave_clusters=(${clusters[@]:1})
```
mkolesnik, line 121: No need for this if #127 is accepted

scripts/shared/deploy.sh
```diff
@@ -45,12 +45,16 @@ import_image quay.io/submariner/submariner-route-agent
 load_deploytool $deploytool
 deploytool_prereqs
 
-run_parallel "{1..3}" prepare_cluster "$SUBM_NS"
+run_parallel "${clusters[*]}" prepare_cluster "$SUBM_NS"
 
-with_context cluster1 setup_broker
+with_context "${clusters[0]}" setup_broker
```
mkolesnik, line 50: Please leave as is, we'll handle broker cluster in a different PR

scripts/shared/lib/deploy_helm
```diff
@@ -91,9 +91,10 @@ function helm_install_subm() {
 
 
 function install_subm_all_clusters() {
-    with_context cluster1 helm_install_subm false
+    with_context "${clusters[0]}" helm_install_subm false
```
mkolesnik, line 94: Please leave as is, we'll handle broker cluster in a different PR

scripts/shared/lib/deploy_operator
```diff
@@ -61,15 +61,15 @@ function install_subm_all_clusters() {
...
+        run_parallel "${clusters[*]}" subctl_install_subm
     fi
 }
 
 function deploytool_postreqs() {
     # FIXME: Make this unnecessary using subctl v0.0.4 --no-label flag
     # subctl wants a gateway node labeled, or it will ask, but this script is not interactive,
     # and E2E expects cluster1 to not have the gateway configured at start, so we remove it
-    with_context cluster1 del_subm_gateway_label
+    with_context "${clusters[0]}" del_subm_gateway_label
```
mkolesnik, line 74: Please leave as is, we'll handle broker cluster in a different PR

scripts/shared/deploy.sh
```diff
@@ -45,12 +45,16 @@ import_image quay.io/submariner/submariner-route-agent
...
 install_subm_all_clusters
 
 deploytool_postreqs
 
-with_context cluster2 connectivity_tests
+if [ "${#clusters[@]}" -gt 3 ]; then
```
mkolesnik, line 55:
```suggestion
if [ "${#clusters[@]}" -gt 2 ]; then
```

Otherwise even with 3 clusters it wont run it

scripts/shared/lib/deploy_helm
```diff
@@ -91,9 +91,10 @@ function helm_install_subm() {
...
 
-    run_parallel "2 3" helm_install_subm true
+    rest=("${clusters[@]:1}")
```
mkolesnik, line 96: No need for this if we have #127

scripts/shared/lib/utils
```diff
@@ -53,37 +53,33 @@ function with_context() {
...
 # 2nd argument is the command to execute, which will have the $cluster variable set.
 # 3rd argument and so forth get passed to the command.
 function run_parallel() {
-    local clusters cmnd
-    clusters=$(eval echo "$1")
-    cmnd=$2
+    local cmnd=$2
     declare -A pids
-    for i in ${clusters}; do
+    for cluster in $(eval echo "$1"); do
```
mkolesnik, line 62:
```suggestion
    for cluster in $1; do
```

The `eval` was there to support `{1..3}` syntax but since it's not used anymore then no need for it

scripts/shared/lib/utils
```diff
@@ -53,37 +53,33 @@ function with_context() {
...
         (
            set -o pipefail
-           with_context "cluster${i}" "$cmnd" "${@:3}" | sed "s/^/[cluster${i}] /"
+           with_context "${cluster}" "$cmnd" "${@:3}" | sed "s/^/[${cluster}] /"
```
mkolesnik, line 65:
```suggestion
           $cmnd "${@:3}" | sed "s/^/[${cluster}] /"
```

No need to call `with_context` wince you've already set cluster

scripts/shared/lib/utils
```diff
@@ -53,37 +53,33 @@ function with_context() {
...
         ) &
-        pids["${i}"]=$!
+        pids["${cluster}"]=$!
     done
 
-    for i in "${!pids[@]}"; do
-        wait ${pids[$i]}
+    for cluster in "${!pids[@]}"; do
+        wait ${pids[$cluster]}
```
mkolesnik, line 71:
```suggestion
    for pid in "${pids[@]}"; do
        wait $pid
```

I just realized this should be possible ;)

scripts/shared/lib/utils
```diff
@@ -105,12 +101,33 @@ function declare_cidrs() {
...
-    KUBECONFIG=$(echo "${KUBECONFIGS_DIR}"/kind-config-cluster{1..3} | sed 's/ /:/g')
+    KUBECONFIG=
+    for cluster in ${clusters}; do
+        [ -n "${KUBECONFIG}" ] && KUBECONFIG+=":" 
+        KUBECONFIG+="${KUBECONFIGS_DIR}/kind-config-${cluster}"
+    done
+}
+
+function print_clusters_message() {
+    num_clusters=$(kind get clusters | wc -l)
```
mkolesnik, line 123: I wouldn't use this since you could have other kind clusters not related... You can just not print the number, or alternatively you could count the number of kubeconfig files.

tpantelis: Wouldn't counting the number of kubeconfig files reveal the same # as `kind get clusters`?  I'm fine with omitting the number.

scripts/shared/cleanup.sh
tpantelis: I guess we'd need to pass the cluster_settings. I think it's reasonable to delete all of them.

scripts/shared/deploy.sh
tpantelis: yeah I meant 2 😄

scripts/shared/lib/utils
tpantelis: Not sure I follow - where have we already done the equivalent of `with_context`?

scripts/shared/cleanup.sh
mkolesnik: We can add this is a later PR anyhow since it's not critical

scripts/shared/clusters.sh
mkolesnik: I think the `echo` of the flags should remain where it was with the values it had.

Here you can add another `echo` which outputs the combined settings of the clusters after it processed cluster_settings.

scripts/shared/clusters.sh
mkolesnik: Yes for now, we'll take care of broker cluster better in a different PR

This is tangential to your PR anyhow..

scripts/shared/lib/utils
mkolesnik: `for cluster in` already has set `$cluster`

scripts/shared/lib/utils
```diff
@@ -105,12 +101,33 @@ function declare_cidrs() {
...
-        add_cluster_cidrs $i
+    i=1
+    for cluster in $(eval echo "$1"); do
+        add_cluster_cidrs "$i" "$cluster"
+        i=$(("$i"+1))
     done
 }
 
 function declare_kubeconfig() {
+    local clusters
```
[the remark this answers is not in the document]
mkolesnik: We already depend on `git` and specifically `git rev-parse` in other places.

Besides we're dogfooding our suggestion to the user this way (perhaps we can even extract this to a constant and use it in both places).

Also you could just put `${DAPPER_SOURCE}` if you wish to avoid `git rev-parse`

scripts/shared/lib/utils
mkolesnik: Not necessarily, anyway just omit it it'll be the simplest :)

scripts/shared/clusters.sh
tpantelis: So duplicate the echo with "clusters=" added? Not sure I see the purpose of that. Unless it's possible the flags outputted earlier might change after _cluster_settings_ is sourced? Even so I don't see the purpose of outputting  an intermediate value.

scripts/shared/clusters.sh
tpantelis: Well not really b/c I'm also trying to remove hard-coded cluster names. There's not much purpose in allowing the user to specify the cluster names if we're going to assume certain names anyway 😄  I'd rather be complete and I don't see the harm in changing these references now. I mean I don't see why it would matter to a subsequent PR (if/whenever that lands) if this is referenced as "cluster1" or "${clusters[0]}".

scripts/shared/lib/utils
tpantelis: OK - I'm not really clear how `with_context` works - the cryptic "${@:3}" syntax is a bit above my rudimentary scripting knowledge but I'll take your word for it 😄

scripts/shared/lib/utils
tpantelis: That's fine - either way submariner will need to change to source _cluster_settings_.

scripts/shared/lib/utils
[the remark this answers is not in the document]
tpantelis: yeah I can use `${DAPPER_SOURCE}`  here.

scripts/shared/lib/utils
```diff
@@ -53,37 +53,33 @@ function with_context() {
 }
 
 # Run cluster commands in parallel.
-# 1st argument is the numbers of the clusters to run for, supports "1 2 3" or "{1..3}" for range
+# 1st argument is the cluster names for which to run.
```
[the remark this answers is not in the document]
tpantelis: I see you pushed PRs to lighthouse and submariner-operator to use the new functions.

scripts/shared/lib/utils
```diff
@@ -120,12 +116,26 @@ function declare_cidrs() {
...
 }
 
 function declare_kubeconfig() {
     export KUBECONFIG
-    KUBECONFIG=$(echo "${KUBECONFIGS_DIR}"/kind-config-cluster{1..3} | sed 's/ /:/g')
+    KUBECONFIG=$(find "${KUBECONFIGS_DIR}" -type f | tr '\n' ':')
+}
+
+function print_clusters_message() {
+cat << EOM
```
mkolesnik, line 132:
```suggestion
    cat << EOM
```

Since it's technically inside the function :)

scripts/shared/clusters.sh
```diff
@@ -14,7 +14,6 @@ version="${FLAGS_k8s_version}"
 [[ "${FLAGS_globalnet}" = "${FLAGS_TRUE}" ]] && globalnet=true || globalnet=false
 [[ "${FLAGS_registry_inmemory}" = "${FLAGS_TRUE}" ]] && registry_inmemory=true || registry_inmemory=false 
 cluster_settings="${FLAGS_cluster_settings}"
-echo "Running with: k8s_version=${version}, globalnet=${globalnet}, registry_inmemory=${registry_inmemory}, cluster_settings=${cluster_settings}"
```
mkolesnik, line 17: Leave this as is

scripts/shared/clusters.sh
tpantelis: Done

tpantelis:
> You missed a comment on utils but its non blocker if you want to move ahead

Sorry there's been a lot of comments.

scripts/shared/lib/utils
tpantelis: Done

scripts/shared/lib/utils
```diff
@@ -120,12 +116,26 @@ function declare_cidrs() {
...
+function print_clusters_message() {
+    cat << EOM
+Your virtual cluster(s) are deployed and working properly and can be accessed with:
+
+export KUBECONFIG=\$(find \$(git rev-parse --show-toplevel)/output/kubeconfigs/ -type f | tr '\n' ':')
+
+$ kubectl config use-context cluster1 # or cluster2, cluster3..
+
+To clean everthing up, just run: make cleanup
+    EOM
```
mkolesnik, line 140:
```suggestion
EOM
```

This should fix the CI (I know it's confusing. but that's linux antics for you :))

mangelajo: I merged it right away so we can check CI.

scripts/shared/lib/utils
```diff
@@ -53,52 +53,48 @@ function with_context() {
...
-    run_parallel "{1..3}" "$@"
+    run_parallel "${clusters[*]}" "$@"
 }
 
 # Run the given command (with any arguments) in parallel on the clusters that install submariner.
 function run_subm_clusters() {
     declare -a subm_clusters
-    for i in {1..3}; do
-        [[ "${cluster_subm[cluster$i]}" != "true" ]] || subm_clusters+=( "$i" )
+    for cluster in ${clusters[*]}; do
```
skitt, line 83:
```suggestion
    for cluster in "${clusters[@]}"; do
```

scripts/shared/lib/utils
```diff
@@ -53,52 +53,48 @@ function with_context() {
...
-        wait ${pids[$i]}
+    for pid in "${pids[@]}"; do
+        wait $pid
     done
 }
 
 # Run the given command (with any arguments) in parallel on all the clusters.
 function run_all_clusters() {
-    run_parallel "{1..3}" "$@"
+    run_parallel "${clusters[*]}" "$@"
```
skitt, line 77:
```suggestion
    run_parallel "${clusters[@]}" "$@"
```

scripts/shared/lib/utils
```diff
@@ -120,12 +116,26 @@ function declare_cidrs() {
     # shellcheck disable=SC2034 # these variables are used elsewhere
     declare -gA cluster_CIDRs service_CIDRs global_CIDRs
 
-    for i in {1..3}; do
-        add_cluster_cidrs $i
+    i=1
+    for cluster in ${clusters[*]}; do
```
skitt, line 120:
```suggestion
    for cluster in "${clusters[@]}"; do
```

scripts/shared/lib/utils
```diff
@@ -120,12 +116,26 @@ function declare_cidrs() {
...
+        add_cluster_cidrs "$i" "$cluster"
+        i=$(("$i"+1))
     done
 }
 
 function declare_kubeconfig() {
     export KUBECONFIG
-    KUBECONFIG=$(echo "${KUBECONFIGS_DIR}"/kind-config-cluster{1..3} | sed 's/ /:/g')
+    KUBECONFIG=$(find "${KUBECONFIGS_DIR}" -type f | tr '\n' ':')
```
skitt, line 128:
```suggestion
    KUBECONFIG=$(find "${KUBECONFIGS_DIR}" -type f -printf %p:)
```

scripts/shared/lib/utils
tpantelis: Using the '@' syntax doesn't work.

skitt: Ah of course, `run_parallel` needs its first argument to contain all the targets...

PR #115, opened 2020-04-26
Add shared e2e.sh script

mkolesnik:
Seems this script (the basic part to run the tests) is pretty much the
same across projects, hence adding a shared version that acknowledges
cluster_settings to generate the correct test contexts.

Also changed the CI config to run the e2e, and use the proper make
variables.


scripts/shared/e2e.sh
```diff
@@ -0,0 +1,69 @@
...
+DEFINE_string 'cluster_settings' '' "Settings file to customize cluster deployments"
+DEFINE_string 'focus' '.*' 'Ginkgo focus for the E2E tests'
+FLAGS "$@" || exit $?
+eval set -- "${FLAGS_ARGV}"
+
+focus="${FLAGS_focus}"
+cluster_settings="${FLAGS_cluster_settings}"
+
+set -o pipefail 
+set -em
```
dfarrell07, line 15: Is there some reason I'm missing to do this `set` in two lines?

```suggestion
set -em -o pipefail
```

mkolesnik: Thanks i was thinking about it at some point but forgot to merge em, i've added the commit to the PR :)

.travis.yml
```diff
@@ -11,9 +11,9 @@ git:
 jobs:
   include:
   - env: CMD="make validate"
-  - env: CMD="make deploy deploytool=helm globalnet=true"
-  - env: CMD="make deploy deploytool=operator" DEPLOY=true
-  - env: CMD="make deploy cable_driver=wireguard"
+  - env: CMD="make deploy" CLUSTERS_ARGS="--globalnet true" DEPLOY_ARGS="--globalnet true --deploytool helm"
+  - env: CMD="make deploy" DEPLOY_ARGS="--cable_driver wireguard"
+  - env: CMD="make e2e" DEPLOY=true
```
skitt, line 16: This should really be a separate commit, but I don’t mind either way.

mkolesnik: Yeah you're right if you want i can break it out

skitt: Overall I think it would be better split out, if only to make it clearer when looking back at the history (why did we add `DEPLOY_ARGS` to Travis? no this doesn’t have anything to do with sharing e2e...).

scripts/shared/e2e.sh
```diff
@@ -0,0 +1,68 @@
...
+
+cat << EOM
+Your 3 virtual clusters are deployed and working properly with your local submariner source code, and can be accessed with:
+
+export KUBECONFIG=\$(echo \$(git rev-parse --show-toplevel)/output/kubeconfigs/kind-config-cluster{1..3} | sed 's/ /:/g')
+
+$ kubectl config use-context cluster1 # or cluster2, cluster3..
+
+To clean evertyhing up, just run: make cleanup
+EOM
```
tpantelis, line 68: I'm not sure of the general usefulness of having this in shipyard. It seems this would only be useful for submariner. Lighthouse has to do additional core-dns setup in its _e2e.sh_.  Admiral and shipyard tests (in progress) don't deploy submariner.  Perhaps it would be better to make more utility functions, eg `test_with_e2e_tests`. I already added utility to print the clusters message.

mkolesnik: Actually lighthouse runs E2E tests so it makes sense to have this shared. The additional setup will be done in an intermediary step.
As for Admiral and shipyard, we'll have to see exactly what we run in e2e, but at least in shipyard we do deploy to test the deployment itself is working so it's not like we're going to remove it. In Admiral I see that you are adding some deployment capability. For that I suggest we can use cluster_settings to skip deployments on certain clusters as it's already possible, instead of running your own deployment procedure. Alternatively you can define your own `deploy` target there and still use the E2E code here.

If we need to make adjustments we can always tweak what we need of course, but we should make the typical use case shared IMHO and not just some functions which are scattered.

.travis.yml
mkolesnik: OK broke it out to PR #117

scripts/shared/e2e.sh
```diff
@@ -0,0 +1,68 @@
...
+        fi
+    done
+}
+
+function test_with_e2e_tests {
+    cd ${DAPPER_SOURCE}/test/e2e
+
+    go test -v -args -ginkgo.v -ginkgo.randomizeAllSpecs \
+        -submariner-namespace $SUBM_NS $(generate_context_flags) \
+        -ginkgo.noColor -ginkgo.reportPassed \
```
tpantelis, line 47: Personally I prefer the color output - I don't know why we started using `-ginkgo.noColor` but I'd like to remove it.

.travis.yml
dfarrell07: Yeah I was confused by that too, thanks.

scripts/shared/e2e.sh
mkolesnik: OK removed it

mkolesnik:
Had to force push due to conflicts with the master, only thing that was changed was removal of `-gonkgo.noColor` flag

scripts/shared/e2e.sh
```diff
@@ -0,0 +1,68 @@
...
+source "${SCRIPTS_DIR}/lib/cluster_settings"
+[[ -z "${cluster_settings}" ]] || source ${cluster_settings}
+
+### Functions ###
+
+function deploy_env_once() {
+    if with_context cluster3 kubectl wait --for=condition=Ready pods -l app=submariner-engine -n "${SUBM_NS}" --timeout=3s > /dev/null 2>&1; then
+        echo "Submariner already deployed, skipping deployment..."
+        return
+    fi
```
tpantelis, line 29: Why is this necessary since the `e2e` make target depends/should depend on the `deploy` target?  Also this assumes that submariner is actually deployed by the project invoking it (admiral does not).

scripts/shared/e2e.sh
```diff
@@ -0,0 +1,68 @@
...
+
+    make deploy
+}
+
+function generate_context_flags() {
+    for cluster in "${!cluster_subm[@]}"; do
+        if [[ "${cluster_subm[$cluster]}" = "true" ]]; then
+            printf " -dp-context $cluster"
+        fi
+    done
```
tpantelis, line 39: I don't think this will generally work as expected for all use cases. First `cluster_subm` iteration order is random - submariner, eg,  expects a specific  order: `cluster2 cluster3 cluster1`.  Perhaps we can utilize the ordering of the `clusters` array being introduced by https://github.com/submariner-io/shipyard/pull/113. Second, I don't think we can assume that if `cluster_subm$cluster]` is false then the cluster shouldn't be passed in.

scripts/shared/e2e.sh
mkolesnik: Yeah it will mostly be true.. It was part of the "lazy deploy" change on submariner https://github.com/submariner-io/submariner/pull/512

However I can add a flag to toggle this, if that makes sense let me know and I'll add one

scripts/shared/e2e.sh
mkolesnik: Yeah if it makes sense I don't mind passing all the clusters.
However I thought it is expecting only ones that have subm deployed?

As for the order I'm not quite sure why that matters to the e2e tests, shouldn't they work with any order? If we need specific cluster to identify a specific condition (eg gateway failover) then we should probably have an extra flag?

I could change this to receive the clusters order as parameters instead of trying to guess it, in which case the caller would be responsible (and would know it for sure since he's defining the deployment anyhow).

WDYT?

BTW it seems that associative arrays (such as the one used) have consistent order (but not easily predictable)

scripts/shared/e2e.sh
mkolesnik: After reviewing the scripts at https://github.com/submariner-io/admiral/pull/49 it seems indeed admiral can benefit from this change (well if we make sure it doesn't try to find subm :)) and just employ it's own `deploy` target.

scripts/shared/e2e.sh
tpantelis: But why have this here at all? How does it differ from or is better than a make target dependency?

scripts/shared/e2e.sh
tpantelis: They're passed in a particular order so the code can distinguish them as they may be used for a certain purpose or the code may assume a certain configuration. For submariner, `ClusterC` in the code is used specifically for the add cluster test which assumes no gateway is configured. Lighthouse passes the 3 clusters in numeric order, not sure what assumptions it makes.

I think it's reasonable to use the ordering of the new `clusters` array as the order they will be passed to the test code. So submariner would specify `clusters=(cluster2, cluster3, cluster1)`. Of course then we'll have reconcile positional assumptions in the deploy scripts, eg where it uses `${clusters[1]}` and `${clusters[2]}` for the connectivity test. I assume this is b/c "cluster1" doesn't have submariner running. Actually I'm not clear why we even run the connectivity test on deploy as the E2E tests do that anyway. It seems the scripts should just install/ deploy the clusters and components and let E2E handle the testing. Removing it would certainly simplify the generic scripting.

scripts/shared/e2e.sh
mkolesnik: This was already discussed on https://github.com/submariner-io/submariner/pull/488 (which you approved) and superseded by https://github.com/submariner-io/submariner/pull/510 (which was itself superseded by https://github.com/submariner-io/submariner/pull/512).

The gist is: if it's a `make` dependency, then running `make e2e` would cause a whole chain of actions (compile, docker image, clusters) to occur even if you only intend to run only the e2e tests since you only changed the files there.
Hence this seemed like a nice way to avoid a hard dependency of `e2e` on `deploy` and allow the requested behavior of running just the e2e tests.

I can leave this code solely in submariner, but I think most projects could benefit from it.

If we have `e2e` depend on `delpoy` in the Makefile, it'll be very hard not to rerun `deploy` each time you run `make e2e` since the `deploy` target itself isnt a real file so make can't know if it should or shouldn't re run it.

Hence I proposed this can be toggle-able via a flag to this script.

scripts/shared/e2e.sh
mkolesnik: I think the safest is to receive them in the arguments then each project can define it's own order (since they define their own potential cluster settings so they know best what order to send).

scripts/shared/e2e.sh
mangelajo: Yeah, we need to optimize that make e2e so the workflow of developing a new e2e test is not a PITA for the developer, or involves logging into the dapper shell to run the e2e tests manually or other kind of setup.

Also having a common "e2e" that will let you re-run your e2e tests in a simple way is better because we can have a single development document then.

I see it's not ideal. May be there's a more appropriate name than "lazy" ? :)

scripts/shared/e2e.sh
mangelajo: I agree argument would make sense. Otherwise we need to handle the logic of which one is configured as cluster etc...

e2e_cluster_contexts=(cluster2, cluster3, cluster1)  ? 

something like that

scripts/shared/e2e.sh
```diff
@@ -0,0 +1,78 @@
...
+
+### Main ###
+
+declare_kubeconfig
+
+deploy_env_once
+test_with_e2e_tests
+
+cat << EOM
+Your 3 virtual clusters are deployed and working properly with your local source code, and can be accessed with:
```
mangelajo, line 71: I'd move this to a function we can also call at the end of deploy. (It'd be useful for the "quickstart on kind" guide).

scripts/shared/e2e.sh
mangelajo: Let's try to converge into a single set of e2e scripts, if we need to readjust later... we can figure out.

scripts/shared/e2e.sh
tpantelis: Done by https://github.com/submariner-io/shipyard/pull/113.

PR #116, opened 2020-04-27
Fix double slash in SCRIPTS_DIR file paths

dfarrell07:
The SHIPYARD_DIR variable was defined with a trailing slash to make
Docker copy the directory into the base image correctly, but when
SHIPYARD_DIR was used to build SCRIPTS_DIR this resulted in an extra
slash. Move the slash to the Docker command that requires it, keeping
the var trailing-slash-free as per norms elsewhere in the codebase.


tpantelis:
I noticed this too 😄

dfarrell07:
> Don't know how much of a bug this is, perhaps a slight nuisance :)

Very true lol

PR #117, opened 2020-04-27
Use CLUSTERS_ARGS and DEPLOY_ARGS

mkolesnik:
Instead of using the Make flags that we want to remove, use the ARGS
flags to send the extra script arguments.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,9 +11,13 @@ git:
 jobs:
   include:
   - env: CMD="make validate"
-  - env: CMD="make deploy deploytool=helm globalnet=true"
-  - env: CMD="make deploy deploytool=operator" DEPLOY=true
-  - env: CMD="make deploy cable_driver=wireguard"
+  - env: CMD="make deploy"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -44,9 +48,9 @@ deploy:
     script: make dapper-image dapper_image_flags="--nocache" release release_images="shipyard-dapper-base"
     on:
       branch: master
-      condition: $DEPLOY = true
+      condition: $RELEASE = true
```
dfarrell07, line 51: Good change, I was thinking recently about how this name is getting overloaded/confusing.

mkolesnik:
> This ends up running
> 
> ```
> [00:10:04.987209] /opt/shipyard/scripts/deploy.sh --globalnet false --deploytool operator --cable_driver '' --globalnet true --deploytool helm --cluster_settings /go/src/github.com/submariner-io/shipyard/scripts/cluster_settings
> ```
> 
> but since the last setting wins, everything works out.

Yeah I know, it's temporary until we remove usage in all projects and can remove these variables from Makefile.inc

PR #118, opened 2020-04-28
Make globalnet a boolean flag

mkolesnik:
Since the flag is actually boolean, make it a boolean flag which makes
more sense.
As part of this change we can't pass it in the Makefil recipe and it has
to be passed in the ARGS.

### DIFF
--- a/Makefile.inc
+++ b/Makefile.inc
@@ -1,5 +1,4 @@
 k8s_version ?= 1.14.6
-globalnet ?= false
 deploytool ?= operator
 release_tag ?= latest
 repo ?= quay.io/submariner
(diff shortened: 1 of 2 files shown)


mkolesnik:
This should be fine to go in once https://github.com/submariner-io/submariner-operator/pull/339 is merged, as it's the last project using the Makefile variable `globalnet`

dfarrell07:
> This should be fine to go in once [submariner-io/submariner-operator#339](https://github.com/submariner-io/submariner-operator/pull/339) is merged, as it's the last project using the Makefile variable `globalnet`

It's merged now, thanks.

PR #121, opened 2020-04-28
Add ability to define desired clusters

mkolesnik:
You can defined the clusters in the cluster settings and only those will
be deployed.


mkolesnik:
Closing in favor of #113 which adopted the basic idea of the change

PR #123, opened 2020-04-29
Add basic connectivity E2E tests

tpantelis:
Run a simple pod-to-pod connectivity test within a single cluster.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@ ifneq (,$(DAPPER_HOST_ARCH))
 CLUSTER_SETTINGS_FLAG = --cluster_settings $(DAPPER_SOURCE)/scripts/cluster_settings
 override CLUSTERS_ARGS += $(CLUSTER_SETTINGS_FLAG)
 override DEPLOY_ARGS += $(CLUSTER_SETTINGS_FLAG)
-override E2E_ARGS += $(CLUSTER_SETTINGS_FLAG) cluster1 cluster2
+override E2E_ARGS += $(CLUSTER_SETTINGS_FLAG) --nolazy_deploy cluster1
 
 include $(SHIPYARD_DIR)/Makefile.inc
 
@@ -23,7 +23,7 @@ deploy: nettest
 nettest:
 	$(SCRIPTS_DIR)/build_image.sh -i nettest -f package/Dockerfile.nettest
 
-e2e: vendor/modules.txt
+e2e: vendor/modules.txt clusters
 
 else
 
--- a/test/e2e/framework/test_context.go
+++ b/test/e2e/framework/test_context.go
@@ -59,7 +59,7 @@ func ValidateFlags(t *TestContextType) {
 		klog.Fatalf("kubeconfig parameter or KUBECONFIG environment variable is required")
 	}
 
-	if len(t.KubeContexts) < 2 && len(t.KubeConfigs) < 2 {
-		klog.Fatalf("several kubernetes contexts are necessary east, west, etc..")
+	if len(t.KubeContexts) < 1 && len(t.KubeConfigs) < 1 {
+		klog.Fatalf("at least one kubernetes context must be specified.")
 	}
 }
--- a/test/e2e/dataplane/tcp_connectivity.go
+++ b/test/e2e/dataplane/tcp_connectivity.go
@@ -1,0 +1,25 @@
+package dataplane
+
+import (
+	. "github.com/onsi/ginkgo"
+	"github.com/submariner-io/shipyard/test/e2e/framework"
+	"github.com/submariner-io/shipyard/test/e2e/tcp"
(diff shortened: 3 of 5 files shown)


test/e2e/framework/test_context.go
```diff
@@ -59,7 +59,7 @@ func ValidateFlags(t *TestContextType) {
 		klog.Fatalf("kubeconfig parameter or KUBECONFIG environment variable is required")
 	}
 
-	if len(t.KubeContexts) < 2 && len(t.KubeConfigs) < 2 {
-		klog.Fatalf("several kubernetes contexts are necessary east, west, etc..")
+	if len(t.KubeContexts) < 1 && len(t.KubeConfigs) < 1 {
```
mangelajo, line 62: This is a problem in the other uses where we always expect clusterA & clusterB

On other E2E framework usages the user will be able to pass only one context, and it will panic when trying to access the index for clusterB

scripts/cluster_settings
```diff
@@ -1,9 +1,9 @@
 . "${SCRIPTS_DIR}"/lib/source_only
 
 # For shipyard we need a very minimal setup to verify the deployment works
-cluster_nodes['cluster1']="control-plane"
-cluster_nodes['cluster2']="control-plane worker"
-cluster_nodes['cluster3']="control-plane worker"
+clusters=(cluster1)
```
mkolesnik, line 4: I don't think this is enough, we should at least validate `deploy` target works and deploys subm properly (with broker and all the other bells & whistles).

If you really want a super minimal deployment for e2e you can have a e2e_cluster_settings or some such thing.

scripts/kind-e2e/e2e.sh
```diff
@@ -0,0 +1,25 @@
...
+        -dp-context cluster1 \
+        -ginkgo.reportFile ${DAPPER_OUTPUT}/e2e-junit.xml 2>&1 | \
+        tee ${DAPPER_OUTPUT}/e2e-tests.log
+}
+
+### Main ###
+
+declare_kubeconfig
+
+test_with_e2e_tests
```
mkolesnik, line 25: Seems this can use #115 once we get it sorted out :)

test/e2e/framework/test_context.go
tpantelis: I think each user of this API needs to validate the context based on their use case. This code can't assume the number of clusters/contexts and certainly the prior check didn't work with only one cluster. So here we just make a basic check that at least 1 context must be specified.

tpantelis: So shipyard e2e tests only need 1 cluster; admiral 2.

scripts/cluster_settings
tpantelis: yes - I was thinking the same thing when I realized the CI tests various deploy options. I plan to have 2 `cluster_settings`: one with 3 clusters for CI as it was before and the other with 1 cluster for e2e.

tpantelis: Done

scripts/kind-e2e/e2e.sh
tpantelis: probably - this is minimal.

test/e2e/framework/test_context.go
mangelajo: Yeah, that would make sense.

scripts/shared/e2e.sh
```diff
@@ -61,7 +60,7 @@ function test_with_e2e_tests {
 
 ### Main ###
 
-declare_kubeconfig
 [[ "${lazy_deploy}" = "false" ]] || deploy_env_once
+declare_kubeconfig
```
mkolesnik, line 64: I suspect this change could break consuming projects, since if you dont declare the kubeconfig before checking the lazy deploy thing, it wont ever find an env even if it is deployed.
Have you checked if other projects still behave as expected with this change present?

tpantelis: I wasn't aware of this.  I moved it b/c it yields an error "_find: '/go/src/github.com/submariner-io/shipyard/output/kubeconfigs': No such file or directory_" if no clusters exist prior to the check. But it'd benign - I'll undo this change to be safe.

mkolesnik: Yeah if it wasnt deployed you'll see this printed, we can fix that separately though

Makefile
```diff
@@ -23,6 +23,8 @@ deploy: nettest
 nettest:
 	$(SCRIPTS_DIR)/build_image.sh -i nettest -f package/Dockerfile.nettest
 
+e2e: clusters
+
 e2e: vendor/modules.txt
```
mkolesnik, line 28:
```suggestion
e2e: vendor/modules.txt clusters
```

PR #125, opened 2020-04-30
Use boolean globalnet flag in deploy.sh also

mkolesnik:
Forgot to change it here, use a boolean flag.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ jobs:
   include:
   - env: CMD="make validate"
   - env: CMD="make deploy"
-         CLUSTERS_ARGS="--globalnet true"
+         CLUSTERS_ARGS="--globalnet"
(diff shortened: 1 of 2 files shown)


dfarrell07:
> CMD="make deploy" DEPLOY_ARGS="--cable_driver wireguard"

Job failed, but @sridhargaddam was hacker-fast on the restart before I could copy the logs. :wink: Maybe I should raise an issue to track down these random failures?

mkolesnik:
> > CMD="make deploy" DEPLOY_ARGS="--cable_driver wireguard"
> 
> Job failed, but @sridhargaddam was hacker-fast on the restart before I could copy the logs. wink Maybe I should raise an issue to track down these random failures?

Its possible though its hard to pinpoint these random failures.. Most of the time theyre not related to Shipyard itself

PR #126, opened 2020-04-30
Save more space by some more UPX

mkolesnik:
Seems we can safetly UPX the executables bigger that 1MB.
Also there's a bunch of executables scattered around /usr/libexec so
added that to the UPX process.

Overall this saves around 70MB on the image size.


mkolesnik:
OK, I rebased this on top of #147 so now the resulting image is 805MB :D 
Of course this is bound to change with time, but I think its quite descent ;)

Also I ran E2E tests on Subm, LH & Operator repos and they all worked so I think this change should be safe enough.

PR #127, opened 2020-04-30
Added cluster funcs that abstract details

mkolesnik:
Instead of calling run_parallel it's now possible to call
run_all_clusters or run_subm_clusters to execute on all the clusters or
just the ones that have subm installed, respectively.

Secret knowledge about specific deployment constraints is now part of
the called functions themselves (e.g. deploy_weave_cni).

### DIFF
--- a/scripts/shared/clusters.sh
+++ b/scripts/shared/clusters.sh
@@ -83,6 +83,11 @@ function deploy_weave_cni(){
         return
     fi
 
+    if [[ "${cluster}" = "cluster1" ]]; then
+       echo "Not deploying weave on broker cluster"
+       return
+    fi
+
     echo "Applying weave network..."
     kubectl apply -f "https://cloud.weave.works/k8s/net?k8s-version=v$version&env.IPALLOC_RANGE=${cluster_CIDRs[${cluster}]}"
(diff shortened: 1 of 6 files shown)


scripts/shared/clusters.sh
```diff
@@ -114,7 +119,7 @@ mkdir -p ${KUBECONFIGS_DIR}
 
 run_local_registry
 declare_cidrs
-with_retries 3 run_parallel "{1..3}" create_kind_cluster
+with_retries 3 run_all_clusters create_kind_cluster
 declare_kubeconfig
-run_parallel "2 3" deploy_weave_cni
+run_subm_clusters deploy_weave_cni
```
tpantelis, line 124: This doesn't look quite equivalent to prior behavior for submariner. `run_subm_clusters` will run it on any cluster where `cluster_subm[X]` is true. Prior behavior hard-coded to omit `cluster1`. However cluster_settings has `cluster_subm['cluster1']` set to true so now weave will be deployed there.

tpantelis: Oh I see you ignore "cluster1" in `deploy_weave_cni`.

PR #128, opened 2020-05-03
Remove run_sequential function

mkolesnik:
This was useful for globalnet workaround, but since we've added support
to deploy in parallel with globalnet this isn't used anymore, so
removing it.

### DIFF
--- a/scripts/shared/lib/utils
+++ b/scripts/shared/lib/utils
@@ -89,19 +89,6 @@ function run_subm_clusters() {
     run_parallel "${subm_clusters[*]}" "$@"
 }
 
-# Run cluster commands sequentially.
-# 1st argument is the numbers of the clusters to run for, supports "1 2 3" or "{1..3}" for range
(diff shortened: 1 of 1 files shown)


mangelajo:
I wonder if we could make use of this to make GH actions an option. By having a global flag that lets you select sequential mode.

What do you think?

mkolesnik:
> I wonder if we could make use of this to make GH actions an option. By having a global flag that lets you select sequential mode.
> 
> What do you think?

GH actions is being silly and i dont think the parallel deployment is whats causing it,

E.G. I see right now this
```
default netshoot-785ffd8c8-2zq2r 0/1 CreateContainerError 0 2m6s
```

I doubt deploying in parallel caused netshoot pod to go into *CreateContainerError* state..
Let's debug GH action and get to the root of it, it can't be that travis works well (most of the time) but it doesn't if they both use VMs.

mkolesnik:
Closing since its being used in post mortem (PR #132)

PR #129, opened 2020-05-03
Test GH Actions

mkolesnik:


mangelajo:
I wonder if enforcing run_sequential would make GH actions workable, although slower. 

The system has the same amount of RAM as travis, but I suspect that much worse disk I/O

mangelajo:
:1st_place_medal:

mkolesnik:
Closing in favor of #141

PR #130, opened 2020-05-04
More robust cluster creation & retries

mkolesnik:
These commits fix the retries mechanism (it didn't really adhere `set -e` because of how that mechanism works), and move the CNI creation logic into the cluster creation function so that in case it fails (sometimes weave deploy is flaky), the script will destroy the cluster and re create it.

This should hopefully alleviate some of the failures in https://github.com/submariner-io/submariner/issues/532

### DIFF
--- a/scripts/shared/clusters.sh
+++ b/scripts/shared/clusters.sh
@@ -78,7 +78,7 @@ function create_kind_cluster() {
 }
 
 function deploy_weave_cni(){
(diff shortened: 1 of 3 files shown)


mangelajo:
This kills deployment in some laptops, can we revert? :)

mkolesnik:
> This kills deployment in some laptops, can we revert? :)

I don't think that's a good solution either, lets work out something and implement it

PR #131, opened 2020-05-04
Export kubeconfigs on dapper image

mkolesnik:
Always export it so that when connecting with an interactive shell you'll
already have it primed with the correct kubeconfigs.


scripts/shared/lib/kubecfg
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+export KUBECONFIG
+KUBECONFIG=$(find "${DAPPER_OUTPUT:-output}"/kubeconfigs -type f | tr '\n' ':')
```
skitt, line 3:
```suggestion
KUBECONFIG=$(find "${DAPPER_OUTPUT:-output}"/kubeconfigs -type f -printf %p:)
```

PR #132, opened 2020-05-05
Add post mortem script

mkolesnik:
In case of CI failure, this will retrieve information for debugging the
failure.


mkolesnik:
You can see post mortem of a failed job here - [log.txt](https://github.com/submariner-io/shipyard/files/4580139/log.txt)

PR #133, opened 2020-05-05
Clean out outputs on cleanup

mkolesnik:
There's a lot of stale files there, clean it out as part of cleanup

### DIFF
--- a/scripts/shared/cleanup.sh
+++ b/scripts/shared/cleanup.sh
@@ -24,6 +24,7 @@ function stop_local_registry {
 ### Main ###
 
 run_all_clusters delete_cluster
+[[ -z "${DAPPER_OUTPUT}" ]] || rm -rf ${DAPPER_OUTPUT}/*
 stop_local_registry
 docker system prune --volumes -f
 


scripts/shared/cleanup.sh
```diff
@@ -24,6 +24,7 @@ function stop_local_registry {
 ### Main ###
 
 run_all_clusters delete_cluster
+rm -rf ${DAPPER_OUTPUT}/*
```
mangelajo, line 27: Can we add a check to make sure $DAPPER_OUTPUT is not empty? 
```suggestion
if [ "${DAPPER_OUTPUT}" != "" ]; then
    rm -rf ${DAPPER_OUTPUT}/*
fi    
```
I was imagining running "cleanup.sh" from bash, by mistake :D

mkolesnik: Makes sense, thanks!

mangelajo:
@mkolesnik I believe you need a manual rebase:


> **This branch cannot be rebased due to conflicts**
>
> Rebasing the commits of this branch on top of the base branch cannot be performed automatically
> due to conflicts encountered while reapplying the individual commits from the head branch.

PR #145, opened 2020-05-08
Don't pass the make binary path down to dapper

mangelajo:
The host could be different from linux or have a weird path for make.

Fixes: submariner-io/#555

### DIFF
--- a/Makefile.dapper
+++ b/Makefile.dapper
@@ -15,7 +15,7 @@
 MAKEOVERRIDES =
 
 %: .dapper
-	./.dapper -m bind make $@ $(MAKEFLAGS)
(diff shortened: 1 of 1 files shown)


Makefile.dapper
```diff
@@ -15,7 +15,7 @@
 MAKEOVERRIDES =
 
 %: .dapper
-	./.dapper -m bind $(MAKE) $@ $(MAKEFLAGS)
+	./.dapper -m bind make $@ $(MAKEFLAGS)
```
skitt, line 18:
```suggestion
	+./.dapper -m bind make $@ $(MAKEFLAGS)
```

See https://github.com/submariner-io/submariner/pull/556#pullrequestreview-408731846

PR #147, opened 2020-05-10
Remove any selinux policy packages

mkolesnik:
We're not using selinux in the container, remove any policy package to
save some space..

This specifically shaves off some 30MB :)

### DIFF
--- a/package/Dockerfile.dapper-base
+++ b/package/Dockerfile.dapper-base
@@ -39,7 +39,7 @@ RUN dnf -y distrosync --nodocs --setopt=install_weak_deps=False && \
                    gcc git-core curl moby-engine make golang kubernetes-client \
                    findutils upx jq golang-x-tools-goimports ShellCheck shflags && \
     dnf -y remove libsemanage && \
-    rpm -qa "selinux*" | xargs rpm -r -e --nodeps && \
+    rpm -qa "selinux*" | xargs -r rpm -e --nodeps && \
     dnf -y clean all && \
     curl -L https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s -- -b $(go env GOPATH)/bin -d ${LINT_VERSION} && \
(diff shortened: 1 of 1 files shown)


package/Dockerfile.dapper-base
```diff
@@ -39,7 +39,7 @@ RUN dnf -y distrosync --nodocs --setopt=install_weak_deps=False && \
                    gcc git-core curl moby-engine make golang kubernetes-client \
                    findutils upx jq golang-x-tools-goimports ShellCheck shflags && \
     dnf -y remove libsemanage && \
-    rpm -e --nodeps selinux-policy-targeted && \
+    rpm -qa "selinux*" | xargs rpm -e --nodeps && \
```
skitt, line 42:
```suggestion
    rpm -qa "selinux*" | xargs rpm -r -e --nodeps && \
```

so this doesn’t fail if `rpm -qa` doesn’t output anything ;-).

mkolesnik: Are you sure it's `-r`?

It seems to fail with
```
rpm: arguments to --root (-r) must begin with a /
```

skitt: Oops, sorry, wrong place, it’s meant for `xargs`: `xargs -r rpm -e --nodeps && \`

package/Dockerfile.dapper-base
```diff
@@ -39,7 +39,7 @@ RUN dnf -y distrosync --nodocs --setopt=install_weak_deps=False && \
                    gcc git-core curl moby-engine make golang kubernetes-client \
                    findutils upx jq golang-x-tools-goimports ShellCheck shflags && \
     dnf -y remove libsemanage && \
-    rpm -qa "selinux*" | xargs rpm -e --nodeps && \
+    rpm -qa "selinux*" | xargs rpm -r -e --nodeps && \
```
skitt, line 42:
```suggestion
    rpm -qa "selinux*" | xargs -r rpm -e --nodeps && \
```

Sorry, this is the correct invocation.

mkolesnik: OK we'll just squash it before merge :)

PR #148, opened 2020-05-11
Shared compilation script

mkolesnik:
Adding a shared compilation script since it is copied over to different
repositories.
The script is quite flexible to allow consuming repositories to adapt it
with their different requirements.


skitt:
The builds fail because install.goreleaser.com’s certificate isn’t valid.

scripts/shared/compile.sh
```diff
@@ -0,0 +1,39 @@
...
+
+mkdir -p ${binary%/*}
+
+echo "Building ${binary@Q} (ldflags: ${ldflags@Q})"
+if [ "$build_debug" = "false" ]; then
+    ldflags="-s -w ${ldflags}"
+fi
+
+CGO_ENABLED=0 go build -ldflags "${ldflags}" -o $binary $source_file
+[[ "$build_upx" = "false" ]] || upx $binary
```
skitt, line 38: I think it would be better to skip `upx` for debug builds too.

mkolesnik: I was leaving it to the caller but I guess it wont hurt to have it hardcoded that way

PR #149, opened 2020-05-12
Remove GCC and related packages

mkolesnik:
Go can build/test on it's own without GCC.
This frees up around 80MB of the image size.


mkolesnik:
:crossed_fingers:

PR #156, opened 2020-05-12
Add --timeout flag to scripts

mkolesnik:
Seems that on CI a small timeout works fine, but on personal laptops a
bigger one is needed (due to constrained bandwidth).

Added a --timemout flag with default of 5m for clusters and deploy
scripts, and used saner value for the CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,9 +12,11 @@ jobs:
   include:
   - env: CMD="make validate"
   - env: CMD="make deploy"
-         CLUSTERS_ARGS="--globalnet"
-         DEPLOY_ARGS="${CLUSTERS_ARGS} --deploytool helm"
+         CLUSTERS_ARGS="--globalnet --timeout 1m"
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -12,9 +12,11 @@ jobs:
   include:
   - env: CMD="make validate"
   - env: CMD="make deploy"
-         CLUSTERS_ARGS="--globalnet"
-         DEPLOY_ARGS="${CLUSTERS_ARGS} --deploytool helm"
+         CLUSTERS_ARGS="--globalnet --timeout 1m"
+         DEPLOY_ARGS="--globalnet --deploytool helm --timeout 2m"
```
sridhargaddam, line 16: You are passing "globalnet=true" to DEPLOY_ARGS as well. Is this intentional?

mkolesnik: If you notice it was `${CLUSTERS_ARGS}` before so it was already getting it, but since i don't want to have "conflicting" timeouts I opted to specify it explicitly

PR #157, opened 2020-05-12
Adding slim nettest image

mkolesnik:
This image has the bare essentials we need for testing connectivity in
a simulated environment.

### DIFF
--- a/scripts/shared/build_image.sh
+++ b/scripts/shared/build_image.sh
@@ -36,7 +36,7 @@ latest_image=${repo}/${image}:latest
 cache_flag=''
 if [[ "$cache" = true ]]; then
     cache_flag="--cache-from ${latest_image}"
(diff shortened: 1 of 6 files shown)


scripts/shared/build_image.sh
```diff
@@ -36,7 +36,7 @@ latest_image=${repo}/${image}:latest
 cache_flag=''
 if [[ "$cache" = true ]]; then
     cache_flag="--cache-from ${latest_image}"
-    docker pull ${latest_image}
+    docker pull ${latest_image} || :
```
skitt, line 39: Hah yes when we introduce a new image, this would have failed ;-).

mangelajo: hehe

mangelajo: I meant thumbs up... ':)

PR #163, opened 2020-05-13
Use nettest image for netshoot pod

mkolesnik:
Since PR #157 added a slim "netshoot" image, use that for the
framework's testing instead of the full blown one.

### DIFF
--- a/test/e2e/framework/deployments.go
+++ b/test/e2e/framework/deployments.go
@@ -47,7 +47,7 @@ func (f *Framework) NewNetShootDeployment(cluster ClusterIndex) *corev1.PodList 
 					Containers: []corev1.Container{
 						{
 							Name:            "netshoot",
-							Image:           "nicolaka/netshoot",
+							Image:           "quay.io/submariner/nettest",
 							ImagePullPolicy: corev1.PullAlways,
 							Command: []string{
 								"sleep", "600",


test/e2e/framework/deployments.go
```diff
@@ -47,7 +47,7 @@ func (f *Framework) NewNetShootDeployment(cluster ClusterIndex) *corev1.PodList
 					Containers: []corev1.Container{
 						{
 							Name:            "netshoot",
```
skitt, line 49: Can we change the container name too, or would that break the tests?

mkolesnik: I don't know since theres no local tests that use it.
Anyhow I don't think it's critical we didn't change the name of the yaml resource.. Maybe a nice to have..

mangelajo: I want to get rid of this, we can just use the standard connectivity framework, or.. in the case where this is used (lighthouse) just dns queries, we don't actually need to check connectivity there

mkolesnik: OK so until you get rid of it, shouldnt we at least use the smaller image?

mangelajo: Yes yes, I'm fine with it, let's leave the name alone just in case, I didn't look at the implementation (that's what I mean, that I don't have time to look at it specially if I plan to remove it all together)

PR #165, opened 2020-05-13
Fix post_mortem script and add a test for it

mkolesnik:
Fixed the script to work after PR #113 changed the way run_ functions
work.

Also added a simple test to verify that the post_mortem itself is
working.


mangelajo:
Thanks mike !

PR #169, opened 2020-05-19
Allow deploying two clusters

### LINKED ISSUE #138: Support 2 cluster deployments
Currently if deploying 2 clusters then connectivity isn't checked. Support deploying just 2 clusters and have connectivity check if they have submariner.
This is related to #136 since if we can deploy broker where submariner is deployed, we can save a cluster.

Shipyard & operator (and perhaps lighthouse) stand to benefit from this.

mkolesnik:
Allow deploying only two clusters (with broker running on one of them).
This will benefit projects such as Shipyard, Lighthouse and others that
don't really need a 3 cluster setup.

The default remains 3 clusters, but now its possible to override to only
deploy 2.

Closes #138

### DIFF
--- a/scripts/shared/lib/deploy_helm
+++ b/scripts/shared/lib/deploy_helm
@@ -88,8 +88,3 @@ function helm_install_subm() {
 function install_subm_all_clusters() {
     run_subm_clusters helm_install_subm
 }
-
-function deploytool_postreqs() {
-    # This function must exist for parity with Operator deploys, but does nothing for Helm
-    :
-}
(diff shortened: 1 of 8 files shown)


scripts/shared/lib/deploy_helm
```diff
@@ -88,8 +88,3 @@ function helm_install_subm() {
 function install_subm_all_clusters() {
     run_subm_clusters helm_install_subm
 }
-
-function deploytool_postreqs() {
```
dfarrell07, line 92: Yay for this old code of mine going away. :smile:

PR #173, opened 2020-05-20
Validate globalIP annotation for Pods when using globalnet

sridhargaddam:
In this PR, we validate the globalIP annotation on the connectorPod
as part of e2e tests.

Partially fixes# https://github.com/submariner-io/submariner/issues/544

### DIFF
--- a/test/e2e/framework/pods.go
+++ b/test/e2e/framework/pods.go
@@ -44,8 +44,8 @@ func (f *Framework) DeletePod(cluster ClusterIndex, podName string, namespace st
 	}, NoopCheckResult)
 }
 
-// AwaitPodByAnnotation queries the Pod and looks for the presence of annotation.
-func (f *Framework) AwaitPodByAnnotation(cluster ClusterIndex, annotation string, podName string, namespace string) *v1.Pod {
+// AwaitUntilAnnotationOnPod queries the Pod and looks for the presence of annotation.
+func (f *Framework) AwaitUntilAnnotationOnPod(cluster ClusterIndex, annotation string, podName string, namespace string) *v1.Pod {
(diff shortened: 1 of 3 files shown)


test/e2e/framework/pods.go
```diff
@@ -42,3 +43,24 @@ func (f *Framework) DeletePod(cluster ClusterIndex, podName string, namespace st
 		return nil, KubeClients[cluster].CoreV1().Pods(namespace).Delete(podName, &metav1.DeleteOptions{})
 	}, NoopCheckResult)
 }
+
+// AwaitPodByAnnotation queries the Pod and looks for the presence of annotation.
+func (f *Framework) AwaitPodByAnnotation(cluster ClusterIndex, annotation string, podName string, namespace string) *v1.Pod {
```
vthapar, line 48: AwaitUntilAnnotationOnPod() sounds like better name. AwaitPodByAnnotation() suggests we're waiting for multiple pods to be spawned with annotation used as a filter.

mangelajo: AwaitUntilAnnotationOnPod sounds like the pod is pre-existing and you are only waiting for the annotation.

mangelajo: Which is probably what we wanted, lol

mangelajo: Well but the code is waiting for a pod to appear, and the annotation to be on the pod.

What about : AwaitPodWithAnnotation ?

sridhargaddam: Mmm.. I used the current name to be aligned with the similar name we used for Service. But I agree the name is confusing and I'm a bit inclined to @vthapar's suggestion - i.e., AwaitUntilAnnotationOnPod.
Also changing the name of the service to keep them in sync.

PR #174, opened 2020-05-20
Don't use 'docker system prune' for reclaiming space

mkolesnik:
Sometimes it takes more than a minute to run, turning swap off should give us enough headroom

### DIFF
--- a/.github/workflows/scripts.yaml
+++ b/.github/workflows/scripts.yaml
@@ -46,10 +46,8 @@ jobs:
 
       - name: Reclaim free space!
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/scripts.yaml
```diff
@@ -60,7 +58,9 @@ jobs:
 
       - name: Post Mortem
         if: failure()
-        run: make post-mortem
+        run: |
+          df -h
+          make post-mortem
```
skitt, line 63: This one doesn’t have `free -h`.

PR #175, opened 2020-05-20
Add OLM install support to cluster deployments

dfarrell07:
Add support for optionally installing the Operator Lifecycle Manager in
clusters deployed by Shipyard.

After deploying OLM, a released version of the Submariner Operator can
be installed with a simple:

kubectl create -f https://dev.operatorhub.io/install/submariner.yaml

Relates-to: submariner-io/submariner-operator#373

### DIFF
--- a/scripts/shared/clusters.sh
+++ b/scripts/shared/clusters.sh
@@ -4,6 +4,8 @@
 
 source ${SCRIPTS_DIR}/lib/shflags
 DEFINE_string 'k8s_version' '' 'Version of K8s to use'
+DEFINE_string 'olm_version' '0.14.1' 'Version of OLM to use'
+DEFINE_boolean 'olm' false 'Deploy OLM'
 DEFINE_boolean 'globalnet' false "Deploy with operlapping CIDRs (set to 'true' to enable)"
 DEFINE_boolean 'registry_inmemory' true "Run local registry in memory to speed up the image loading."
 DEFINE_string 'cluster_settings' '' "Settings file to customize cluster deployments"
@@ -12,11 +14,13 @@ FLAGS "$@" || exit $?
 eval set -- "${FLAGS_ARGV}"
 
 version="${FLAGS_k8s_version}"
+olm_version="${FLAGS_olm_version}"
+[[ "${FLAGS_olm}" = "${FLAGS_TRUE}" ]] && olm=true || olm=false
 [[ "${FLAGS_globalnet}" = "${FLAGS_TRUE}" ]] && globalnet=true || globalnet=false
 [[ "${FLAGS_registry_inmemory}" = "${FLAGS_TRUE}" ]] && registry_inmemory=true || registry_inmemory=false 
 cluster_settings="${FLAGS_cluster_settings}"
 timeout="${FLAGS_timeout}"
-echo "Running with: k8s_version=${version}, globalnet=${globalnet}, registry_inmemory=${registry_inmemory}, cluster_settings=${cluster_settings}, timeout=${timeout}"
+echo "Running with: k8s_version=${version}, olm_version=${olm_version}, olm=${olm}, globalnet=${globalnet}, registry_inmemory=${registry_inmemory}, cluster_settings=${cluster_settings}, timeout=${timeout}"
 
 set -em
 
@@ -85,9 +89,9 @@ function create_kind_cluster() {
     kind create cluster $image_flag --name=${cluster} --config=${RESOURCES_DIR}/${cluster}-config.yaml
     kind_fixup_config
 
-    ( deploy_cni; ) &
+    ( deploy_cluster_capabilities; ) &
     if ! wait $! ; then
-        echo "Failed to deploy custom CNI, removing the cluster"
+        echo "Failed to deploy cluster capabilities, removing the cluster"
         kind delete cluster --name=${cluster}
         return 1
     fi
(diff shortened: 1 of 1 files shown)


Makefile.inc
```diff
@@ -1,4 +1,6 @@
 k8s_version ?= 1.14.6
+olm_version ?= 0.14.1
+olm ?= false
```
skitt, line 3: I think the idea now is to rely only on `CLUSTERS_ARGS` instead of adding Make variables. @mkolesnik what do you think?

dfarrell07: Done, thanks. Should `k8s_version` be in `CLUSTER_ARGS`? That's what confused me.

scripts/shared/clusters.sh
```diff
@@ -4,6 +4,8 @@
 
 source ${SCRIPTS_DIR}/lib/shflags
 DEFINE_string 'k8s_version' '' 'Version of K8s to use'
+DEFINE_string 'olm_version' '0.14.1' 'Version of OLM to use'
+DEFINE_string 'olm' 'false' 'Deploy OLM'
```
mkolesnik, line 8:
```suggestion
DEFINE_boolean 'olm' false 'Deploy OLM'
```

scripts/shared/clusters.sh
```diff
@@ -12,11 +14,13 @@ FLAGS "$@" || exit $?
 eval set -- "${FLAGS_ARGV}"
 
 version="${FLAGS_k8s_version}"
+olm_version="${FLAGS_olm_version}"
+olm="${FLAGS_olm}"
```
mkolesnik, line 18:
```suggestion
[[ "${FLAGS_olm}" = "${FLAGS_TRUE}" ]] && olm=true || olm=false
```

scripts/shared/clusters.sh
```diff
@@ -91,6 +95,10 @@ function create_kind_cluster() {
         kind delete cluster --name=${cluster}
         return 1
     fi
+
+    if [[ $olm = "true" ]]; then
+      deploy_olm
+    fi
```
mkolesnik, line 101:
```suggestion
    [[ $olm != "true" ]] || deploy_olm
```

I prefer this syntax (alternatively, you can stick it in deploy_olm right at the start).

Should the OLM deployment fail though, would you want to retry the cluster (as in deploy_cni above)?

scripts/shared/clusters.sh
```diff
@@ -123,6 +131,19 @@ function run_local_registry() {
...
 }
 
+function deploy_olm() {
+    echo "Applying OLM CRDs..."
+    kubectl apply -f https://github.com/operator-framework/operator-lifecycle-manager/releases/download/$olm_version/crds.yaml --validate=false
+    echo "Applying OLM resources..."
+    kubectl apply -f https://github.com/operator-framework/operator-lifecycle-manager/releases/download/$olm_version/olm.yaml
+
+    echo "Waiting for olm-operator deployment to be ready..."
+    kubectl rollout status -w deployment/olm-operator --namespace=olm
```
mkolesnik, line 141:
```suggestion
    kubectl rollout status deployment/olm-operator --namespace=olm --timeout="${timeout}"
```

scripts/shared/clusters.sh
```diff
@@ -123,6 +131,19 @@ function run_local_registry() {
...
+    echo "Waiting for catalog-operator deployment to be ready..."
+    kubectl rollout status -w deployment/catalog-operator --namespace=olm
```
mkolesnik, line 143:
```suggestion
    kubectl rollout status deployment/catalog-operator --namespace=olm --timeout="${timeout}"
```

scripts/shared/clusters.sh
```diff
@@ -123,6 +131,19 @@ function run_local_registry() {
...
+    echo "Waiting for packageserver deployment to be ready..."
+    kubectl rollout status -w deployment/packageserver --namespace=olm
```
mkolesnik, line 145:
```suggestion
    kubectl rollout status deployment/packageserver --namespace=olm --timeout="${timeout}"
```

scripts/shared/clusters.sh
```diff
@@ -4,6 +4,8 @@
 
 source ${SCRIPTS_DIR}/lib/shflags
 DEFINE_string 'k8s_version' '' 'Version of K8s to use'
+DEFINE_string 'olm_version' '0.14.1' 'Version of OLM to use'
```
mkolesnik, line 7:
```suggestion
DEFINE_string 'olm_version' '0.15.0' 'Version of OLM to use'
```
FYI 0.15.0 was just released

scripts/shared/clusters.sh
dfarrell07: Humm, good question. I haven't seen it fail yet, and it seems like it should be more reliable than Weave, so I'm not sure retries are super important. Idk though, I'm open to recommendations.

Makefile.inc
dfarrell07: @mkolesnik sent a patch to fix this here https://github.com/submariner-io/shipyard/pull/177. Thanks for clarifying!

dfarrell07:
-2 for now, it seems moving to 0.15.0 wasn't as simple as a version bump.

```
 unable to recognize "https://github.com/operator-framework/operator-lifecycle-manager/releases/download/0.15.0/crds.yaml": no matches for kind "CustomResourceDefinition" in version "apiextensions.k8s.io/v1"
```

scripts/shared/clusters.sh
mkolesnik: Well just because it hasn't happened yet doesn't mean it can't :)

Perhaps have a function 'deploy_cluster_capabilities'  or something, which will be called in line 88 (instead of deploy_cni) and in turn it'll call `deploy_cni` and `deploy_olm`.

This way if that function fails, it would destroy the cluster and retry..

skitt:
> -2 for now, it seems moving to 0.15.0 wasn't as simple as a version bump.
> 
> ```
>  unable to recognize "https://github.com/operator-framework/operator-lifecycle-manager/releases/download/0.15.0/crds.yaml": no matches for kind "CustomResourceDefinition" in version "apiextensions.k8s.io/v1"
> ```

OLM 0.15 requires K8s 1.16 or later.

dfarrell07:
Okay, back to 0.14.1 then.

I wonder what's up with these 0.14.2 and 0.15.1 that don't have the normal artifacts?

https://github.com/operator-framework/operator-lifecycle-manager/releases/tag/0.14.2
https://github.com/operator-framework/operator-lifecycle-manager/releases/tag/0.15.1

The 0.14.2 one says it was tagged in February, but I don't remember seeing it when I was looking at the releases page last week. :man_shrugging:

scripts/shared/clusters.sh
dfarrell07: @mkolesnik okay, good ideas, done in the latest. :)

scripts/shared/clusters.sh
```diff
@@ -85,12 +89,13 @@ function create_kind_cluster() {
...
 
-    ( deploy_cni; ) &
+    ( deploy_cluster_capabilities; ) &
     if ! wait $! ; then
-        echo "Failed to deploy custom CNI, removing the cluster"
+        echo "Failed to deploy cluster capabilities, removing the cluster"
         kind delete cluster --name=${cluster}
         return 1
     fi
+
```
mkolesnik, line 98:
```suggestion
```
Think you left this here ;)

dfarrell07: Lol, yes, that was my litter. Cleaned it up in the latest. ;)

PR #180, opened 2020-05-21
Add GH release workflow

mkolesnik:
This is a proposed workflow for GH actions for releasing the base image
and nettest image.


.github/workflows/release.yaml
```diff
@@ -0,0 +1,37 @@
+on:
+  push:
+    branches:
+      - master
+    tags:
+      - 'v**'
```
mangelajo, line 6:
```suggestion
on:
  release:
    types: [created]
```

This is triggered by tags, etc too.

mkolesnik: Are you sure that event makes sense?

I thought it's for actual releases - e.g. https://github.com/submariner-io/submariner-operator/releases

In such a case it will *only* trigger for tags but never for merge to the master branch..

Please correct me if I'm understanding this wrong

mangelajo: I think you're right. I forgot of master

PR #181, opened 2020-05-21
No need to reclaim free space

mkolesnik:
After testing this, we have ample free space in shipyard's jobs

### DIFF
--- a/.github/workflows/scripts.yaml
+++ b/.github/workflows/scripts.yaml
@@ -44,12 +44,6 @@ jobs:
     steps:
       - uses: actions/checkout@master
 
-      - name: Reclaim free space!
-        run: |
-          sudo swapoff -a
-          sudo rm -f /swapfile
-          df -h
-
       - name: Run the deploy script
         env:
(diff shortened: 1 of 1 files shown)


.github/workflows/scripts.yaml
```diff
@@ -44,12 +44,6 @@ jobs:
     steps:
       - uses: actions/checkout@master
 
-      - name: Reclaim free space!
```
mangelajo, line 47: Why isn't it needed anymore? :?

dfarrell07: I'd guess the Shipyard jobs use less space than the jobs that build on Shipyard and do more stuff. :man_shrugging:

mkolesnik: I actually found that by default when running KIND, each node (which runs in a docker container) has a volume allocated with the same size as the image, so for our current image it's about 1.3GB per node, so for shipyard that only had 2 clusters with 2 nodes each it's roughly 1.3*2*2=5.2GB needed, which is much less than the 10GB that GH actions runners guarantee (but in reality have even more space since they leave headroom).

Issue #182, opened 2020-05-22
Lighthouse now requires a ServiceExport


mangelajo:
Without that ServiceExport, submariner-operator which smoke-tests the
discovery of the IP will fail.

Related-commit: https://github.com/submariner-io/lighthouse/commit/c3bac6af3358ded33c47b63bb9ff1e269820a062

vthapar:
Added to Lighthouse E2E by https://github.com/submariner-io/lighthouse/pull/137

tpantelis:
It seems this was addressed.

PR #183, opened 2020-05-22
Provide a ServiceExport resource file for other projects to use

mangelajo:
Lighthouse now expects a ServiceExport along with the service
to make the service exported.

Fixes-Issue: #182
Depends-On: https://github.com/submariner-io/shipyard/issues/182


mangelajo:
Ok, following discussion with @vthapar we didn't want to have specific lighthouse stuff in shipyard, so I will move the service export part to the operator

PR #185, opened 2020-05-26
Increase go test timeout to 15 mins

sridhargaddam:
Currently, we are not configuring any timeout while running the e2e
tests, so the default timeout of 10 minutes is used. However, it
was seen that since we are now running more number of tests, at times,
the tests are unable to complete its execution in 10 mins and because
of this the test is abruptly stopped with PANIC. This PR explicitly
configures the timeout to 15 mins to give sufficient time for tests
to run.

Fixes Issue: https://github.com/submariner-io/submariner/issues/593
Partially fixes: https://github.com/submariner-io/submariner/issues/544

### DIFF
--- a/scripts/shared/e2e.sh
+++ b/scripts/shared/e2e.sh
@@ -53,7 +53,7 @@ function test_with_e2e_tests {
 
     go test -v -args -ginkgo.v -ginkgo.randomizeAllSpecs \
         -submariner-namespace $SUBM_NS $(generate_context_flags) \
-        -ginkgo.reportPassed \
+        -ginkgo.reportPassed -test.timeout 15m \
         -ginkgo.focus "\[${focus}\]" \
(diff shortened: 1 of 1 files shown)


mkolesnik:
Probably tests need to be fixed to take less time, this is just a stopgap

sridhargaddam:
> Probably tests need to be fixed to take less time, this is just a stopgap

It's not really a stopgap solution as the timeout is for all the tests (and not for a single test-case). 
When we add more test cases, we anyway have to increase the test.timeout and cant expect all of them to complete in the 10m interval ;-) 

Sometimes a run might take a longer time if the internet bandwidth is not so strong (as it directly affects the time it takes for a Pod to get to a running state). Recently I pushed a WIP PR where I disabled the redundancy tests and I could see that tests ran successfully 8 times - https://github.com/submariner-io/submariner/pull/571#issuecomment-629754565

PR #188, opened 2020-05-27
Use GITHUB_SHA if available

mkolesnik:
To support GitHub Actions release, need to use the SHA from there.

Also added space for tag in case it's needed (otherwise it'll fail)

### DIFF
--- a/scripts/shared/release.sh
+++ b/scripts/shared/release.sh
@@ -3,7 +3,7 @@
 ## Process command line flags ##
 
 source ${SCRIPTS_DIR}/lib/shflags
-DEFINE_string 'tag' '' "Additional tag to use for the image (prefix 'v' will be stripped)"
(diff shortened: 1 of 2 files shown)


scripts/shared/release.sh
```diff
@@ -24,14 +24,16 @@ source ${SCRIPTS_DIR}/lib/version
...
-    local images=("${image}:${TRAVIS_COMMIT:0:7}" "${image}:${release_tag#v}")
+    local images=("${image}:${commit_hash:0:7}" "${image}:${release_tag#v}")
 
     for target_image in "${images[@]}"; do
         docker tag ${image}:${VERSION} ${target_image}
         docker push ${target_image}
     done
 }
 
+commit_hash=${GITHUB_SHA:-${TRAVIS_COMMIT}}
```
mangelajo, line 35: haha, makes sense to keep compatibility with both :)

mkolesnik: For now, until we get rid of travis :)

PR #190, opened 2020-05-27
Pin the dependencies to K8s 1.14.1

skitt:
We need to harmonise dependencies across Submariner projects; in
particular, other projects are pinned to Kubernetes 1.14.1, which is
our target (until we upgrade KIND).

This also bumps to Go 1.13 which other projects use.


mangelajo:
This is for the dependency issue lighthouse/shipyard/submariner/operator, right?

skitt:
> This is for the dependency issue lighthouse/shipyard/submariner/operator, right?

Yes, I’ll update the commit message to explain why I’m suggesting this.

dfarrell07:
> > This is for the dependency issue lighthouse/shipyard/submariner/operator, right?
> 
> Yes, I’ll update the commit message to explain why I’m suggesting this.

Thanks, I was wondering the same.

PR #192, opened 2020-05-27
Use shflags version 1.0.3 by default

mkolesnik:
For now thats what we were using, we can update later

### DIFF
--- a/scripts/shared/lib/shflags
+++ b/scripts/shared/lib/shflags
@@ -9,9 +9,8 @@ ensure_shflags() {
     fi
 
     if [ ! -f .shflags ]; then
-        echo "Downloading shflags ${SHFLAGS_VERSION:=1.2.3}"
-        if ! (curl -L "https://github.com/kward/shflags/archive/v${SHFLAGS_VERSION}.tar.gz" | tar -xzf - shflags-${SHFLAGS_VERSION}/shflags &&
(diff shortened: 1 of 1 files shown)


scripts/shared/lib/shflags
```diff
@@ -9,8 +9,8 @@ ensure_shflags() {
     fi
 
     if [ ! -f .shflags ]; then
-        echo "Downloading shflags ${SHFLAGS_VERSION:=1.2.3}"
-        if ! (curl -L "https://github.com/kward/shflags/archive/v${SHFLAGS_VERSION}.tar.gz" | tar -xzf - shflags-${SHFLAGS_VERSION}/shflags &&
+        echo "Downloading shflags ${SHFLAGS_VERSION}"
+        if ! (curl -L "https://github.com/kward/shflags/archive/${SHFLAGS_VERSION}.tar.gz" | tar -xzf - shflags-${SHFLAGS_VERSION}/shflags &&
               mv shflags-${SHFLAGS_VERSION}/shflags .shflags); then
```
skitt, line 14:
```suggestion
        if ! curl -L "https://raw.githubusercontent.com/kward/shflags/${SHFLAGS_VERSION}/src/shflags" > .shflags; then
```

mkolesnik: Only problem is for later versions this path is different.. :/

PR #196, opened 2020-06-01
Fix e2e kubeconfigs no context

mangelajo:
The e2e framework can be used by passing a list of kubeconfigs,
or a kubeconfig and multiple contexts.

The differences are handled in the BeforeSuite function
of framework.go and all the other functions can just consume
the clusters via KubeClients and RestConfigs.

This enables the lighthouse e2e tests from being consumed in 
subctl

### DIFF
--- a/test/e2e/framework/exec.go
+++ b/test/e2e/framework/exec.go
@@ -2,13 +2,11 @@ package framework
 
 import (
 	"bytes"
-	"fmt"
 	"io"
 	"net/url"
 	"strings"
 	"time"
 
-	. "github.com/onsi/gomega"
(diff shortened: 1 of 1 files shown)


test/e2e/framework/exec.go
```diff
@@ -68,10 +63,10 @@ func (f *Framework) ExecWithOptions(options ExecOptions, index ClusterIndex) (st
 	}
 
 	if options.PreserveWhitespace {
-		return stdout.String(), stderr.String(), err
+		return stdout.String(), stderr.String(), nil
```
vthapar, line 66: why discard error? what if we fail even after 5 attempts, we need to return error.

PR #205, opened 2020-06-09
Remove the --no-dataplane deprecated flag call

mangelajo:
During subctl deploy-broker we pass down the --no-dataplane
flag which was deprecated and it's being removed now. This
commit removes the flag usage.


mangelajo:
Once this merges we can retry this: https://github.com/submariner-io/submariner-operator/pull/448/checks#step:5:1727

PR #206, opened 2020-06-11
Update to subctl 0.4.0 rc3

mangelajo:


skitt:
This fails because the artifacts have `dev` in their name, not `0.4.0-rc2`: https://github.com/submariner-io/submariner-operator/releases/tag/v0.4.0-rc2

aswinsuryan:
@mangelajo  Can we change this to 0.4.0 and merge along with the release? The artifacts name should also be as expected then

mkolesnik:
> This fails because the artifacts have `dev` in their name, not `0.4.0-rc2`: https://github.com/submariner-io/submariner-operator/releases/tag/v0.4.0-rc2

This seems very similar to the problem seen on LH for this tag, which should be solved by #217 
Basically, the version is calculated by to be the tag, if one exists, but only if the working tree is clean. Seems that something was dirty when the binaries were built (go.sum perhaps?) and thus the `dev` version was calculated and that causes this screw up.

Perhaps we need to change the way version is calculated to always take the tag if we're on a tag commit, regardless of the state of the working tree state?

PR #207, opened 2020-06-11
Add using-flag support for OLM deploys

dfarrell07:
Support shortcut `using=olm` for `CLUSTERS_ARGS="--olm true"`.

Relates-to: submariner-io/submariner-operator#373

### DIFF
--- a/Makefile.inc
+++ b/Makefile.inc
@@ -16,6 +16,10 @@ ifneq (,$(filter helm,$(_using)))
 override DEPLOY_ARGS += --deploytool helm
 endif
 
+ifneq (,$(filter olm,$(_using)))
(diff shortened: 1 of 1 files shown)


Makefile.inc
```diff
@@ -16,6 +16,10 @@ ifneq (,$(filter helm,$(_using)))
 override DEPLOY_ARGS += --deploytool helm
 endif
 
+ifneq (,$(filter olm,$(_using)))
+override CLUSTERS_ARGS += --olm true
```
mkolesnik, line 20:
```suggestion
override CLUSTERS_ARGS += --olm
```

Since it's a boolean flag you send it like that

PR #212, opened 2020-06-17
Add a shared image building make target

### LINKED ISSUE #209: Add shared build target for images
Right now images are rebuilt each time, but we can only rebuild them if we need to.

It would be best to put this logic in Shipyard and use everywhere else.

mkolesnik:
This target knows to depend on the files being copied into the image,
and on the image itself, to make sure it rebuilds the image when
necessary (and doesn't when not).

Resolves #209


Makefile
```diff
@@ -20,29 +20,26 @@ $(TARGETS):
...
+dapper-image: package/.image.shipyard-dapper-base
 
 .DEFAULT_GOAL := validate
 # Shipyard-specific ends
 
 include Makefile.dapper
 
 endif
-
-# Disable rebuilding Makefile
```
skitt, line 47: Why remove this?

Makefile.dapper
```diff
@@ -10,7 +10,7 @@
 	@./.dapper.tmp -v
 	@mv .dapper.tmp .dapper
 
-%: .dapper
+$(filter-out .dapper dapper-image shell,$(MAKECMDGOALS)): .dapper
```
skitt, line 13: Ah, I guess that explains the `Makefile Makefile.dapper...` removal.

Makefile
mkolesnik: > The `MAKECMDFLAGS` change could do with an explanation in the commit message (or even a separate commit!).

Sure, I'll add a comment.

It had to be in this commit or it won't work (probably before it would work but I don't see a good reason to split it out now)

PR #218, opened 2020-06-22
Add GHA check for go.mod & go.sum being up to date

mkolesnik:
This check makes sure go.mod and go.sum are up to date for the code.

### DIFF
--- a/.github/workflows/tidy.yaml
+++ b/.github/workflows/tidy.yaml
@@ -1,0 +1,22 @@
+on:
+  pull_request:
+  push:
+    branches:
+      - master
+    tags:
+      - 'v**'
+
+name: Tidyness Validation
+jobs:
+  tidy-check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+
+      - run: make tidy
+      - name: Check that go.mod and go.sum files are up to date
+        run: |
+          [ -n "$(git status --porcelain --untracked-files=no)" ] || exit 0
+          git diff
+          echo "##[error] The go.mod and/or go.sum files are out of date, please run 'make tidy' and update your changes"
+          exit 1
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,7 @@ else
 include Makefile.images
 
 # Shipyard-specific starts
-clusters deploy e2e nettest post-mortem release unit-test validate: dapper-image
+clusters deploy e2e nettest post-mortem release tidy unit-test validate: dapper-image
 
 dapper-image: export SCRIPTS_DIR=./scripts/shared
 
--- a/Makefile.inc
+++ b/Makefile.inc
@@ -41,6 +41,9 @@ release:
 post-mortem:
 	$(SCRIPTS_DIR)/post_mortem.sh
 
+tidy:
+	go mod tidy && go mod vendor
+
 unit-test: vendor/modules.txt
 	$(SCRIPTS_DIR)/unit_test.sh $(UNIT_TEST_ARGS)
 


.github/workflows/go-mod.yaml
```diff
@@ -0,0 +1,22 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+
+      - run: go mod tidy
+      - run: go mod vendor
+      - run: |
+          if [ -n "$(git status --porcelain --untracked-files=no)" ]; then
+              echo "go.mod and/or go.sum are out of date, please run `go mod tidy && go mod vendor` and update your changes"
+              exit 1
```
skitt, line 21:
```suggestion
              git diff
              exit 1
```

would be useful to see the changes!

mkolesnik: thanks, ill add it!

mkolesnik:
> Actually let me retract my approval, sorry; I think this check should only be done if `go.mod` itself is being changed as part of the commit, otherwise we’ll end up with `go.mod` changes being included in unrelated commits.

Problem with this approach is that if the files are stale in the repository, it will cause VERSION to be calculated as `dev` since one or both of these files will change during the build.. Can these files become stale (especially go.sum) even if you don't touch go.mod? I think they can..

skitt:
> > Actually let me retract my approval, sorry; I think this check should only be done if `go.mod` itself is being changed as part of the commit, otherwise we’ll end up with `go.mod` changes being included in unrelated commits.
> 
> Problem with this approach is that if the files are stale in the repository, it will cause VERSION to be calculated as `dev` since one or both of these files will change during the build.. Can these files become stale (especially go.sum) even if you don't touch go.mod? I think they can..

I’ve seen `go.sum` go stale, yes, even though it shouldn’t if `go.mod` is correctly specified (in theory).

What I *really* don’t want is to see people adding `go.sum` updates to unrelated PRs, because that makes commits harder to cherry-pick etc.

Perhaps we could do the following: check `go.mod` / `go.sum` for changes if `go.mod` is updated in the current PR, *and* for release jobs...

mkolesnik:
> I’ve seen `go.sum` go stale, yes, even though it shouldn’t if `go.mod` is correctly specified (in theory).
> 
> What I _really_ don’t want is to see people adding `go.sum` updates to unrelated PRs, because that makes commits harder to cherry-pick etc.

Another idea is a job that runs periodically/on merge `go mod tidy && go mod vendor` such as this - https://github.com/crazy-max/diun/blob/master/.github/workflows/gosum.yml

This create a commit that syncs the go.sum if one is necessay.

Perhaps it can propose a PR instead of a commit if we want to be more careful.

> 
> Perhaps we could do the following: check `go.mod` / `go.sum` for changes if `go.mod` is updated in the current PR, _and_ for release jobs...

mangelajo:
> Actually let me retract my approval, sorry; I think this check should only be done if `go.mod` itself is being changed as part of the commit, otherwise we’ll end up with `go.mod` changes being included in unrelated commits.

I think that would be fine as would only happen for a short timeframe until this patch is in action for all merged commits.

.github/workflows/go-mod.yaml
```diff
@@ -0,0 +1,24 @@
...
+    steps:
+      - uses: actions/checkout@master
+
+      - run: go mod tidy
+      - run: go mod vendor
+      - name: Check that go.mod and go.sum files are up to date
+        run: |
+          if [ -n "$(git status --porcelain --untracked-files=no)" ]; then
+              git diff
+              echo "##[error] The go.mod and/or go.sum files are out of date, please run 'go mod tidy && go mod vendor' and update your changes"
```
mangelajo, line 22:
```suggestion
              echo "##[error] The go.mod and/or go.sum files are out of date, please run 'go mod tidy && go mod vendor' inside 'make shell' and update your changes"
```

PR #221, opened 2020-06-25
Need to fetch the tags in order for the release to work

mkolesnik:
Since we're using `git describe`, we need the CI to fetch everything
including the tags (it only fetches the specific commit by default).

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -26,6 +26,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@master
+        with:
+          fetch-depth: 0
 
       - name: Release the images
         env:


.github/workflows/release.yaml
```diff
@@ -26,6 +26,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@master
+        with:
+          fetch-depth: 0
```
mangelajo, line 30:
```suggestion
      - uses: actions/checkout@master
        with:
          fetch-depth: 0

      - run: git fetch origin +refs/tags/*:refs/tags/*
```

mkolesnik: No need, it's pulling the tags on the checkout step

mangelajo: Ok, I had issues with that on subctl, may be they changed the checkout action?

PR #224, opened 2020-06-25
Use more precise arguments in unit test

mkolesnik:
Intead of blindly sending the top level firectory and letting `go test`
find out the directories (and possibly include ignored ones), act only
on the directories we find to contain unit tests.

### DIFF
--- a/scripts/shared/lib/find_functions
+++ b/scripts/shared/lib/find_functions
@@ -1,11 +1,21 @@
 # shellcheck shell=bash
-function find_go_pkg_dirs() {
-    local find_exclude excluded_dirs="${*} vendor .git .trash-cache bin"
+function _find_pkg_dirs() {
+    local find_exclude
+    excluded_dirs+=" vendor .git .trash-cache bin"
 
     for dir in $excluded_dirs; do
         find_exclude+=" -path ./$dir -prune -o"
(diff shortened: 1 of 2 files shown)


scripts/shared/unit_test.sh
```diff
@@ -9,7 +9,7 @@ echo "Looking for packages to test"
 
 # This can’t be done as simply with parameter substitution
 # shellcheck disable=SC2001
```
skitt, line 11:
```suggestion
```

scripts/shared/lib/find_functions
```diff
@@ -1,11 +1,21 @@
...
+    local find_exclude
+    excluded_dirs+=" vendor .git .trash-cache bin"
 
     for dir in $excluded_dirs; do
         find_exclude+=" -path ./$dir -prune -o"
     done
 
     # shellcheck disable=SC2086
-    find . ${find_exclude} -path './*/*.go' -print | cut -f2 -d/ | sort -u
+    find . ${find_exclude} -path "$1" -exec dirname {} \; | sort -u
```
skitt, line 11:
```suggestion
    find . ${find_exclude} -path "$1" -printf "%h\n" | sort -u
```

PR #228, opened 2020-06-29
Split out a system image

skitt:
Working on scripts in the Dapper image is quite painful because of the
constant rebuilds of the Fedora base, so this patch splits out our
“system” image (not named after Fedora so it’s not tied to that).

### DIFF
--- a/package/Dockerfile.shipyard-system-base
+++ b/package/Dockerfile.shipyard-system-base
@@ -1,0 +1,34 @@
+FROM fedora:32
+
+ENV DAPPER_HOST_ARCH=amd64 SHIPYARD_DIR=/opt/shipyard
+ENV HOST_ARCH=${DAPPER_HOST_ARCH} ARCH=${DAPPER_HOST_ARCH}
+
+# Requirements:
+# Component      | Usage
+# -----------------------------------------------------------
+# gcc            | needed by `go test -race` (https://github.com/golang/go/issues/27089)
+# git            | find the workspace root
+# curl           | download other tools
+# moby-engine    | Dapper (Docker)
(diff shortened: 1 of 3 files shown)


package/Dockerfile.shipyard-system-base
```diff
@@ -0,0 +1,34 @@
...
+# ginkgo         | tests
+# goimports      | code formatting
+# make           | OLM installation
+# findutils      | validate (find go packages)
+# subctl *       | Submariner's deploy tool (operator)
+# upx            | binary compression
+# jq             | JSON processing (GitHub API)
+# diffutils      | required for goimports
+# ShellCheck     | shell script linting
+RUN dnf -y install --nodocs --setopt=install_weak_deps=False \
```
mangelajo, line 27: Why don't we keep this part as a more static first layer of the image, and then... a 2nd run with all the less-static parts?

I prefer the single image approach if that works, limiting complexity to some extent.

mkolesnik: I second that approach.

skitt: > Why don't we keep this part as a more static first layer of the image, and then... a 2nd run with all the less-static parts?

Because I tried that and it doesn’t work, with this image at least. The intermediary layers are discarded and never re-used when building updated images.

mkolesnik: I've investigated it and proposed a solution at #244 :)

skitt: You’re a star @mkolesnik, I’ll check that out and see how it behaves.

PR #237, opened 2020-07-01
Add E2E job to test consuming projects

### LINKED ISSUE #211: Add capability to run consuming projects e2e
We need a way to run CI jobs that test that we don't break consuming projects.

One idea is to trigger such jobs in GHA by comments, eg "/testprojects" or something like that

mkolesnik:
This job will run if the PR is labeled with `test-projects` and will
make sure to check out the consuming projects and run them with the
latest Shipyard image that includes the PR contents.

Resolves #211 

### DIFF
--- a/.github/workflows/e2e-consuming.yaml
+++ b/.github/workflows/e2e-consuming.yaml
@@ -1,12 +1,15 @@
 on:
   pull_request:
+    types: [labeled, opened, synchronize, reopened]
 
 name: E2E Tests (Consuming Projects)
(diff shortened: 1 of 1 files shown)


mkolesnik:
When it skips the check (no `test-projects` label):

image

mkolesnik:
Force pushed to have it run with the label applied (seems that it uses the original event data, so if you set a label afterwards it doesn't recognize it, I'll see if this can be optimized)

PR #238, opened 2020-07-01
Add admiral to the e2e-consuming projects

skitt:

### DIFF
--- a/.github/workflows/e2e-consuming.yaml
+++ b/.github/workflows/e2e-consuming.yaml
@@ -13,9 +13,11 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        project: ['submariner', 'submariner-operator', 'lighthouse']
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e-consuming.yaml
```diff
@@ -13,7 +13,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        project: ['submariner', 'submariner-operator', 'lighthouse']
+        project: ['admiral', 'submariner', 'submariner-operator', 'lighthouse']
         deploytool: ['operator', 'helm']
         exclude:
```
mkolesnik, line 18:
```suggestion
        exclude:
          - project: admiral
            deploytool: helm
```

We should exclude helm as it's redundant in the case of admiral and admiral repo doesnt test it anyhow

skitt: Done!

PR #249, opened 2020-07-06
Align DEV_VERSION with expected tags

skitt:
The “latest” image is now “devel”, but we locally build “latest” and
“dev”, which complicates consuming local images. This changes
DEV_VERSION to “devel” so that we build “devel” images instead.


skitt:
Dropped in favour of https://github.com/submariner-io/submariner-website/pull/189

PR #254, opened 2020-07-07
Turn the version script into a Makefile snippet

skitt:
... and include it as appropriate from Makefile.inc and Makefile (the latter is necessary for the initial Dapper image build).

This ensures CUTTING_EDGE, DEV_VERSION and VERSION are exported everywhere, with the same value, calculated only once.


mkolesnik:
One thing I don't like though, it that now we always have to remember to checkout the GH repos with depth 0 :/

skitt:
> One thing I don't like though, it that now we always have to remember to checkout the GH repos with depth 0 :/

I’m not 100% sure we need this all the time, it just results in an error message without breaking anything.

PR #255, opened 2020-07-08
Saner UPX

mkolesnik:
Turns out we inadvertently ruined the image caching since we're using upx -1 for CI but -9 for release, which causes docker cache misses.

This PR fixes that by only UPXing to a higher level on a tagged release, and also changes the default UPX level to -5 which seems to be a sweet spot between build time and space saved.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -14,6 +14,8 @@ jobs:
       - uses: actions/checkout@master
 
       - run: make e2e
+        env:
+          IMAGES_ARGS: --nocache
 
       - name: Post Mortem
         if: failure()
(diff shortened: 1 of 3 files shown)


mkolesnik:
> LGTM, but this is starting to get quite complicated. It’s fine to merge this from my point of view, but I wonder if `-5` wouldn’t be an acceptable compromise in all cases...

We can see from that one go at different levels that the difference between the two was 15MB (obviously depends on the image contents, but I think it scales overall):
UPX level = 5
Image size = 846MB
UPX level = 9
Image size = 831MB

I don't mind considering the 15MB negligible and going with -5 as the default level all the time since it's close to the point of diminishing returns.
Mind you, at the default -8 level the difference was even smaller (12MB) while time difference is huge (-8 is >3x slower).

If everyone's ok with that I don't mind keeping just -5 and leave the option to change it if we chose so in the future.

PR #273, opened 2020-08-10
Fix MD lint violations

dfarrell07:
Configure custom maximum line length of 140 chars to match Go linting.

Per discussions on Automation call and PR, preemptively wrap at the end
of sentences that are shorter than 140c but can not fully fit the next
sentence. This will make diffs better, as changes are typically
sentence-based, compared to re-wrapping a block of 140c lines. We don't
know of a good way to enforce this in automation, but we would like to.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,7 @@
 # Shipyard
 
-The Shipyard project provides tooling for creating K8s clusters with [kind](K8s in Docker) and provides a Go framework for creating E2E tests. 
+The Shipyard project provides tooling for creating K8s clusters with [kind](K8s in Docker) and provides a Go framework for creating E2E
+tests.
 
 [![Build Status](https://travis-ci.com/submariner-io/shipyard.svg?branch=master)](https://travis-ci.com/submariner-io/shipyard)
 [![Go Report Card](https://goreportcard.com/badge/github.com/submariner-io/shipyard)](https://goreportcard.com/report/github.com/submariner-io/shipyard)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -37,7 +39,8 @@ You can also refer to the project's own [Dockerfile.dapper](Dockerfile.dapper) a
 
 ### Makefile
 
-The Makefile should include targets to run everything in Dapper. They're defined in [Makefile.dapper](Makefile.dapper) and can be copied as-is and included. To use Shipyard's target, simply include the [Makefile.inc](Makefile.inc) file in your own Makefile.
+The Makefile should include targets to run everything in Dapper. They're defined in [Makefile.dapper](Makefile.dapper) and can be copied
+as-is and included. To use Shipyard's target, simply include the [Makefile.inc](Makefile.inc) file in your own Makefile.
```
skitt, line 43:
```suggestion
The Makefile should include targets to run everything in Dapper.
They're defined in [Makefile.dapper](Makefile.dapper) and can be copied as-is and included.
To use Shipyard's target, simply include the [Makefile.inc](Makefile.inc) file in your own Makefile.
```

It’s better to break long lines following sentence structure; this reduces the impact of future changes (which tend to be sentence-bound) in diffs.

dfarrell07: Interesting idea. I agree it might make diffs better, but I for one plan on using automation to auto-wrap lines, and I don't see a good option for automatically following this pattern. For this tiny README it's not a problem, but I also have in mind the massive LL PR still pending for the website.

PR #281, opened 2020-08-18
Add commit message validations

dfarrell07:
Run gitlint to validate the format of commit messages. Add corresponding
make target. Tests commits between master and HEAD.

### DIFF
--- a/.github/workflows/gitlint.yaml
+++ b/.github/workflows/gitlint.yaml
@@ -1,0 +1,18 @@
+on:
+  pull_request:
+  push:
+    branches:
+      - master
+    tags:
+      - 'v**'
+
+name: Commit Message Validations
+jobs:
+  gitlint:
+    name: Gitlint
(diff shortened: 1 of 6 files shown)


dfarrell07:
You can see it works for PRs with multiple commits where not-the-first commit has errors here: https://github.com/submariner-io/shipyard/pull/282/checks?check_run_id=997725545

dfarrell07:
Maybe we shouldn't do this as part of the validate script, because it's not currently run against non-Go projets (like the website, releases repo, enhancements repo).

mangelajo (approved):

dfarrell07:
> Maybe we shouldn't do this as part of the validate script, because it's not currently run against non-Go projets (like the website, releases repo, enhancements repo).

I re-did it as a separate make target and job. I think that makes a lot more sense as I work through the details of how it will integrate into other repos.

skitt (approved):

Issue #276, opened 2020-08-13
make deploy using=prometheus unusable outside Shipyard


skitt:
The deployment assumes the YAML files are accessible, they aren’t in projects consuming Shipyard.

skitt:
This blocks https://github.com/submariner-io/submariner-operator/issues/625.

PR #284, opened 2020-08-20
Move the Prometheus descriptors so they're shared

skitt:
This will allow downstream projects to deploy with Prometheus.

Fixes: #276


skitt:
Verified by https://github.com/submariner-io/shipyard/pull/275.

PR #289, opened 2020-08-26
Remove most e2e framework dependencies on ginkgo

mangelajo:
This will enable the usage of the Framework object
outside of ginkgo. For example from subctl, when we
need a multi-cluster client for benchmarking, etc..

There is still work that needs to be done in the
logging module, although it's safe now, because
GinkgoWriter defaults to stdout unless you change
the reporters.

### DIFF
--- a/test/e2e/framework/framework.go
+++ b/test/e2e/framework/framework.go
@@ -4,10 +4,8 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
-	"strings"
 	"time"
 
-	"github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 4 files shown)


test/e2e/framework/framework.go
```diff
@@ -86,25 +84,48 @@ var (
...
+		return "shipyard-framework-agent"
+	}
+}
+
 func BeforeSuite() {
-	ginkgo.By("Creating kubernetes clients")
+	By("Creating kubernetes clients")
 
 	if len(TestContext.KubeConfig) > 0 {
 		Expect(len(TestContext.KubeConfigs)).To(BeZero(),
```
maayanf24, line 131: Do we need a wrap to `Except`?

skitt (approved):

maayanf24 (approved):

test/e2e/framework/framework.go
mangelajo: That's a gomega dependency, it's ok, that one is note easily switchable, but framework users can just do:
```
gomega.RegisterFailHandler(func(message string, callerSkip ... int) { panic(message) })
```
For now

PR #296, opened 2020-09-03
Add netperf to nettest image

maayanf24:
To be used in benchmark testing

Fix issue #299

Signed-off-by: Maayan Friedman <maafried@redhat.com>

### DIFF
--- a/package/Dockerfile.nettest
+++ b/package/Dockerfile.nettest
@@ -10,4 +10,12 @@ RUN apk add --no-cache \
 	iperf3 \
 	tcpdump
 
+RUN apk add --update --no-cache g++ make curl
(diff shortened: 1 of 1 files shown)


sridhargaddam (approved):
LGTM, thanks @maayanf24

skitt:
It would be better to clean up after the build, but that doesn’t actually appear to reduce the final image size, so the real fix would be to use a multi-stage build, but that breaks our image-building scripts.

I’ll open a separate PR with fixes for the multi-stage build and a more compact `nettest` image.

skitt (approved):

tpantelis (approved):

PR #301, opened 2020-09-07
Fix multi-stage builds, simplify the nettest image

skitt:
When calculating the dependencies for images, the --from option to
COPY breaks our call to find. Such dependencies need to be ignored
altogether, so we skip COPY when it is followed by --from=.

To avoid bloating the nettest image with the remnants of the netperf
build (even after clean up), use a multi-stage build: the first builds
netperf, the second copies the binaries to the target image. This
results in a 24MiB image.

This builds on top of #296.

### DIFF
--- a/Makefile.images
+++ b/Makefile.images
@@ -12,8 +12,10 @@ force_image_rebuild = $(if $(shell docker images | grep '\<$(1)\W*latest\>'),,FO
 FORCE_IMAGE: ;
 
 # Dockerfile dependencies are the file and any file copied into it
-# We have to run it through a variable in order to expand any * that might be in the COPY command
-docker_deps = $(shell files="$(1) $$(grep COPY $(1) | sed 's/COPY \(.*\) .*/\1/' )" && find $${files[*]})
+# We have to run it through a variable in order to expand any * that might be
+# in the COPY command; find is used to handle directories as dependencies
+# Files copied from another image are ignored
+docker_deps = $(shell files=($(1) $$(awk '/COPY/ { if (substr($$2, 1, 7) != "--from=") { for (i = 2; i < NF; i++) { print $$i } } }' $(1))) && find $${files[*]} -type f)
 
 # Patterned recipe to use to build any image from any Dockerfile
 # An empty file is used for make to figure out if dependencies changed or not
@@ -21,5 +23,3 @@ docker_deps = $(shell files="$(1) $$(grep COPY $(1) | sed 's/COPY \(.*\) .*/\1/'
 package/.image.%: $$(call docker_deps,package/Dockerfile.$$*) $$(call force_image_rebuild,$$*)
 	$(SCRIPTS_DIR)/build_image.sh -i $(lastword $(subst ., ,$@)) -f $< $(IMAGES_ARGS)
 	touch $@
-
-
--- a/package/Dockerfile.nettest
+++ b/package/Dockerfile.nettest
@@ -2,6 +2,17 @@ FROM alpine
 
 WORKDIR /app
 
+RUN apk add --update --no-cache gcc libc-dev make curl
+
+RUN curl -L https://github.com/HewlettPackard/netperf/archive/netperf-2.7.0.tar.gz | tar xzf -
+RUN cd netperf-netperf-2.7.0 && ./configure \
+    && make && make install
+
+
+FROM alpine
+
+WORKDIR /app
+
 RUN apk add --no-cache \
 	bash \
 	bind-tools \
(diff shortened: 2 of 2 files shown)


Makefile.images
```diff
@@ -12,8 +12,10 @@ force_image_rebuild = $(if $(shell docker images | grep '\<$(1)\W*latest\>'),,FO
 FORCE_IMAGE: ;
 
 # Dockerfile dependencies are the file and any file copied into it
-# We have to run it through a variable in order to expand any * that might be in the COPY command
-docker_deps = $(shell files="$(1) $$(grep COPY $(1) | sed 's/COPY \(.*\) .*/\1/' )" && find $${files[*]})
+# We have to run it through a variable in order to expand any * that might be
+# in the COPY command; find is used to handle directories as dependencies
+# Files copied from another image are ignored
+docker_deps = $(shell files="$(1) $$(awk '/COPY/ { if (substr($$2, 1, 7) != "--from=") { print $$2 } }' $(1))" && find $${files[*]})
```
maayanf24, line 18: @skitt Thank you for working on this.
Can you explain how this code works?

skitt: The goal of this is to list all the local dependencies of a container image.

This defines a Make function, which is designed to operate on a `Dockerfile`. `$(1)` gets replaced with the file being processed, for example `package/Dockerfile.shipyard-dapper-base`:

    $(shell files="package/Dockerfile.shipyard-dapper-base $$(awk '/COPY/ { if (substr($$2, 1, 7) != "--from=") { print $$2 } }' package/Dockerfile.shipyard-dapper-base)" && find $${files[*]})

`$(shell ...)` causes Make to run the given command using the shell, and replace the whole `$(shell ...)` expression with its output. `$$` becomes`$`, so the shell command is

    files="package/Dockerfile.shipyard-dapper-base $(awk '/COPY/ { if (substr($2, 1, 7) != "--from=") { print $2 } }' package/Dockerfile.shipyard-dapper-base)" && find ${files[*]}

Working from the inside out, the first part to run is the AWK invocation, which matches lines containing “COPY”, and then checks whether the second word in the line starts with `--from=`; if it doesn’t, it outputs the second word (the source being copied — incidentally, we’ll miss other arguments if more than one set of files is copied, I’ll fix that). The result of all that is all the files and directories being copied into the image. This is then assigned to the `files` variable, along with the name of the file we’re looking at; this assignment causes any wildcards to be expanded, so for example `Makefile.*` becomes the list of matching files.

Once we have the list of files in the `files` variable, we pass that to `find`, which lists all the files and directories (which we don’t really need, I’ll fix that too). The result of `find` is the final output of the shell command, and that’s what ends up being used as the result of the `docker_deps` call.

maayanf24 (approved):

Makefile.images
```diff
@@ -12,8 +12,10 @@ force_image_rebuild = $(if $(shell docker images | grep '\<$(1)\W*latest\>'),,FO
 FORCE_IMAGE: ;
 
 # Dockerfile dependencies are the file and any file copied into it
-# We have to run it through a variable in order to expand any * that might be in the COPY command
-docker_deps = $(shell files="$(1) $$(grep COPY $(1) | sed 's/COPY \(.*\) .*/\1/' )" && find $${files[*]})
+# We have to run it through a variable in order to expand any * that might be
+# in the COPY command; find is used to handle directories as dependencies
+# Files copied from another image are ignored
+docker_deps = $(shell files=($(1) $$(awk '/COPY/ { if (substr($$2, 1, 7) != "--from=") { for (i = 2; i < NF; i++) { print $$i } } }' $(1))) && find $${files[*]} -type f)
```
mkolesnik, line 18: As if it wasn't complicated enough? ;)

package/Dockerfile.nettest
```diff
@@ -2,6 +2,17 @@ FROM alpine
 
 WORKDIR /app
 
+RUN apk add --update --no-cache gcc libc-dev make curl
+
+RUN curl -L https://github.com/HewlettPackard/netperf/archive/netperf-2.7.0.tar.gz | tar xzf -
+RUN cd netperf-netperf-2.7.0 && ./configure \
+    && make && make install
+
```
mkolesnik, line 10:
```suggestion
```

OCD

mkolesnik (approved):
Thanks for fixing the `--from` thing :)

Makefile.images
skitt: Yeah, and it doesn’t actually work in all cases (but not because of this PR) — `find` skips files which don’t exist, so container dependencies which don’t exist aren’t listed! I’ll fix that in another PR.

package/Dockerfile.nettest
skitt: That was done on purpose, to separate the two images.

dfarrell07 (approved):

PR #308, opened 2020-09-11
Improve post-mortem output          This commit dumps the logs of submariner-operator and routeagent even     if those pods are healthy, which sometimes will help understanding     specific issues.          We start also making use of `subctl show all` to report the status     of connections, endpoints, etc.          Signed-off-by: Miguel Angel Ajo <majopela@redhat.com>

mangelajo:


skitt (approved):

scripts/shared/post_mortem.sh
```diff
@@ -26,6 +26,14 @@ function post_analyze() {
 
     echo "++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++"
 
+    namespace="submariner-operator"
+    for pod in $(kubectl get pods --selector=name=submariner-operator -n $namespace -o jsonpath='{.items[*].metadata.name}'); do
```
sridhargaddam, line 30: The PR in itself is good. But I have a feeling that we will be dumping too many logs during post-mortem. Let's keep an eye on the logs and if we feel something is logged frequently, we can make the necessary changes in the code.

sridhargaddam (approved):

scripts/shared/post_mortem.sh
mangelajo: Yeah, we could remove logs later. But I find so many times that I want something from post mortem and it isn't there ':-)

PR #310, opened 2020-09-11
Fix intermittent failures in test_connection

sridhargaddam:
Fixes issue: https://github.com/submariner-io/shipyard/issues/309

### DIFF
--- a/scripts/shared/lib/deploy_funcs
+++ b/scripts/shared/lib/deploy_funcs
@@ -85,6 +85,11 @@ function connectivity_tests() {
 
 function verify_gw_status() {
     sleep_duration=5
+    # helm doesn't use the operator yet, and connection status is based on the operator object
+    if subctl show connections 2>&1 | grep "the server could not find the requested resource"; then
+        return 0
+    fi
+
     for i in {1..90}; do
        if ! subctl show connections | grep "connected"; then
           echo "iter: $i. Clusters not yet connected. sleeping for $sleep_duration secs"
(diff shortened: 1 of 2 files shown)


mangelajo (approved):

mangelajo (commented):
I would add also a check with 

subctl show connections 

to make sure we at least have one connection (it'd be better to check that we have the same amount as "joined" clusters if we can get that number somehow...) but not sure if it's easy to get.

sridhargaddam:
> I would add also a check with
> 
> subctl show connections
> 
> to make sure we at least have one connection (it'd be better to check that we have the same amount as "joined" clusters if we can get that number somehow...) but not sure if it's easy to get.

Okay @mangelajo. I'll update the PR. Thanks for reviewing.

scripts/shared/lib/deploy_funcs
```diff
@@ -75,6 +83,19 @@ function connectivity_tests() {
     with_context "$target_cluster" remove_resource "${RESOURCES_DIR}/nginx-demo.yaml"
 }
 
+function verify_gw_status() {
+    sleep_duration=5
```
mangelajo, line 87:
```suggestion
    sleep_duration=5
    # helm doesn't use the operator yet, and connection status is based on the operator object
    if subctl show connections 2>&1 | grep "the server could not find the requested resource"; then
        return 0
    fi

```

skitt (approved):

dfarrell07:
> Expected "Miguel Angel Ajo Pelayo <miguelangel@ajo.es>", but got "Miguel Angel Ajo Pelayo <majopela@redhat.com>".

Is your sign-off email differently configured than your commit author email @mangelajo?

dfarrell07 (approved):

tpantelis (approved):

tpantelis:
DCO check failed. Also can you update the PR summary to add details as to the problem/solution?

mangelajo:
I have handled the DCO & Summary as part of the squash

PR #312, opened 2020-09-14
Add clang and enable -msan

skitt:
We’ll need clang sooner or later for eBPF, and having it available
allows us to run unit tests with memory sanitisation (but not at the
same time as race detection).


dfarrell07 (approved):

tpantelis (approved):

skitt:
Note that this adds over *200 MiB* to the Docker image (uncompressed).

dfarrell07:
Do we know what's going on with the connectivity test failures?

mangelajo (approved):
not sure if we will end up using ebpf in the end... still investigating

PR #316, opened 2020-09-18
Increase golangci-lint timeout

skitt:
I'm increasingly running into golangci-lint timeouts; increasing it to
10 minute would avoid these.

### DIFF
--- a/Makefile.inc
+++ b/Makefile.inc
@@ -44,7 +44,7 @@ gitlint:
 
 golangci-lint: vendor/modules.txt
 	golangci-lint linters
-	golangci-lint run --timeout 5m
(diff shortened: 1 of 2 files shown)


dfarrell07 (dismissed):

dfarrell07:
It would be better to also increase it here: https://github.com/submariner-io/shipyard/blob/master/Makefile.inc#L47

skitt:
> It would be better to also increase it here: https://github.com/submariner-io/shipyard/blob/master/Makefile.inc#L47

Indeed, done! (I’ll leave the clean-up for another PR.)

dfarrell07 (approved):

tpantelis (approved):

PR #321, opened 2020-09-25
Use the latest development subctl

skitt:
... until we're closer to releasing 0.7, to avoid problems caused by
mis-matches between devel images and older subctl.


mangelajo (commented):
This looks good, I wonder how we can retrigger a build later without dummy changes.
Probably just by tagging shipyards: v0.7.0-pre1, pre2, pre3 may be that only releases the tag, I'm not sure

I think we shall move the subctl downloads to happen in runtime, what do you think?

dfarrell07 (approved):

mangelajo:
Ok, let's merge this for now so we unbreak existing jobs, then let's make the download dynamic.

mangelajo (approved):

PR #329, opened 2020-10-09
Scalability improvements

mangelajo:
In separate commits:

* Support setting up single-node clusters.
* Allow up to 255 clusters by widening the usable IPs

Co-Authored-by: Mike Kolesnik <mkolesni@redhat.com>

### DIFF
--- a/scripts/shared/lib/deploy_funcs
+++ b/scripts/shared/lib/deploy_funcs
@@ -103,12 +103,12 @@ function verify_gw_status() {
 
 function add_subm_gateway_label() {
     [[ ${cluster_subm[$cluster]} = "true" ]] || return 0
-    kubectl label node "${cluster}-worker" "submariner.io/gateway=true" --overwrite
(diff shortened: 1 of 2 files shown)


skitt (approved):

mangelajo:
It seems like sem can't work with functions defined in bash, makes sense...
```
[shipyard]$ [cluster1] sem -j 8 prepare_cluster submariner-operator
[shipyard]$ [cluster1] sed /\[cluster1]/!s/^/[cluster1] /
[cluster1] /usr/bin/bash: prepare_cluster: command not found
[cluster2] /usr/bin/bash: prepare_cluster: command not found
```

I'll remove that from this PR and consider a different solution

mangelajo:
Hmmm, the deployments seem to connect but something goes wrong. Conflicting routes may be?, May be we should give a thought to the IP ranges used

mangelajo:
Ok, I splitted this in #333 and #334

PR #342, opened 2020-10-19
Batch of GHA CI updates

dfarrell07:


dfarrell07:
Note that this will require both updates to the GH admin config and the labels.

.github/workflows/e2e-consuming.yml
```diff
@@ -1,16 +1,18 @@
...
-name: E2E Tests (Consuming Projects)
 jobs:
   e2e:
+    name: E2E
     timeout-minutes: 30
     runs-on: ubuntu-latest
     if: |
-      ( github.event.action == 'labeled' && github.event.label.name == 'test-projects' )
-      || ( github.event.action != 'labeled' && contains(github.event.pull_request.labels.*.name, 'test-projects') )
+      ( github.event.action == 'labeled' && github.event.label.name == 'e2e-projects' )
```
tpantelis: https://github.com/submariner-io/submariner-website/blob/master/src/content/contributing/release-process/_index.en.md references `test-projects` so that will need to be changed as well.

tpantelis (approved):

.github/workflows/e2e-consuming.yml
dfarrell07: Good catch @tpantelis, thanks! I sent a PR to update those docs: https://github.com/submariner-io/submariner-website/pull/321.

skitt (approved):

PR #350, opened 2020-11-02
Deploy Lighthouse images

skitt:
... now that it's enabled by default (otherwise the Lighthouse pods
never start in a Shipyard-based deployment).

### DIFF
--- a/scripts/shared/deploy.sh
+++ b/scripts/shared/deploy.sh
@@ -42,6 +42,8 @@ import_image quay.io/submariner/nettest
 import_image quay.io/submariner/submariner
 import_image quay.io/submariner/submariner-route-agent
 [[ $globalnet != "true" ]] || import_image quay.io/submariner/submariner-globalnet
+import_image quay.io/submariner/lighthouse-agent
+import_image quay.io/submariner/lighthouse-coredns
 
 load_deploytool $deploytool
 deploytool_prereqs


nyechiel (approved):

mkolesnik (changes requested):
Actually this doesnt make too much sense.

Projects using lighthouse preload the images, and those that aren't (such as Submariner) need not be concerned with it.

We already discussed it previously and agreed to keep lighthouse logic out of Shipyard.

skitt:
> Actually this doesnt make too much sense.
> 
> Projects using lighthouse preload the images, and those that aren't (such as Submariner) need not be concerned with it.
> 
> We already discussed it previously and agreed to keep lighthouse logic out of Shipyard.

Yup, I remember that, but the problem now is that `make deploy` in Shipyard produces a cluster with broken pods — the Lighthouse pods are deployed but can’t pull their images.

mkolesnik:
> Yup, I remember that, but the problem now is that `make deploy` in Shipyard produces a cluster with broken pods — the Lighthouse pods are deployed but can’t pull their images.

Right, and whats the problem with that specifically?

Also I'm not suite sure why it results in broken pods, if the lighthouse pods point to the original quay.io repo then they should deploy fine as long as the actual image is there.

If you want to deploy stuff with working lighthouse youre not expected to do it from Shipyard anyhow.

I think what we discussed previously with @mangelajo is the actual solution here, allowing specifying overrides of specific images for subctl so only those can be pushed locally and everything else is just taken from Quay (or whatever hub were using). 
https://github.com/submariner-io/submariner-operator/pull/726

skitt:
> > Yup, I remember that, but the problem now is that `make deploy` in Shipyard produces a cluster with broken pods — the Lighthouse pods are deployed but can’t pull their images.
> 
> Right, and whats the problem with that specifically?

It’s disconcerting for developers using `make deploy`.

> Also I'm not suite sure why it results in broken pods, if the lighthouse pods point to the original quay.io repo then they should deploy fine as long as the actual image is there.

That’s the problem though, the pods point to the local registry.

> If you want to deploy stuff with working lighthouse youre not expected to do it from Shipyard anyhow.

I don’t care about Lighthouse when deploying from Shipyard, but I do expect the default deployment to be entirely functional, with no errors from Kubernetes.

> I think what we discussed previously with @mangelajo is the actual solution here, allowing specifying overrides of specific images for subctl so only those can be pushed locally and everything else is just taken from Quay (or whatever hub were using).
> [submariner-io/submariner-operator#726](https://github.com/submariner-io/submariner-operator/pull/726)

OK, that is indeed better.

mangelajo:
May be we (I mean... I) should finish this instead: https://github.com/submariner-io/submariner-operator/pull/726

PR #358, opened 2020-11-09
Add CI to find periodic flaky E2E failures

dfarrell07:
Run E2E for consuming projects periodically to find flaky failures.

Because this runs against code that has already passed CI to be merged
to the main branch, any failures are likely to be sporadic flakes. By
having a single GHA workflow to find these failures, we can more easily
spot them and track data about them.

The frequency of this job should be turned up or down depending on how
actively we're hunting for flaky failures.

### DIFF
--- a/.github/workflows/flake_finder.yml
+++ b/.github/workflows/flake_finder.yml
@@ -1,29 +1,24 @@
 ---
-name: Consuming Projects
+name: Flake Finder
 
 on:
-  pull_request:
-    types: [labeled, opened, synchronize, reopened]
+  schedule:
+    - cron: "0 0 * * *"
 
 jobs:
   e2e:
-    name: E2E
+    name: E2E Consuming Projects
     timeout-minutes: 30
     runs-on: ubuntu-latest
-    if: |
-      ( github.event.action == 'labeled' && github.event.label.name == 'e2e-projects' )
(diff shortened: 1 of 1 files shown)


dfarrell07:
I guess one major issue with this unified-Shipyard-job approach is that we aren't covering the full matrix used by the consuming projects. For example, from the main repo:

```
        globalnet: ['', 'globalnet']
        cable_driver: ['', 'wireguard', 'libreswan']
```

dfarrell07:
I'm starting to think it's better to use per-repo Periodic E2E workflows vs this central one in Shipyard. See: https://github.com/submariner-io/lighthouse/pull/364#issuecomment-723978658.

dfarrell07:
It's weird that the same `using=helm` job fails when triggered in Lighthouse from Shipyard but passes in Lighthouse's CI.

https://github.com/submariner-io/shipyard/pull/358/checks?check_run_id=1373973762
https://github.com/submariner-io/lighthouse/runs/1359928980?check_suite_focus=true

dfarrell07:
It's also weird that the `submariner-operator` and `lighthouse` part of this matrix failed, but the same tests seem to run/pass in the CI of the operator repo.

dfarrell07:
Closing in favor of https://github.com/submariner-io/submariner-operator/pull/786.

PR #359, opened 2020-11-09
Check inotify settings

skitt:
The default inotify settings prevent successful deployments of
configurations with three or more clusters in our default settings.
This trips developers up regularly, so adding an explicit check would
save time in the future.


maayanf24 (approved):
@skitt  Thank you for adding this

dfarrell07 (approved):

tpantelis (approved):

Issue #364, opened 2020-11-18
Add using= flag support for cable-driver, fixing hard-coded "strongswan" and ignored overrides in operator CI


dfarrell07:
We currently implement `using=<cable-driver>` only in submariner-io/submariner, not in Shipyard. This has caused confusion in submariner-io/submariner-operator, where we try to use `using=<cable-driver>` although it isn't respected. We also have the cable-driver flag hard coded to strongswan, which is an issue as we try to switch the default to libreswan this release.

dfarrell07:
There seems to be some nuance to Helm/Operator cable-driver flags that we should keep in mind.

```
ifneq (,$(cable_driver))
ifneq (,$(filter helm,$(_using)))
override DEPLOY_ARGS += --deploytool_submariner_args '--set cable-driver=$(cable_driver)'
else
override DEPLOY_ARGS += --deploytool_submariner_args '--cable-driver $(cable_driver)'
endif
endif
```

dfarrell07:
We're still not acting on the cable-driver flag in Helm (https://github.com/submariner-io/submariner-charts/pull/100#issuecomment-7424299640), but imho it's still correct to close this as it was focused on fixing the Operator.

PR #365, opened 2020-11-23
Add cable-driver support to using flag, add tests

dfarrell07:
Adds support for setting the cable driver via using= flag.

make e2e using=libreswan

Relates-to: #364

### DIFF
--- a/.github/workflows/consuming.yml
+++ b/.github/workflows/consuming.yml
@@ -19,8 +19,10 @@ jobs:
         project: ['admiral', 'submariner', 'submariner-operator', 'lighthouse']
         deploytool: ['operator', 'helm']
         exclude:
+          # Admiral E2E doesn't respect deploy-tool params, as it uses clusters without Submariner
           - project: admiral
             deploytool: helm
+          # Operator and Helm are mutually exclusive, don't try to use Helm in Operator repo
           - project: submariner-operator
             deploytool: helm
     steps:
@@ -46,7 +48,7 @@ jobs:
         run: sed -i 's/shipyard-dapper-base:*.*/shipyard-dapper-base:dev/' Dockerfile.dapper
 
       - name: Run E2E deployment and tests
-        run: make e2e using=${{ matrix.deploytool }}
+        run: make e2e using="${{ matrix.deploytool }}"
 
       - name: Post mortem
         if: failure()
--- a/Makefile.inc
+++ b/Makefile.inc
@@ -6,6 +6,15 @@ include $(SHIPYARD_DIR)/Makefile.versions
 
 # Process extra flags from the `using=a,b,c` optional flag
 
+ifneq (,$(filter libreswan,$(_using)))
+override DEPLOY_ARGS += --cable_driver libreswan
+else ifneq (,$(filter strongswan,$(_using)))
+override DEPLOY_ARGS += --cable_driver strongswan
+else ifneq (,$(filter wireguard,$(_using)))
+# Wireguard requires kernel module install on the host
+override DEPLOY_ARGS += --cable_driver wireguard
+endif
+
 ifneq (,$(filter globalnet,$(_using)))
 override CLUSTERS_ARGS += --globalnet
 override DEPLOY_ARGS += --globalnet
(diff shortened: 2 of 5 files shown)


dfarrell07:
The failing lighthouse jobs are a known flaky failure. The flag is being set/consumed correctly.

```
2020-11-23T12:39:29.8545562Z [36m[lighthouse]$ [cluster2] grep connected[0m
2020-11-23T12:39:30.5921227Z cluster1-worker                 cluster1                172.17.0.9      strongswan          100.1.0.0/16, 10.1.0.0/16               connected       
2020-11-23T12:39:30.5922570Z cluster2-worker                 cluster2                172.17.0.5      strongswan          100.2.0.0/16, 10.2.0.0/16               connected       
```

Job: Consuming Projects / E2E (lighthouse, operator)
Logs: https://pastebin.com/AWScEvqF

```
2020-11-23T12:41:02.7330133Z [36m[lighthouse]$ [cluster2] grep connected[0m
2020-11-23T12:41:03.0383850Z cluster1-worker                 cluster1                172.17.0.8      libreswan           100.1.0.0/16, 10.1.0.0/16               connected       
2020-11-23T12:41:03.0385082Z cluster2-worker                 cluster2                172.17.0.6      libreswan           100.2.0.0/16, 10.2.0.0/16               connected 
```

Job: Consuming Projects / E2E (lighthouse, operator, libreswan)
Logs: https://pastebin.com/FmMYBQfW

skitt (approved):

dfarrell07:
I added a second commit with better wireguard handling, re-requesting @skitt's review since it was a substantial change.

mangelajo (approved):

mkolesnik (changes requested):
It seems to me this is inflating the testing matrix waaay too much that is necessary.

I feel to test that the flag is passed a simple test can be added with the 3 options, and no need to actually test the whole deployment

dfarrell07:
> It seems to me this is inflating the testing matrix waaay too much that is necessary.
> 
> I feel to test that the flag is passed a simple test can be added with the 3 options, and no need to actually test the whole deployment

How would you test that the flag is passed/consumed correctly?

It seems to me we should exercise the flags we provide in Shipyard. If we're concerned about running too many jobs, then let's merge https://github.com/submariner-io/submariner-operator/pull/902.

.github/workflows/consuming.yml
```diff
@@ -18,9 +18,14 @@ jobs:
       matrix:
         project: ['admiral', 'submariner', 'submariner-operator', 'lighthouse']
         deploytool: ['operator', 'helm']
+        cable_driver: ['libreswan', 'strongswan', 'wireguard']
```
mangelajo, line 21: I hadn't noticed that we make the test matrix quite explode.

I propose we test with only one cable driver, other than the default, `wireguard`?, that way we can always check that such driver is being used.

mangelajo:
> > It seems to me this is inflating the testing matrix waaay too much that is necessary.
> > I feel to test that the flag is passed a simple test can be added with the 3 options, and no need to actually test the whole deployment
> 
> How would you test that the flag is passed/consumed correctly?
> 
> It seems to me we should exercise the main flags we provide in Shipyard. If we're concerned about running too many jobs, then let's merge [submariner-io/submariner-operator#902](https://github.com/submariner-io/submariner-operator/pull/902).

I made a proposal, see the comments

.github/workflows/consuming.yml
mkolesnik: I see no reason, in this particular case, to specify it (as it'll double the amount of jobs).

Instead, I propose a simple "unit test" style test which will see that this was indeed passed properly.
Alternatively, for the time being we can just not test it here, since it is being tested by other projects

.github/workflows/scripts.yml
```diff
@@ -46,15 +46,22 @@ jobs:
       matrix:
         globalnet: ['', 'globalnet']
         deploytool: ['operator', 'helm']
+        cable_driver: ['libreswan', 'strongswan', 'wireguard']
```
mkolesnik, line 49: This shouldn't be here as well.

.github/workflows/consuming.yml
mangelajo: I mean, just having one which won't be the default may be?

But I agree that we cannot grow the test matrix, the e2e is by no ways a cheap test to run.

dfarrell07:
I guess if we're okay with actually-testing all the cable drivers in the main repo (as we're doing now), that's good enough validation.

.github/workflows/consuming.yml
dfarrell07: By this reasoning, why do we test different deploy tools? They are already covered in other repos and they expand the test matrix.

.github/workflows/scripts.yml
dfarrell07: Should I remove the globalnet tests too then, to be consistent with this reasoning?

dfarrell07:
I made the changes @mkolesnik requested, but I don't see how that reasoning is consistent with running any part of our test matrix.

mkolesnik (approved):

PR #367, opened 2020-11-24
Build with buildx, add platform support

skitt:
This will allow building images for different architectures. For the
time being only a single architecture can be built at a time, because
multiarch images can only be pushed to a registry.


mangelajo (approved):

mangelajo:
amazing! :)

tpantelis (approved):

PR #370, opened 2020-11-30
fix: use default subctl relying on the PATH

SteveMattar:
Signed-off-by: Steve Mattar <smattar@redhat.com>

### DIFF
--- a/scripts/shared/lib/deploy_operator
+++ b/scripts/shared/lib/deploy_operator
@@ -10,12 +10,13 @@ readonly SUBM_COLORCODES=blue
 readonly SUBM_ENGINE_IMAGE_REPO=localhost:5000
 readonly SUBM_ENGINE_IMAGE_TAG=${image_tag:-local}
 # shellcheck disable=SC2034 # this variable is used elsewhere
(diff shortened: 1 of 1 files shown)


skitt (approved):

skitt (changes requested):
Ah, you’ll need to change

```
test -x "${SUBCTL}"
```

further down in the file to

```
command -v subctl > /dev/null 2>&1
```

or just drop it entirely since `"${SUBCTL}" version` will fail if the shell can’t find the command (which is what the `test` is checking).

While you’re at it, it would be worth adding a comment above the `SUBCTL=` declaration to document that some projects rely on `subctl` being picked from the `PATH`:

```
# Some projects rely on the default subctl being determined by the PATH
```

SteveMattar:
> Ah, you’ll need to change
> 
> ```
> test -x "${SUBCTL}"
> ```
> 
> further down in the file to
> 
> ```
> command -v subctl > /dev/null 2>&1
> ```

Right it's not a file anymore

mkolesnik (approved):

skitt (approved):

PR #372, opened 2020-11-30
Allow building without buildx

skitt:
We build the initial Shipyard base image without Dapper (to avoid a
chicken-and-egg problem). Since we don't need buildx for that, allow
build_image.sh to fall back to BuildKit if buildx isn't available, so
that we can get things going with plain Docker on the host. (buildx
will be installed in the Dapper image.)


mangelajo (approved):

scripts/shared/build_image.sh
```diff
@@ -48,6 +49,13 @@ fi
 # Rebuild the image to update any changed layers and tag it back so it will be used.
 buildargs_flag='--build-arg BUILDKIT_INLINE_CACHE=1'
 [[ -z "${buildargs}" ]] || buildargs_flag="${buildargs_flag} --build-arg ${buildargs}"
-docker buildx use buildx_builder || docker buildx create --name buildx_builder --use
-docker buildx build --load -t ${local_image} ${cache_flag} -f ${dockerfile} --platform ${platform} ${buildargs_flag} .
-docker tag ${local_image} ${cache_image}
```
mkolesnik, line 53: Was this removed intentionally?

skitt: Ah, no, thanks for spotting it!

scripts/shared/build_image.sh
```diff
@@ -4,14 +4,15 @@ source ${SCRIPTS_DIR}/lib/version
 
 ## Process command line flags ##
 
+default_platform=linux/amd64
```
mkolesnik, line 7: For now it's fine but maybe this should be considering $ARCH in the future.

mkolesnik (approved):
Thanks, I tested and it works

dfarrell07 (approved):

PR #373, opened 2020-12-01
If deployment failed, warn on inotify if necessary

mkolesnik:
Instead of always failing if inotify is too low, only warn about it in
case the deployment failed.

### DIFF
--- a/scripts/shared/clusters.sh
+++ b/scripts/shared/clusters.sh
@@ -161,21 +161,25 @@ function deploy_cluster_capabilities() {
     [[ $prometheus != "true" ]] || deploy_prometheus
 }
 
-### Main ###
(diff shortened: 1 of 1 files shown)


skitt (approved):
LGTM, thanks, the test logs show the function was called after the failure. (For other reviewers, this mustn’t be merged as is, the `return 1` line needs to be removed!)

dfarrell07 (approved):

mkolesnik:
> LGTM, thanks, the test logs show the function was called after the failure. (For other reviewers, this mustn’t be merged as is, the `return 1` line needs to be removed!)

Yes left that there by mistake but it'll be recorded for posterity :)

PR #376, opened 2020-12-07
Stop building images for deploy-latest

skitt:
When deploying the latest release, for the upgrade test, we don't want
to build images: otherwise we end up deploying newly-built images for
those images which are built by a given project.


mkolesnik (approved):
Sounds reasonable to me

dfarrell07 (approved):

PR #377, opened 2020-12-07
Test the same cable drivers as consuming projects

dfarrell07:
In E2E CI for consuming projects, which is non-default and only
triggered if a label is explicitly added, add cable driver to the test
matrix. Only test exactly the combinations that are covered in the CI of
the consuming projects.

### DIFF
--- a/.github/workflows/consuming.yml
+++ b/.github/workflows/consuming.yml
@@ -18,6 +18,7 @@ jobs:
       matrix:
         project: ['admiral', 'submariner', 'submariner-operator', 'lighthouse']
         deploytool: ['operator', 'helm']
+        cabledriver: ['libreswan']
         exclude:
           # Admiral E2E doesn't respect deploy-tool params, as it uses clusters without Submariner
           - project: admiral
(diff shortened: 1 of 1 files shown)


skitt (approved):

.github/workflows/consuming.yml
```diff
@@ -18,17 +18,32 @@ jobs:
       matrix:
         project: ['admiral', 'submariner', 'submariner-operator', 'lighthouse']
         deploytool: ['operator', 'helm']
+        cabledriver: ['libreswan']
```
mkolesnik, line 21: I think this isn't needed, can you try without it?

mkolesnik (changes requested):

.github/workflows/consuming.yml
```diff
@@ -18,17 +18,32 @@ jobs:
...
         exclude:
           # Admiral E2E doesn't respect deploy-tool params, as it uses clusters without Submariner
           - project: admiral
             deploytool: helm
           # Operator and Helm are mutually exclusive, don't try to use Helm in Operator repo
           - project: submariner-operator
             deploytool: helm
+        include:
+          # Parallel the consuming projects' test matrices for cable driver CI coverage
```
mkolesnik, line 30: I'm not sure what this means..
From the name of the PR I thought it's about running in parallel?

Perhaps something like:
"Test additional combinations of cable drivers on select consuming projects"

dfarrell07: It means we test the same combinations here as the consuming projects do in their CI. They aren't just additional, they make the two sets of tests equal. I can try to reword it to be more clear.

.github/workflows/consuming.yml
dfarrell07: I did try without, it is required the matrix var exists before `include`ing it.

nyechiel (approved):

.github/workflows/consuming.yml
mkolesnik: OK good to know since in their guide they seem to indicate you can create vars simply by including them

mkolesnik (approved):

PR #379, opened 2020-12-11
Fetch actual cluster IDs

tpantelis:
There are cases where the kube context name != the actual cluster ID so try to obtain it from the env var passed to the gateway pod.

### DIFF
--- a/test/e2e/framework/framework.go
+++ b/test/e2e/framework/framework.go
@@ -8,6 +8,7 @@ import (
 
 	. "github.com/onsi/gomega"
 	"github.com/pkg/errors"
+	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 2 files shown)


skitt (approved):
Having to introduce a variant for the int parameter gave me pause but I don’t see a way around it!

tpantelis:
> Having to introduce a variant for the int parameter gave me pause but I don’t see a way around it!

yeah.  Thanks for the quick review. I'm still testing this - I didn't see the expected output in the downstream tests.

tpantelis:
> > Having to introduce a variant for the int parameter gave me pause but I don’t see a way around it!
> 
> yeah. Thanks for the quick review. I'm still testing this - I didn't see the expected output in the downstream tests.

Verified it's working.

mangelajo (approved):

vthapar (approved):

dfarrell07 (approved):

test/e2e/framework/framework.go
```diff
@@ -234,6 +237,45 @@ func DetectGlobalnet() {
 	})
 }
 
+func fetchClusterIDs() {
+	for i := range KubeClients {
```
mkolesnik, line 241: nit: I'd restructure it so that this function receives the index and returns the cluster ID
Then when called it can be run in a loop from `BeforeSuite` and the code will be more straightforward and easier to read (also one indentation level less in the function).

We can do it in a follow up patch, though.

mkolesnik (approved):

PR #382, opened 2020-12-14
Use nginx image from quay.io

vthapar:
Due to dockerhub api limits, we run into sporadic failures
when running e2e as it fails to load nginx image.
Use `quay.io/bitnami/nginx:latest` instead to avoid such issues.

### DIFF
--- a/scripts/shared/resources/nginx-demo.yaml
+++ b/scripts/shared/resources/nginx-demo.yaml
@@ -15,7 +15,7 @@ spec:
     spec:
       containers:
         - name: nginx-demo
-          image: nginxinc/nginx-unprivileged:stable-alpine
+          image: quay.io/bitnami/nginx:latest
           ports:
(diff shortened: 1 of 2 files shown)


manosnoam (approved):

skitt (approved):

mkolesnik:
Did you make sure this container can run without needing root permissions?

We switched at some point to `nginx-unprivileged` because otherwise you need root (and IIRC in OpenShift and such the security model doesn't allow it by default)

vthapar:
> Did you make sure this container can run without needing root permissions?
> 
> We switched at some point to `nginx-unprivileged` because otherwise you need root (and IIRC in OpenShift and such the security model doesn't allow it by default)

Noam confirmed this is rootless and works. This image was suggested by him and he already uses this to test on OpenShift.

PR #385, opened 2020-12-17
Add a Submariner image pruning command

skitt:
This defines

        make prune-subm-images

which prunes all Submariner-provided images, ensuring that the next
deployment will pull or build all the images it needs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,6 +45,20 @@ images: package/.image.shipyard-dapper-base
 .DEFAULT_GOAL := lint
 # Shipyard-specific ends
 
+# This removes all Submariner-provided images and all untagged images
+# Use this to ensure you use current images
+prune-images:
+	docker images | grep -E '(admiral|lighthouse|nettest|shipyard|submariner|<none>)' | while read image tag hash _; do \
+	    if [ "$$tag" != "<none>" ]; then \
+	        docker rmi $$image:$$tag; \
+	    else \
+	        docker rmi $$hash; \
+	    fi \
+	done
+
+NON_DAPPER_GOALS += prune-images
+.PHONY: prune-images
+
 include Makefile.dapper
 
 endif
--- a/Makefile.inc
+++ b/Makefile.inc
@@ -90,3 +90,16 @@ endif
 
 CODEOWNERS: CODEOWNERS.in
 	$(SCRIPTS_DIR)/gen-codeowners
+
+# This removes all Submariner-provided images and all untagged images
+# Use this to ensure you use current images
+# Copied from Makefile to provide this everywhere (until we can share
+# non-Dapper goals across projects)
+prune-images:
+	docker images | grep -E '(admiral|lighthouse|nettest|shipyard|submariner|<none>)' | while read image tag hash _; do \
+	    if [ "$$tag" != "<none>" ]; then \
+	        docker rmi $$image:$$tag; \
+	    else \
+	        docker rmi $$hash; \
+	    fi \
+	done
--- a/Makefile.dapper
+++ b/Makefile.dapper
@@ -12,7 +12,7 @@
 
 # Only run command line goals in dapper (except things that have to run outside of dapper).
 # Otherwise, make applies this rule to various files and tries to build them in dapper (which doesn't work, obviously).
-$(filter-out .dapper images shell,$(MAKECMDGOALS)): .dapper
+$(filter-out .dapper images shell $(NON_DAPPER_GOALS),$(MAKECMDGOALS)): .dapper
 	+./.dapper -m bind make $@ $(MAKEFLAGS)
 
 shell: .dapper


Makefile
```diff
@@ -45,6 +45,20 @@ images: package/.image.shipyard-dapper-base
 .DEFAULT_GOAL := lint
 # Shipyard-specific ends
 
+# This removes all Submariner-provided images and all untagged images
+# Use this to ensure you use current images
+prune-subm-images:
```
mkolesnik, line 50:
```suggestion
prune-images:
```

Makefile.inc
```diff
@@ -94,3 +94,16 @@ endif
 
 CODEOWNERS: CODEOWNERS.in
 	$(SCRIPTS_DIR)/gen-codeowners
+
+# This removes all Submariner-provided images and all untagged images
+# Use this to ensure you use current images
+# Copied from Makefile to provide this everywhere (until we can share
+# non-Dapper goals across projects)
+prune-subm-images:
```
mkolesnik, line 102: Won't this try to remove the temp image used by `make shell` and fail?

Makefile
```diff
@@ -45,6 +45,20 @@ images: package/.image.shipyard-dapper-base
...
+	docker images | grep -E '(admiral|lighthouse|nettest|shipyard|submariner|<none>)' | while read image tag hash _; do \
+	    if [ "$$tag" != "<none>" ]; then \
+	        docker rmi $$image:$$tag; \
```
mkolesnik, line 53: Why not just o something like `docker rmi $$hash || :` all the time?

Makefile.inc
```diff
@@ -94,3 +94,16 @@ endif
 
 CODEOWNERS: CODEOWNERS.in
 	$(SCRIPTS_DIR)/gen-codeowners
+
+# This removes all Submariner-provided images and all untagged images
+# Use this to ensure you use current images
+# Copied from Makefile to provide this everywhere (until we can share
```
mkolesnik, line 100: What's the problem with running it in dapper always?

mkolesnik (changes requested):

Makefile.inc
skitt: Yes, but we’re not running with `set -e` so the other prunable images are still pruned.

Makefile.inc
skitt: We can’t remove the Dapper image if we’re running in Dapper.

Makefile
skitt: Right, we might as well just use the hash. We don’t need `|| :` here.

mangelajo (approved):

tpantelis (approved):

Makefile
skitt: Oh so I ran into the reason I did this originally: images can’t be deleted by reference if multiple tags point to them. So tagged images need to be deleted by tag first (and the image is deleted once the last tag referencing it is deleted).

dfarrell07 (approved):

nyechiel:
@skitt please let's also not forget to document this new command

skitt:
> @skitt please let's also not forget to document this new command

https://github.com/submariner-io/submariner-website/pull/405

This still needs @mkolesnik to update his review.

Makefile.inc
mkolesnik: Right kind of chicken and egg, though when tested locally on *submariner* it did remove the shipyard base image (but failed to remove the container image since it's running, maybe thats what you were seeing)?

skitt: Yes, the Dapper image, not the Shipyard base image.

mkolesnik: Theoretically it can be identified as such
```bash
set +m
shopt -s lastpipe
docker ps | grep $HOSTNAME | read _ dapper_image _
```

And then you can skip trying to delete `$dapper_image`

mkolesnik (approved):
Small suggestion but looks good otherwise :)

Makefile.inc
skitt: That’s not what I’m trying to fix. I don’t care about skipping the image we’re running in, `docker rmi` does that for us, and since we’re not running with `set -e`, it doesn’t break out of the task. I would like to support running all this without Dapper everywhere at some point so that we _can_ delete the Dapper image.

nyechiel:
@skitt I was looking for this PR as I couldn't remember the exact syntax of the command.. can you please document this as part of https://submariner.io/development/shipyard/?

skitt:
> @skitt I was looking for this PR as I couldn't remember the exact syntax of the command.. can you please document this as part of https://submariner.io/development/shipyard/?

@nyechiel it’s documented in https://submariner.io/development/building-testing/#building but I can mention it in the Shipyard docs too.

nyechiel:
> > @skitt I was looking for this PR as I couldn't remember the exact syntax of the command.. can you please document this as part of https://submariner.io/development/shipyard/?
> 
> @nyechiel it’s documented in https://submariner.io/development/building-testing/#building but I can mention it in the Shipyard docs too.

Oh here it is :) I think it's fine, thanks for the pointer!

Issue #335, opened 2020-10-13
Shipyard support for deploying OVN


sridhargaddam:
As part of this task, we want to implement necessary changes in Shipyard to install OVN CNI.

https://github.com/ovn-org/ovn-kubernetes/blob/master/contrib/kind.sh

See:  https://github.com/submariner-io/enhancements/pull/11
Related-Issue: https://github.com/submariner-io/submariner/issues/778

PR #387, opened 2020-12-18
Initial support for OVN in Kind

vthapar:
Use `make clusters using=ovn` to deploy kind with ovn

* Installs j2cli to dapper image
* Adds `using=ovn` flag
* Add scripts to deploy ovn on kind

Currently this copies over ovn-kind scripts and templates from
 * https://github.com/ovn-org/ovn-kubernetes/blob/master/contrib/
 * https://github.com/ovn-org/ovn-kubernetes/tree/master/dist

Fixes: #335

### DIFF
--- a/ovn/daemonset.sh
+++ b/ovn/daemonset.sh
@@ -1,0 +1,335 @@
+#!/bin/bash
+#set -x
+
+#Always exit on errors
+set -e
+
+# The script renders j2 templates into yaml files in .${OVN_YAML_DIR}/
+
+# ensure j2 renderer installed
+## pip freeze | grep j2cli || pip install j2cli[yaml] --user
+export PATH=~/.local/bin:$PATH
+
+OVN_IMAGE=""
+OVN_IMAGE_PULL_POLICY=""
+OVN_NET_CIDR=""
+OVN_SVC_DIDR=""
+OVN_K8S_APISERVER=""
+OVN_GATEWAY_MODE=""
+OVN_GATEWAY_OPTS=""
+OVN_DB_REPLICAS=""
+OVN_MTU=""
+OVN_SSL_ENABLE=""
+KIND=""
+OVN_UNPRIVILEGED_MODE=""
+MASTER_LOGLEVEL=""
+NODE_LOGLEVEL=""
+DBCHECKER_LOGLEVEL=""
+OVN_LOGLEVEL_NORTHD=""
+OVN_LOGLEVEL_NB=""
+OVN_LOGLEVEL_SB=""
+OVN_LOGLEVEL_CONTROLLER=""
+OVN_LOGLEVEL_NBCTLD=""
+OVNKUBE_LOGFILE_MAXSIZE=""
+OVNKUBE_LOGFILE_MAXBACKUPS=""
+OVNKUBE_LOGFILE_MAXAGE=""
+OVN_MASTER_COUNT=""
+OVN_REMOTE_PROBE_INTERVAL=""
+OVN_HYBRID_OVERLAY_ENABLE=""
+OVN_DISABLE_SNAT_MULTIPLE_GWS=""
+OVN_MULTICAST_ENABLE=""
+OVN_EGRESSIP_ENABLE=
+OVN_V4_JOIN_SUBNET=""
+OVN_V6_JOIN_SUBNET=""
+
+# Parse parameters given as arguments to this script.
+while [ "$1" != "" ]; do
+  PARAM=$(echo $1 | awk -F= '{print $1}')
+  VALUE=$(echo $1 | cut -d= -f2-)
+  case $PARAM in
+  --image)
+    OVN_IMAGE=$VALUE
+    ;;
+  --image-pull-policy)
+    OVN_IMAGE_PULL_POLICY=$VALUE
+    ;;
+  --gateway-mode)
+    OVN_GATEWAY_MODE=$VALUE
+    ;;
+  --gateway-options)
+    OVN_GATEWAY_OPTS=$VALUE
+    ;;
+  --net-cidr)
+    OVN_NET_CIDR=$VALUE
+    ;;
+  --svc-cidr)
+    OVN_SVC_CIDR=$VALUE
+    ;;
+  --k8s-apiserver)
+    OVN_K8S_APISERVER=$VALUE
+    ;;
+  --db-replicas)
+    OVN_DB_REPLICAS=$VALUE
+    ;;
+  --mtu)
+    OVN_MTU=$VALUE
+    ;;
+  --kind)
+    KIND=true
+    ;;
+  --ovn-unprivileged-mode)
+    OVN_UNPRIVILEGED_MODE=$VALUE
+    ;;
+  --master-loglevel)
+    MASTER_LOGLEVEL=$VALUE
+    ;;
+  --node-loglevel)
+    NODE_LOGLEVEL=$VALUE
+    ;;
+  --dbchecker-loglevel)
+    DBCHECKER_LOGLEVEL=$VALUE
+    ;;
+  --ovn-loglevel-northd)
+    OVN_LOGLEVEL_NORTHD=$VALUE
+    ;;
+  --ovn-loglevel-nb)
+    OVN_LOGLEVEL_NB=$VALUE
+    ;;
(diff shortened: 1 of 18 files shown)


ovn/daemonset.sh
```diff
@@ -0,0 +1,335 @@
...
+OVN_HYBRID_OVERLAY_ENABLE=""
+OVN_DISABLE_SNAT_MULTIPLE_GWS=""
+OVN_MULTICAST_ENABLE=""
+OVN_EGRESSIP_ENABLE=
+OVN_V4_JOIN_SUBNET=""
+OVN_V6_JOIN_SUBNET=""
+
+# Parse parameters given as arguments to this script.
+while [ "$1" != "" ]; do
+  PARAM=$(echo $1 | awk -F= '{print $1}')
```
skitt, line 47:
```suggestion
  PARAM=${1%%=*}
```

ovn/daemonset.sh
```diff
@@ -0,0 +1,335 @@
...
+  VALUE=$(echo $1 | cut -d= -f2-)
```
skitt, line 48:
```suggestion
  VALUE=${1#*=}
```

ovn/daemonset.sh
```diff
@@ -0,0 +1,335 @@
...
+OVN_REMOTE_PROBE_INTERVAL=""
+OVN_HYBRID_OVERLAY_ENABLE=""
+OVN_DISABLE_SNAT_MULTIPLE_GWS=""
+OVN_MULTICAST_ENABLE=""
+OVN_EGRESSIP_ENABLE=
+OVN_V4_JOIN_SUBNET=""
+OVN_V6_JOIN_SUBNET=""
+
+# Parse parameters given as arguments to this script.
+while [ "$1" != "" ]; do
```
skitt, line 46:
```suggestion
while [ "$#" -gt 0 ]; do
```

ovn/daemonset.sh
```diff
@@ -0,0 +1,335 @@
...
+ovn_sb_raft_port=${OVN_SB_RAFT_PORT:-6644}
+echo "ovn_sb_raft_port: ${ovn_sb_raft_port}"
+ovn_multicast_enable=${OVN_MULTICAST_ENABLE}
+echo "ovn_multicast_enable: ${ovn_multicast_enable}"
+ovn_v4_join_subnet=${OVN_V4_JOIN_SUBNET}
+echo "ovn_v4_join_subnet: ${ovn_v4_join_subnet}"
+ovn_v6_join_subnet=${OVN_V6_JOIN_SUBNET}
+echo "ovn_v6_join_subnet: ${ovn_v6_join_subnet}"
+
+ovn_image=${image} \
```
skitt, line 242: Wouldn’t it be simpler to export all the relevant variables?

skitt (changes requested):

ovn/daemonset.sh
mangelajo: This is a copy-over from the ovn-kubernetes repository

I think we should not focus on improving those here, but we should instead try enhance in their repo, and use them from here as soon as we can.

ovn/daemonset.sh
vthapar: @skitt These files are taken mostly as-is from `ovn-kubernetes` Plan is to refactor all this to fit within our framework in subsequent PRs/commits. Wanted to get this initial PR up and ready so that @mangelajo and @sridhargaddam have something to work with. We need some changes in subm and operator before we can deploy subm on top of kind+ovn, so crucial to have something working asap.

Also, with this patch, we can enable a job that does `make clusters using=ovn` Would be good to have that in place before we start refactor.

skitt (approved):
Please ignore my requested changes, I’ll suggest PRs in the upstream repos.

scripts/shared/clusters.sh
```diff
@@ -11,6 +11,7 @@ DEFINE_boolean 'globalnet' false "Deploy with operlapping CIDRs (set to 'true' t
 DEFINE_boolean 'registry_inmemory' true "Run local registry in memory to speed up the image loading."
 DEFINE_string 'cluster_settings' '' "Settings file to customize cluster deployments"
 DEFINE_string 'timeout' '5m' "Timeout flag to pass to kubectl when waiting (e.g. 30s)"
+DEFINE_boolean 'ovn' false "Deploy with OVN as CNI"
```
mkolesnik, line 14: This goes against what we currently have and rolls us back to the days of specifying CNI on the command line.

Why not specify this in cluster_settings?

mkolesnik (changes requested):
Would it be possible to clone the `ovn-kuberentes` and just call the scripts there or did you do any changes to them?

I think it'll be much better like that (if we cant figure out how to deploy it on kind ourselves), it'll save us having to maintain their code in our repo and will allow us to easily update should we need it.

vthapar:
> Would it be possible to clone the `ovn-kuberentes` and just call the scripts there or did you do any changes to them?
> 
> I think it'll be much better like that (if we cant figure out how to deploy it on kind ourselves), it'll save us having to maintain their code in our repo and will allow us to easily update should we need it.

There are some changes to the scripts to make it work off our shipyard dapper image. Discussed with @mangelajo earlier and plan is to make it work in our setup first. Then we figure out changes to be done in `ovn-kubernetes` in such a way that we can consume it by just cloning and contribute it there.

scripts/shared/clusters.sh
vthapar: > This goes against what we currently have and rolls us back to the days of specifying CNI on the command line.
> 
> Why not specify this in cluster_settings?

Will it be possible to add CNI to job matrix if we use cluster_settings? I yes, then we don't need extra flag.

mangelajo (approved):

mkolesnik:
> > Would it be possible to clone the `ovn-kuberentes` and just call the scripts there or did you do any changes to them?
> > I think it'll be much better like that (if we cant figure out how to deploy it on kind ourselves), it'll save us having to maintain their code in our repo and will allow us to easily update should we need it.
> 
> There are some changes to the scripts to make it work off our shipyard dapper image. Discussed with @mangelajo earlier and plan is to make it work in our setup first. Then we figure out changes to be done in `ovn-kubernetes` in such a way that we can consume it by just cloning and contribute it there.

What we could do is clone and apply it dynamically (if it's not too much changes) using `sed` or `cp`, this way we have a good indication what's different so that we can easily push it to ovn-kubernetes later.

If we copy a lot of things in our repo its hard to tell what are the changes and whats just plain copied from ovn-k8s

skitt:
> What we could do is clone and apply it dynamically (if it's not too much changes) using `sed` or `cp`, this way we have a good indication what's different so that we can easily push it to ovn-kubernetes later.
> 
> If we copy a lot of things in our repo its hard to tell what are the changes and whats just plain copied from ovn-k8s

Isn’t this one of the things git is good at? Instead of re-inventing quilt, we could add a fork of the ovn-kubernetes repo as a submodule...

vthapar:
> > > Would it be possible to clone the `ovn-kuberentes` and just call the scripts there or did you do any changes to them?
> > > I think it'll be much better like that (if we cant figure out how to deploy it on kind ourselves), it'll save us having to maintain their code in our repo and will allow us to easily update should we need it.
> > 
> > 
> > There are some changes to the scripts to make it work off our shipyard dapper image. Discussed with @mangelajo earlier and plan is to make it work in our setup first. Then we figure out changes to be done in `ovn-kubernetes` in such a way that we can consume it by just cloning and contribute it there.
> 
> What we could do is clone and apply it dynamically (if it's not too much changes) using `sed` or `cp`, this way we have a good indication what's different so that we can easily push it to ovn-kubernetes later.
> 
> If we copy a lot of things in our repo its hard to tell what are the changes and whats just plain copied from ovn-k8s
That was my initial plan but changes required to make it work kept increasing. This should give an idea of changes in `kind.sh` which has most changes. 
```
git diff  --shortstat ../../ovn-org/ovn-kubernetes/contrib/kind.sh ovn/kind.sh 
 1 file changed, 70 insertions(+), 28 deletions(-)
```
I can look into reducing it further by about 40% now that I've got it working, but would still be too many `sed` or `cp`. A simple diff can always show the changes we've done.

vthapar:
Replaced by #401

PR #395, opened 2021-01-04
Fix link to job results in the periodic link check

skitt:
The periodic link check files issues with links to
submariner-operator; this fixes that.


dfarrell07 (approved):
Oops, good catch, thanks @skitt!

tpantelis (approved):

PR #399, opened 2021-01-06
Add ovn-kubernetes submodule

vthapar:


mkolesnik (changes requested):
Should it be under ` scripts/shared/ovn-kubernetes ` though?

Consuming projects wont be able to utilize it since itll be inside the docker image.

Perhaps its best that the script just clone our fork?

vthapar:
> Should it be under `scripts/shared/ovn-kubernetes` though?
> 
> Consuming projects wont be able to utilize it since itll be inside the docker image.
> 
> Perhaps its best that the script just clone our fork?

I cloned here coz this is where rest of our kind scripts are. I think making script clone the fork works fine enough as we only need this for cluster/deploy scripts, at least for now.

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "scripts/shared/ovn-kubernetes"]
+	path = scripts/shared/ovn-kubernetes
```
mkolesnik, line 2: Hmm although its comfortable to just stick this in there, it seems wrong as it's not a script.

I'd create a new directory entirely like `3rd-party` or something under the root directory (or maybe simply `submodules`) and stick it there.

Then, in the Dockerfile for the base image we can have a directive to copy this to `/go/src/github.com/submariner-io/` or some other location (`${SHIPYARD_DIR}/submodules` can also work).

.dockerignore
```diff
@@ -0,0 +1 @@
+**/.git
```
mkolesnik, line 1: While nice this is actually not needed as `git` seems smart enough not to initialize it in submodules. We could keep it as a safeguard though.

The real heavy weight is in `vendor` directory under `ovn-kubernetes/go-controller`. We could ignore that directory and init it when its needed using `go mod vendor`. This shaves off ~54M off the added size (~57M).

mkolesnik (changes requested):

.dockerignore
vthapar: Currently we're not doing any go builds on kubernetes code, just using the scripts. Will add, good to have it.

package/Dockerfile.shipyard-dapper-base
```diff
@@ -7,7 +7,7 @@ ENV DAPPER_HOST_ARCH=amd64 SHIPYARD_DIR=/opt/shipyard SHELL=/bin/bash
 ENV HOST_ARCH=${DAPPER_HOST_ARCH} ARCH=${DAPPER_HOST_ARCH} PATH=/go/bin:/usr/local/go/bin:$PATH \
     GOLANG_ARCH_amd64=amd64 GOLANG_ARCH_arm=armv6l GOLANG_ARCH=GOLANG_ARCH_${DAPPER_HOST_ARCH} \
     GOPATH=/go GO111MODULE=on GOFLAGS=-mod=vendor GOPROXY=https://proxy.golang.org \
-    SCRIPTS_DIR=${SHIPYARD_DIR}/scripts
+    SCRIPTS_DIR=${SHIPYARD_DIR}/scripts OVN_DIR=${SHIPYARD_DIR}/submodules/ovn-kubernetes
```
skitt, line 10: Inside the shared directory, we don’t care that it’s a submodule:

```suggestion
    SCRIPTS_DIR=${SHIPYARD_DIR}/scripts OVN_DIR=${SHIPYARD_DIR}/ovn-kubernetes
```

skitt (changes requested):

.github/workflows/release.yml
```diff
@@ -14,6 +14,8 @@ jobs:
     steps:
       - name: Check out the repository
         uses: actions/checkout@v2
+        with:
```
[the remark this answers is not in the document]
vthapar: done

skitt (approved):

mkolesnik (approved):

dfarrell07:
This was backported in #436.

PR #407, opened 2021-01-13
ovn-kubernetes support cherry picks to release-0.8

mangelajo:
Includes commits:

9368ae5 Import ovn-daemonset-f image
fcb9773 Deploy kind with OVN CNI (#401)
45e07b2 Add ovn-kubernetes submodule
b580847 Add makefile support for ovn

PR #408, opened 2021-01-14
Disable Markdown link checker for modified files

mangelajo:
As seen in [1] the check-modified-files-only flag doesn't work properly
making the stable branch "red".

I propose we disable this for now at least until it's resolved.

Backport-To: release-0.8

[1] https://github.com/submariner-io/shipyard/pull/407/checks?check_run_id=1696280421

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -39,19 +39,6 @@ jobs:
       - name: Run golangci-lint
         run: make golangci-lint
 
-  markdown-link-check:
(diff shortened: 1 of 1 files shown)


mkolesnik (approved):
Weird that it doesn't work, but makes sense to disable it for stable for now.

skitt (changes requested):
Let’s configure the linter correctly instead: https://github.com/submariner-io/submariner/pull/1085

PR #409, opened 2021-01-14
Don't tag as `devel` by default

mkolesnik:
Instead, we should set the tag to the branch name that's being merged
to, in accordance with the stable release policy.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -5,6 +5,7 @@ on:
   push:
     branches:
       - master
+      - release-*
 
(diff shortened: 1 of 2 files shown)


mkolesnik:
We should probably rename `master` to `devel` prior to merging this

mangelajo (approved):

tpantelis (approved):

dfarrell07:
This was backported in https://github.com/submariner-io/shipyard/pull/436.

PR #420, opened 2021-01-25
Dump cluster info before deleting clusters

skitt:
When a cluster deployment fails, dump the cluster info before deleting
the cluster. This will help debugging deployment issues when the
deployment gets far enough for the Kubernetes API to be reachable.


scripts/shared/clusters.sh
```diff
@@ -100,6 +100,7 @@ function create_kind_cluster() {
     ( deploy_cluster_capabilities; ) &
     if ! wait $! ; then
         echo "Failed to deploy cluster capabilities, removing the cluster"
+        kubectl cluster-info dump
```
mkolesnik: perhaps its better to dump to stderr?

mangelajo: What's the difference on dumping to stderr?

mangelajo (approved):
but watch out for mike's comments.

tpantelis (approved):

mkolesnik (approved):

sridhargaddam (approved):

PR #421, opened 2021-01-26
Create prepare-host github actions

mangelajo:


.github/workflows/consuming.yml
```diff
@@ -39,17 +39,9 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Install WireGuard kernel module (only in WireGuard jobs)
```
mkolesnik, line 41: Should probably change the description :)

gh-actions/prepare-host/action.yaml
```diff
@@ -0,0 +1,24 @@
+name: 'Prepare host'
```
mkolesnik, line 1: Generally this is only ever used for E2E, can we just fold it into the "Run E2E" shared action?

test/e2e/framework/network_pods.go
```diff
@@ -282,7 +282,7 @@ func (np *NetworkPod) buildThroughputClientPod() {
 					Name:            "nettest-client-pod",
```
mkolesnik, line 282: Why was this file changed?

mkolesnik (changes requested):
Interesting :)

test/e2e/framework/network_pods.go
mangelajo: PEBKAK

gh-actions/prepare-host/action.yaml
mangelajo: What do you mean, wait...

mkolesnik: We only need to "prepare host" before we run the E2E.
I think this can all be in a shared "run E2E" shared action (and if necessary we can split it out in the future though I doubt it will be).

PR #425, opened 2021-01-27
Add E2E testing to composite actions

mkolesnik:
Preparing hosts is an implementation detail of E2E testiong, folded it
into the E2E testing actions.
Also added an upgrade-e2e composite action.

Unfortunately composite actions can't use other actions, so some
duplication still exists. Hopefully this will be addressed in the
future.


gh-actions/e2e/action.yaml
```diff
@@ -22,11 +31,17 @@ runs:
         free -h
         echo "::endgroup::"
 
-    - name: Install any cable driver specific modules
+    - name: Install wireguard specific modules
```
dfarrell07, line 34:
```suggestion
    - name: Install Wireguard-specific modules
```

skitt (approved):

gh-actions/e2e/action.yaml
skitt: It’s really “WireGuard” ;-).

dfarrell07 (approved):

tpantelis (approved):

PR #463, opened 2021-03-01
Import the right submariner gateway and latest images 

mangelajo:

### DIFF
--- a/scripts/shared/deploy.sh
+++ b/scripts/shared/deploy.sh
@@ -49,6 +49,7 @@ bash -c "curl -Ls https://get.submariner.io | VERSION=${CUTTING_EDGE} DESTDIR=/g
 import_image quay.io/submariner/nettest
 import_image quay.io/submariner/submariner-operator ${image_tag}
 import_image quay.io/submariner/submariner ${image_tag}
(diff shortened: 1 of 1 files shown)


nyechiel (approved):

SteveMattar (approved):

mangelajo:
Dependent projects still fail because the shipyard go code needs to be updated in those repos after the submariner-gateway rename.

mkolesnik (changes requested):
Let me suggest a different approach which i think would be better. I'll send a patch shortly

mangelajo:
> Let me suggest a different approach which i think would be better. I'll send a patch shortly

ack, but let's be quick, we have to release M1 , we are 2 weeks late already.

If it starts to look complex or risky, please let's merge this one and improve it for M2

mangelajo:
@nyechiel @SteveMattar  can you move votes to this one? https://github.com/submariner-io/shipyard/pull/465

mkolesnik:
Propose #465 instead which offers a cleaner solution and can be used easily in consuming projects

mangelajo:
Handled in https://github.com/submariner-io/shipyard/pull/465

PR #470, opened 2021-03-08
Revert "Add clang and enable -msan"

skitt:
Reverts submariner-io/shipyard#312

We don’t need eBPF, and adding 200MiB just for -msan seems overkill.


mangelajo (approved):
Thank you @skitt

mangelajo:
@skitt this is missing the DCO bit of the revert

PR #478, opened 2021-03-09
Add various "make clean" targets

skitt:
"make clean-generated" removes generated files (by relying on the fact
that we tell git to ignore them).

"make clean-clusters" removes running clusters (this is the old "make
cleanup", which is preserved for backwards compatibility).

"make clean" cleans everything.

### DIFF
--- a/Makefile.inc
+++ b/Makefile.inc
@@ -45,8 +45,17 @@ endif
 
 # Shipyard provided targets
 
-cleanup:
+# clean cleans everything (running clusters, generated files ...)
+clean: clean-clusters clean-generated
+
+# clean-generated removes files we generated
+clean-generated:
+	git clean -X -f
+
+# clean-clusters removes running clusters
+clean-clusters:
 	$(SCRIPTS_DIR)/cleanup.sh
+cleanup: clean-clusters
 
 clusters:
 	$(SCRIPTS_DIR)/clusters.sh $(CLUSTERS_ARGS)
--- a/.github/workflows/clean-tests.yml
+++ b/.github/workflows/clean-tests.yml
@@ -1,0 +1,42 @@
+---
+name: Clean Target Verification
+
+on:
+  pull_request:
+
+jobs:
+  clean-clusters:
+    name: Cluster Clean-up
+    timeout-minutes: 20
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out the repository
+        uses: actions/checkout@v2
+
+      - name: Deploy clusters
+        env:
+          CLUSTERS_ARGS: --timeout 1m
+        run: make clusters
+
+      - name: Clean up clusters
+        run: make clean-clusters
+
+      - name: Check that clusters are gone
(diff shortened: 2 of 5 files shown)


mangelajo (approved):

mkolesnik (approved):
Great idea :)

Makefile.inc
```diff
@@ -45,6 +45,9 @@ endif
 
 # Shipyard provided targets
 
+clean:
+	git clean -X -f
+
 cleanup:
```
dfarrell07: It's a bit confusing to have both "clean" an "cleanup" targets that are totally different. Maybe we can think of some more clear names?

dfarrell07: Per discussion on the automation sync, I think the consensus was to use `make clean` for both types of cleaning and `make clean-clusters`/`make clean-git` or similar for specifically addressing the two types of cleaning. That should give a simple UX to casual users and still enable use cases for experts.

dfarrell07 (changes requested):
(see inline comment)

mangelajo (approved):

mangelajo:
@dfarrell07 please re-review :-)

Makefile.inc
```diff
@@ -45,9 +45,20 @@ endif
 
 # Shipyard provided targets
 
-cleanup:
+# clean cleans everything (running clusters, generated files ...)
+clean: clean-clusters clean-git
```
mkolesnik:
```suggestion
clean: clean-clusters clean-generated
```

mkolesnik (dismissed):
Perhaps also you can add a test for it to make sure it works as expected?

dfarrell07 (approved):

Makefile.inc
```diff
@@ -45,9 +45,20 @@ endif
...
+# clean-generated removes files we generated
+clean-generated:
+	git clean -X -f
+
+# clean-clusters removes running clusters
+clean-clusters:
 	$(SCRIPTS_DIR)/cleanup.sh
 
+# cleanup is the historical clean-clusters
+cleanup: clean-clusters
```
tpantelis, line 66: I propose we just remove this. We could also have it just print a message that it's deprecated and then remove later but I think that would be overkill.

skitt: Done.

dfarrell07 (approved):

tpantelis (approved):

Makefile.inc
skitt: It turns out other projects need this, so it’s still in for now, but I’ll remove it ASAP.

mangelajo (approved):
Wow, verification and all, cool :) is it time-limited to 1 minute deploy of clusters?

skitt:
> Wow, verification and all, cool :) is it time-limited to 1 minute deploy of clusters?

I reckoned that since we do a deployment anyway, spending the time to do it again and check everything is cleaned up was OK... But we could combine the two tests!

skitt:
@mkolesnik could you re-review?

PR #486, opened 2021-03-12
Fix consuming project E2E, use using= flag for GHA

dfarrell07:
The var-specific GHA inputs were recently removed, which broke these
tests. Move to the new GHA input, the using= flag.


tpantelis (approved):

mangelajo (approved):
oops self breakage

skitt (approved):

PR #494, opened 2021-03-22
Show previous logs on terminated pods

skitt:
When pods are hitting errors, we want the post-mortem to show the
previous logs, instead of nothing.

### DIFF
--- a/scripts/shared/post_mortem.sh
+++ b/scripts/shared/post_mortem.sh
@@ -18,8 +18,13 @@ function print_pods_logs() {
 
     print_section "** Pods logs for NS $namespace using selector '$selector' **"
     for pod in $(kubectl get pods --selector="$selector" -n "$namespace" -o jsonpath='{.items[*].metadata.name}'); do
-        print_section "*** $pod ***"
-        kubectl -n $namespace logs $pod
+        if [ "$(kubectl get pods -n $namespace $pod -o jsonpath='{.status.containerStatuses[*].ready}')" != true ]; then
+            print_section "*** $pod (terminated) ***"
+            kubectl -n $namespace logs -p $pod
+        else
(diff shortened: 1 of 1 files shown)


scripts/shared/post_mortem.sh
```diff
@@ -18,8 +18,13 @@ function print_pods_logs() {
 
     print_section "** Pods logs for NS $namespace using selector '$selector' **"
     for pod in $(kubectl get pods --selector="$selector" -n "$namespace" -o jsonpath='{.items[*].metadata.name}'); do
-        print_section "*** $pod ***"
-        kubectl -n $namespace logs $pod
+        if [ "$(kubectl get pods -n $namespace $pod -o jsonpath='{.status.containerStatuses[*].lastState.*.reason}')" = "Error" ]; then
```
mangelajo: Why don't we do != Running ? (I believe it could also be CrashBackoff loop or something like that)

Why don't we simply always use -p?

mangelajo (changes requested):
Great idea, left you one comment.

scripts/shared/post_mortem.sh
skitt: > Why don't we do != Running ? (I believe it could also be CrashBackoff loop or something like that)

I have a container in CrashLoopBackoff and its reason is given as “Error”. But yes, `!= "Running` would be better.

> Why don't we simply always use -p?

That doesn’t show the logs of the current container when it’s actually running, does it? If we have a running container we want to see those logs, not older ones ;-).

tpantelis (approved):

dfarrell07 (approved):

skitt:
@mangelajo could you re-review?

mangelajo (approved):

PR #495, opened 2021-03-23
Warn on inconsistent Makefile.dapper

skitt:
If the local Makefile.dapper doesn't match the version stored in the
Shipyard base image, display a warning.


skitt:
I’m open to ideas of how to do this differently; for example, we could update the local Makefile.dapper automatically (from the base image), and error out (since the whole thing needs to be restarted). Do people ever need to test local changes to Makefile.dapper?

mangelajo (approved):

sridhargaddam (approved):

tpantelis (approved):

PR #487, opened 2021-03-12
Add k8s_version to shared E2E GHA options

dfarrell07:
Allow setting the Kubernetes version used in clusters deployed by the
shared E2E GitHub Action. Pass the version via CLUSTER_ARGS to the
clusters.sh script.

Builds on fix in https://github.com/submariner-io/shipyard/pull/486.

PR #496, opened 2021-03-23
Test additional K8s versions

dfarrell07:
Test recent patch-releases for the 1.18, 1.19, and 1.20 minor releases.

The recentness of versions available to test are limited by kindest/node
image available.

https://hub.docker.com/r/kindest/node/tags

Only test in the submariner-operator repository for now, to give maximum
coverage for minimal jobs.

### DIFF
--- a/.github/workflows/consuming.yml
+++ b/.github/workflows/consuming.yml
@@ -19,6 +19,7 @@ jobs:
         project: ['admiral', 'submariner', 'submariner-operator', 'lighthouse']
         deploytool: ['operator', 'helm']
         cabledriver: ['libreswan']
+        k8s_version: ['1.17.17']
         exclude:
           # Admiral E2E doesn't respect deploy-tool params, as it uses clusters without Submariner
           - project: admiral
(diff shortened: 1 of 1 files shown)


dfarrell07:
Builds on https://github.com/submariner-io/shipyard/pull/487.

.github/workflows/consuming.yml
```diff
@@ -31,6 +32,15 @@ jobs:
           - project: submariner
             cabledriver: wireguard
             deploytool: operator
+            k8s_version: 1.17.17
```
mkolesnik, line 35:
```suggestion
```

Seems redundant?

dfarrell07: Here's a PR to demo the failure that results from removing this:

https://github.com/dfarrell07/shipyard/pull/14/checks?check_run_id=2185768163

skitt: Ah, because if it’s *not* specified, we don’t provide a default, so the kind default is used instead (1.20.2).

dfarrell07: Yeah, the strange bit is that we do provide defaults that seem to work in every case except this one. 

If the new E2E GHA feature isn't used at all, the default works:

https://github.com/submariner-io/submariner/runs/2189218309?check_suite_focus=true

That should be caused by the default here:

https://github.com/submariner-io/shipyard/blob/devel/gh-actions/e2e/action.yaml#L7

If no k8s_version flag is passed to clusters.sh (skipping the GHA), it also handles the default correctly:

https://github.com/submariner-io/shipyard/pull/496/checks?check_run_id=2187434544

That should be caused by the default here:

https://github.com/submariner-io/shipyard/blob/devel/scripts/shared/clusters.sh#L6

dfarrell07: I think it's something weird about how GHA matrix includes work in this case. The default in the GHA is not being used which results in [setting `--k8s_version=""`](https://github.com/submariner-io/shipyard/blob/devel/gh-actions/e2e/action.yaml#L45), which I guess overrides the default in clusters.sh with null.

Anyway, can we move ahead? I think if there's a fix to be made, it's pretty low prio and shouldn't be made here.

mangelajo (approved):

mkolesnik (approved):

PR #499, opened 2021-03-25
Add docker and network pod functions to e2e framework

mkimuram:
This PR is to add docker and network pod functions to e2e framework. These functions are originally needed for https://github.com/submariner-io/submariner/pull/1202 , however, as a result of discussion, we came to conclusion that it would be nice to add them to this framework to make room to leverage from other tests.

### DIFF
--- a/test/e2e/framework/docker.go
+++ b/test/e2e/framework/docker.go
@@ -1,0 +1,76 @@
+/*
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package framework
+
+import (
+	"bytes"
+	"fmt"
+	"os/exec"
+	"strings"
+
+	. "github.com/onsi/gomega"
+)
+
+type Docker struct {
+	Name string
+}
+
+func New(name string) *Docker {
+	return &Docker{Name: name}
+}
+
+func (d *Docker) GetIP(networkName string) string {
+	var stdout bytes.Buffer
+
+	cmdargs := []string{"inspect", d.Name, "-f",
+		fmt.Sprintf("{{(index .NetworkSettings.Networks \"%s\").IPAddress}}", networkName),
+	}
+	cmd := exec.Command("docker", cmdargs...)
+	cmd.Stdout = &stdout
+	err := cmd.Run()
+	Expect(err).NotTo(HaveOccurred())
+
+	// output has trailing "\n", so it needs to be trimed
+	return strings.TrimSuffix(stdout.String(), "\n")
+}
+
+func (d *Docker) GetLog() (string, string) {
+	var stdout, stderr bytes.Buffer
+
+	// get stdout and stderr of `docker log {d.Name}` command
+	cmd := exec.Command("docker", "logs", d.Name)
+	cmd.Stdout = &stdout
+	cmd.Stderr = &stderr
+
+	err := cmd.Run()
+	Expect(err).NotTo(HaveOccurred())
+
+	return stdout.String(), stderr.String()
+}
+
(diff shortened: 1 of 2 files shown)


test/e2e/framework/docker.go
```diff
@@ -0,0 +1,73 @@
...
+
+	// output has trailing "\n", so it needs to be trimed
+	return strings.TrimSuffix(stdout.String(), "\n")
+}
+
+func (d *Docker) GetLog() string {
+	var stderr bytes.Buffer
+	// get stderr of `docker log {d.rame}` command
+	cmd := exec.Command("docker", "logs", d.Name)
+	cmd.Stderr = &stderr
```
sridhargaddam, line 55: Why only stderr and not stdout?

mangelajo: :+1:

mangelajo (commented):
Looks good, only the sridhar comment.

test/e2e/framework/docker.go
```diff
@@ -0,0 +1,73 @@
...
+	"strings"
+
+	. "github.com/onsi/gomega"
+)
+
+type Docker struct {
+	Name string
+}
+
+func NewFromExistingDocker(name string) *Docker {
```
tpantelis:
```suggestion
func New(dockerName string) *Docker {
```

test/e2e/framework/network_pods.go
```diff
@@ -178,6 +188,61 @@ func (np *NetworkPod) CreateService() *v1.Service {
 	return np.framework.CreateTCPService(np.Config.Cluster, np.Pod.Labels[TestAppLabel], np.Config.Port)
 }
 
+// RunCommand run the specified comand in this NetworkPod
```
tpantelis:
```suggestion
// RunCommand run the specified command in this NetworkPod
```

test/e2e/framework/network_pods.go
```diff
@@ -178,6 +188,61 @@ func (np *NetworkPod) CreateService() *v1.Service {
...
+
+// GetLog returns container log from this NetworkPod
+func (np *NetworkPod) GetLog() string {
+	req := KubeClients[np.Config.Cluster].CoreV1().RESTClient().Get().
+		Namespace(np.Pod.Namespace).Name(np.Pod.Name).Resource("pods").SubResource("log")
+
+	req.Param("follow", strconv.FormatBool(false)).
+		Param("container", np.Config.ContainerName).
+		Param("previous", strconv.FormatBool(false)).
+		Param("timestamps", strconv.FormatBool(false))
```
tpantelis: It seems like you could use 

`KubeClients[np.Config.Cluster].CoreV1().Pods(np.Pod.Namespace).GetLogs(np.Pod.Name, &corev1.PodLogOptions{...})`

test/e2e/framework/network_pods.go
```diff
@@ -404,6 +469,38 @@ func (np *NetworkPod) buildLatencyServerPod() {
...
+			},
+			Tolerations: []v1.Toleration{{Operator: v1.TolerationOpExists}},
+		},
+	}
+
+	pc := KubeClients[np.Config.Cluster].CoreV1().Pods(np.framework.Namespace)
+	var err error
+	np.Pod, err = pc.Create(&customPod)
+	Expect(err).NotTo(HaveOccurred())
+	np.AwaitReady()
```
tpantelis: maybe add some newlines for readability...

```suggestion
	pc := KubeClients[np.Config.Cluster].CoreV1().Pods(np.framework.Namespace)
	
	var err error
	np.Pod, err = pc.Create(&customPod)
	Expect(err).NotTo(HaveOccurred())
	
	np.AwaitReady()
```

tpantelis (changes requested):

test/e2e/framework/network_pods.go
mkimuram: Good catch. Actually, I also found that I was specifying default values for `.PodLogOptions` (Even `container` isn't necessary, if there is only one). It now become much simple.

mkimuram:
@mangelajo @sridhargaddam @tpantelis 

Thank you for your reviews. I've fixed and squashed the commit.
I tested locally and confirmed that it works well with the submariner side's PR with "Fix for shipyard framework changes" commit.

PTAL

tpantelis (approved):

mangelajo (approved):

mangelajo:
Great work thanks @mkimuram

sridhargaddam (approved):
Great work @mkimuram

PR #502, opened 2021-03-25
Allow e2e action to call other make target than e2e

### LINKED ISSUE #501: Allow e2e action to call other make target than e2e
Current [e2e action](https://github.com/submariner-io/shipyard/blob/devel/gh-actions/e2e/action.yaml) only allows calling `make e2e`. Allowing other make targets to run other tests will be helpful. For example, in [this use case](https://github.com/submariner-io/submariner/pull/1202#discussion_r601524012).

mkimuram:
This PR allows e2e action to call other make target than e2e.

Fixes: #501


mangelajo (approved):
Thank you @mkimuram

This is what we talked about with @mkolesnik to improve the work in https://github.com/submariner-io/submariner/pull/1202/files

dfarrell07 (commented):
I understand the idea here, but this seems to be a bit of a stretch of this gh-action. :sweat_smile: 

Maybe @mkolesnik has some design idea for this + https://github.com/submariner-io/submariner/pull/1202?

dfarrell07 (approved):
We can iterate on the design, maybe to run external E2E using the same `e2e` target/GHA with different args.

tpantelis (approved):

PR #504, opened 2021-03-26
Allow submariner deploy script to add CIDR args

mkimuram:
This PR allows submariner deploy script to add CIDR args.

Fixes: #503

PR #505, opened 2021-03-29
Pass cluster/service CIDRs to subctl join

dfarrell07:
These CIDRs are always generated in utils/add_cluster_cidrs, along with
the existing global_cidr. They are also already passed by default in
deploy_helm/helm_install_subm.

Alternative implementation of idea in #504.

### DIFF
--- a/scripts/shared/lib/deploy_operator
+++ b/scripts/shared/lib/deploy_operator
@@ -66,6 +66,8 @@ function subctl_install_subm() {
                 --nattport "${CE_IPSEC_NATTPORT}" \
                 --ikeport "${CE_IPSEC_IKEPORT}" \
                 --colorcodes "${SUBM_COLORCODES}" \
+                --clustercidr "${cluster_CIDRs[$cluster]}" \
+                --servicecidr "${service_CIDRs[$cluster]}" \
                 --globalnet-cidr "${global_CIDRs[$cluster]}" \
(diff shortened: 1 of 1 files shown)


dfarrell07:
Here it is working, from the logs of the Shared Scripts/Deploy (operator) job:

> subctl join --kubeconfig /go/src/github.com/submariner-io/shipyard/output/kubeconfigs/kind-config-cluster1 --clusterid cluster1 --repository localhost:5000 --version local --nattport 4500 --ikeport 500 --colorcodes blue --clustercidr 10.1.0.0/16 --servicecidr 100.1.0.0/16 --globalnet-cidr  --disable-nat --cable-driver libreswan /go/src/github.com/submariner-io/shipyard/output/broker-info.subm

skitt (approved):
Given that the CIDR arrays are always supposed to be populated, this seems fine to me.

mangelajo (changes requested):
Is there any reason why we need this always?

We will stop testing auto-detection of that which is done by the operator .

I think it's better to just do it in the necessary cases (like mkimura's patch for external network support).

dfarrell07:
Ah good catch regarding not testing auto-detection @mangelajo, I didn't think about that.

Reopened @mkimuram's https://github.com/submariner-io/shipyard/pull/504 implementation.

PR #513, opened 2021-04-05
Quick fix to ignore any devel versions

mkolesnik:
Downstream rpeositores (notably submariner-operator) use extra "devel"
version tags which mess up version calculations.

For now, ignore any tags matching devel versions, until we have a more
proper solution.

### DIFF
--- a/Makefile.versions
+++ b/Makefile.versions
@@ -1,7 +1,7 @@
 # Calculate versions; these can be overridden
 CUTTING_EDGE := devel
 DEV_VERSION := dev
-override CALCULATED_VERSION := $(shell git describe --tags --dirty="-$(DEV_VERSION)" --exclude="$(CUTTING_EDGE)" --exclude="latest")
+override CALCULATED_VERSION := $(shell git describe --tags --dirty="-$(DEV_VERSION)" --exclude="*$(CUTTING_EDGE)*" --exclude="latest")
 VERSION ?= $(CALCULATED_VERSION)
 
 export CUTTING_EDGE DEV_VERSION VERSION


tpantelis (approved):

Makefile.versions
```diff
@@ -1,7 +1,7 @@
 # Calculate versions; these can be overridden
 CUTTING_EDGE := devel
 DEV_VERSION := dev
-override CALCULATED_VERSION := $(shell git describe --tags --dirty="-$(DEV_VERSION)" --exclude="$(CUTTING_EDGE)" --exclude="latest")
+override CALCULATED_VERSION := $(shell git describe --tags --dirty="-$(DEV_VERSION)" --exclude="*$(CUTTING_EDGE)*" --exclude="latest")
```
tpantelis, line 4: The submariner-operator Makefile does this:

`-X=github.com/submariner-io/submariner-operator/pkg/versions.DefaultSubmarinerOperatorVersion=$(if eq($(patsubst subctl-devel-%,devel,$(CALCULATED_VERSION)),devel),devel,$(CALCULATED_VERSION))"`

so this change would seem to break that (even tho CALCULATED_VERSION evaluates to "" in E2E jobs).

tpantelis (dismissed):

Makefile.versions
mkolesnik: No, in the E2E it evaluates to nothing since there's no history being pulled on the checkout so theres no tags or anything for `git describe` to go by, but on the release job it'll be fine since we clone with all the history.

I made a test PR to show it works, specifically the output youre referring to is:
https://github.com/mkolesnik/shipyard/pull/20/checks?check_run_id=2275653729#step:6:820

mangelajo (approved):

skitt (approved):

Makefile.versions
mangelajo:
```
[majopela@bluehat submariner-operator]$ git describe --tags --dirty="-dev" --exclude="devel" 
subctl-devel
[majopela@bluehat submariner-operator]$ git describe --tags --dirty="-dev" --exclude="*devel*" 
v0.9.0-m1-69-g3f1d238
```

I tried on a fresh checkout of submariner-operator and it works.

PR #514, opened 2021-04-06
Use a simpler version scheme

mkolesnik:
`git describe` is rather finicky and doesn't offer too many benefits for
the cost it incurrs (always have to have tags checked out and updated).

This versioning scheme uses the base branch name (which is available in
all relevant repos to allow for stable branching) and adds the short
commit hash and an optional tag if there are uncommitted changes.


dfarrell07:
Can we also remove the fetch-depth arguments from the release jobs elsewhere?

mkolesnik:
> Can we also remove the fetch-depth arguments from the release jobs elsewhere?

Yes if this is accepted then it's not needed anymore

Makefile.versions
```diff
@@ -1,7 +1,7 @@
 # Calculate versions; these can be overridden
 CUTTING_EDGE := devel
 DEV_VERSION := dev
-override CALCULATED_VERSION := $(shell git describe --tags --dirty="-$(DEV_VERSION)" --exclude="*$(CUTTING_EDGE)*" --exclude="latest")
+override CALCULATED_VERSION := $(BASE_BRANCH)-$(shell git rev-parse --short HEAD)$(shell test -z `git diff-index HEAD` || echo "-$(DEV_VERSION)")
```
mangelajo, line 4: We need to handle tags to, when we are on 0.9.0, 0.9.1, etc... so we can publish $image:0.9.0, $image:0.9.1

Or am I missing something?

mangelajo: Yes I'm missing that we do that now on the releases repository ':D (tagging the final versions...)

dfarrell07 (approved):
Seems like this will be more robust than what we have.

mangelajo (approved):

tpantelis (approved):

PR #530, opened 2021-04-27
Delete listener service once the pod is terminated

sridhargaddam:
The e2e framework transparently creates a service infront of
listenerPod for Globalnet deployments. With k8s 1.20 version
it was seen that when when backend pods finish their execution,
kubeproxy-iptables driver tries to delete the iptables-chain
associated with the service (even when the service is present)
as there are no active backend pods. Since the iptables-chain
is also referenced by Globalnet Ingress rules, kubeproxy fails
to delete the chain and gets into some kind of error state. It
keeps retrying the delete operation for every 30 secs (the default
value with KIND setups). kubeproxy will only be able to succeed
once Globalnet removes the corresponding reference which happens
when the service is deleted.

We have plans to remove this dependency in Globalnet 2.0
https://github.com/submariner-io/submariner/issues/1166

Until its addressed in Globalnet2.0, its better to delete the
service and serviceExport objects in the framework once the
listener pod is terminated. This will fix the flaky issues
seen with Globalnet jobs + K8s 1.20 version in CI.

### DIFF
--- a/test/e2e/tcp/connectivity.go
+++ b/test/e2e/tcp/connectivity.go
@@ -21,6 +21,7 @@ import (
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	"github.com/submariner-io/shipyard/test/e2e/framework"
+	v1 "k8s.io/api/core/v1"
 )
 
 const globalnetGlobalIPAnnotation = "submariner.io/globalIp"
@@ -124,9 +125,10 @@ func createPods(p *ConnectivityTestParams) (*framework.NetworkPod, *framework.Ne
 	})
 
 	remoteIP := listenerPod.Pod.Status.PodIP
+	var service *v1.Service
 	if p.ToEndpointType == ServiceIP || p.ToEndpointType == GlobalIP {
 		By(fmt.Sprintf("Pointing a service ClusterIP to the listener pod in cluster %q", framework.TestContext.ClusterIDs[p.ToCluster]))
-		service := listenerPod.CreateService()
+		service = listenerPod.CreateService()
 		remoteIP = service.Spec.ClusterIP
 
 		if p.ToEndpointType == GlobalIP {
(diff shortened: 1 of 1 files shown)


aswinsuryan (approved):

test/e2e/tcp/connectivity.go
```diff
@@ -167,5 +169,17 @@ func createPods(p *ConnectivityTestParams) (*framework.NetworkPod, *framework.Ne
...
 	framework.Logf("Connector pod has IP: %s", connectorPod.Pod.Status.PodIP)
 
+	// In Globalnet deployments, when backend pods finish their execution, kubeproxy-iptables driver tries
+	// to delete the iptables-chain associated with the service (even when the service is present) as there are
+	// no active backend pods. Since the iptables-chain is also referenced by Globalnet Ingress rules, the chain
+	// cannot be deleted (kubeproxy errors out and continues to retry) until Globalnet removes the reference.
+	// Globalnet removes the reference only when the service itself is deleted. Until Globalnet is enhanced [*]
+	// to remove this dependency with iptables-chain, lets delete the service after the listener Pod is terminated.
+	// [*] https://github.com/submariner-io/submariner/issues/1166
+	if p.ToEndpointType == GlobalIP {
```
mangelajo, line 179: Shouldn't we just go ahead and delete them for ServiceIP ones too?

mangelajo (changes requested):
Just one comment. Thanks for addressing this @sridhargaddam !! :wave:

mangelajo (approved):
Moving to approve since we're only trying to fix the issue with Globalnet, and for other cases the namespace cleanup will do it.

test/e2e/tcp/connectivity.go
sridhargaddam: We don't have any issue with Vanilla Submariner (where ServiceIP) are used. Only globalnet e2e tests use endpointType as tcp.GlobalIP.

As you know in Vanilla Submariner since we don't program anything based on the iptable-chains, its not a problem for Vanilla Submariner use-cases.

skitt (approved):

PR #531, opened 2021-04-27
Add support for target comment to `make targets`

mkolesnik:
To make it easier to understand what each target does, `make targets`
will seek for a comment and if it finds one it'll print it.


Makefile.dapper
```diff
@@ -28,6 +28,10 @@ shell: .dapper
 # This only lists targets accessible inside dapper (which are 99% of targets we use)
 targets:
 	@$(RUN_IN_DAPPER) make -pRrq : 2>/dev/null |\
-		grep -oP '^(?!Makefile.*)[-[:alnum:]]*(?=:)' | sort -u
+		grep -oP '^(?!Makefile.*)[-[:alnum:]]*(?=:)' | sort -u > /tmp/targets.txt;\
```
skitt: Any particular reason to go through a temporary file here?

Makefile.dapper
```diff
@@ -28,6 +28,10 @@ shell: .dapper
...
+		for target in $$(cat /tmp/targets.txt); do\
+			printf "$${target}\t\t";\
```
skitt: To align the descriptions consistently, this works better:

```suggestion
			printf "%-24s" "$${target}";\
```

skitt (changes requested):
Very nice!

Makefile.dapper
mkolesnik: It was simpler than figuring out how to do it in array inside the dapper call :)

If it bothers you we can change it.

Makefile.dapper
mkolesnik: I knew there must be a better way but I'm not that familiar with all the Linux command :)

Makefile.dapper
mkolesnik: Best I got it to output so far
```bash
$ make targets
sha256:f5c634c28eca812d104eebb8fd2f131be18e21dd1264e82e07dd07694dbb7977
[shipyard]$ trap chown -R 15900:15900 . exit
[shipyard]$ mkdir -p bin dist output
[shipyard]$ eval ${SCRIPTS_DIR}/targets.sh
[shipyard]$ /go/src/github.com/submariner-io/shipyard/scripts/shared/targets.sh
CODEOWNERS              
clean                    cleans everything (running clusters, generated files ...)
clean-clusters           removes running clusters
clean-generated          removes files we generated
cleanup                 
clusters                 creates KIND clusters that can then be used to deploy Submariner
deploy                   deploys Submariner on KIND clusters
deploy-latest            deploys the latest release
e2e                      executes the project's end to end testing on the deployed KIND clusters
gitlint                  validates the commits are valid
golangci-lint            validates Go code in the project
images                   builds all the container images for the project
lint                    
markdownlint             validates Markdown files in the project
nettest                 
post-mortem              prints a heap on information, to help in debugging failures on the KIND clusters
preload-images          
print-version           
prune-images             removes all Submariner-provided images and all untagged images
release                  uploads the project's images to a public repository
shellcheck              
unit                     executes the Go unit tests of the project
upgrade-e2e             
yamllint                 validates YAML files in the project
[shipyard]$ eval ${SCRIPTS_DIR}/targets.sh
```

Makefile.dapper
skitt: This is C really :stuck_out_tongue: https://pubs.opengroup.org/onlinepubs/9699919799/functions/snprintf.html

Makefile.dapper
skitt: Oh, I wasn’t thinking as far as that — just

```suggestion
		for target in $((grep -oP '^(?!Makefile.*)[-[:alnum:]]*(?=:)' | sort -u); do\
```

skitt: The pipe isn’t obvious, `make` runs, its output is piped to the `for` and consumed by the `grep` in the command substitution.

mkolesnik: Yes, one does not simply pipe output from make to a for and expect it to still run in dapper.

mkolesnik: Has to be a script run inside dapper, otherwise some of it runs inside and some outside (unless we want multiple dapper calls which would be not very efficient).

skitt: Right, because we need to search for stuff in the Makefiles from the base image too.

scripts/shared/targets.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+
+function print_indent() {
+    printf "%-24s" "$1"
+    printf "$2"
```
skitt:
```suggestion
    printf "%-24s%s\n" "$1" "$2"
```

scripts/shared/targets.sh
```diff
@@ -0,0 +1,19 @@
...
+}
+
+title=$(print_indent "Target" "Description")
+echo "$title"
+echo "$title" | tr '[:alnum:]' '-'
```
skitt: Combined with the above:

```suggestion
print_indent Target Description | tee >(tr '[:alnum:]' '-')
```

scripts/shared/targets.sh
```diff
@@ -0,0 +1,19 @@
...
+
+make_targets=($(make -pRrq : 2>/dev/null |\
+    grep -oP '^(?!Makefile.*)[-[:alnum:]]*(?=:)' | sort -u))
+
+for target in ${make_targets[*]}; do
+    print_indent "${target}"
+    (grep -hoP "(?<=\[${target}\] ).*" Makefile* ${SHIPYARD_DIR}/Makefile* || printf '\n') | head -1
```
skitt:
```suggestion
    print_indent "${target}" \
    "$(grep -hoP -m1 "(?<=\[${target}\] ).*" Makefile* ${SHIPYARD_DIR}/Makefile*)"
```

skitt (changes requested):
Sorry, some more code golfing ;-).

mkolesnik:
> Sorry, some more code golfing ;-).

Thanks, appreciate it :)

skitt (approved):

dfarrell07 (approved):

PR #534, opened 2021-04-28
Only remove dotnet on cleanup

mkolesnik:
Seems that the other directories are full of small files which take up
to 2 minutes to remove, while the dotnet directory takes up ~10x space
of them cobmined, and deleted within a few seconds.


mangelajo (approved):
big plus one to big timesavers!

nyechiel (approved):

PR #536, opened 2021-04-29
Add the base branch as a container build argument

skitt:
This allows container images to reference the base branch, e.g. to
pull in appropriate images they depend on.


mkolesnik (approved):
Seems legit though how was it working before this change without problems?

skitt:
> Seems legit though how was it working before this change without problems?

For the Dapper image, Dapper itself adds environment variables mentioned in `Dockerfile.dapper` as build arguments. Now I’d like to use `BASE_BRANCH` as an `ARG` in the descriptors of the container images we build, so we need to add the argument explicitly.

tpantelis (approved):

dfarrell07 (approved):

PR #539, opened 2021-05-03
Bump Go dependencies to Kubernetes 1.19

skitt:
This is a breaking change for downstream projects since the Kubernetes
client APIs now require contexts. The Shipyard framework APIs
themselves don't change as a result of this patch, but the dependency
bump will propagate and require updates.


dfarrell07 (approved):
ACK that merging will have to be coordinated with other PRs, but +1 and thank you @skitt!

nyechiel (approved):
@skitt this have a conflict now

tpantelis (approved):

tpantelis (approved):

PR #543, opened 2021-05-04
Increase connect-timeout in benchmark tests

sridhargaddam:
Currently, while using subctl benchmark tests, the client pod is
using a connect-timeout of 10ms which is too low. It was working
fine in KIND Clusters as the latency between the clusters is generally
less, but in an OnPrem vs Public cluster environments the benchmark
tests it was creating issues. This PR fixes the issue by increasing
the connect-timeout value to 10 secs.

(cherry picked from commit 3a6730cb36e99ee633f3ce01a3acc99e9906b17d)


maayanf24 (approved):

mangelajo (approved):
:man_dancing:  Thanks @sridhargaddam ! you're the best

nyechiel (approved):

Issue #540, opened 2021-05-03
Break out linting commands to a separate image


mkolesnik:
Our linting CI jobs are rather self contained and as such would benefit from being based on a separate image.

This will allow much faster feedback than we currently have (50% less time) and would also be "greener" as we'll be pulling smaller images.
We could also remove these tools from the base image (once unused) which should reduce it's size (albeit the reduction is unknown at this point and could prove rather negligible).

An added benefit is allowing these commands to be embedded in git commit hooks and run on each commit (or perhaps push) since the image, even if needs to be pulled, is very small.

A [POC](https://github.com/mkolesnik/shipyard/pull/23) proves that an alpine based image results in ~80M size instead of the ~870M the current base image is:
![Screenshot from 2021-05-03 14-46-51](https://user-images.githubusercontent.com/51078027/116872632-e53e6500-ac1e-11eb-94f8-229d919fe868.png)

Compared with execution with the current image:
![Screenshot from 2021-05-03 14-47-00](https://user-images.githubusercontent.com/51078027/116872663-f4bdae00-ac1e-11eb-8d51-ad3677a33249.png)

While this might not seem a great saving (in absolute terms it's ~15s per job), in the long run the pros could outweigh the cons.

The work estimation is rather small as the POC already proves such a feat is quite obtainable and as such shouldn't take to long to implement, should we chose to do so.

mkolesnik:
Resolved by https://github.com/submariner-io/shipyard/pull/564

PR #552, opened 2021-05-13
Linting in a small image

mkolesnik:
Most linting commands (except golangci-lint) are self contained programs
with a relatively small footprint.
This change breaks them out to a separate Dockerfile which is based on
alpine and results in a tiny image (~80Mb on the filesystem) which
should make linting happen much faster (especially on the CI that has to
download the linting image).

If all consuming projects adopt this change, we can remove the linting
commands from the base image which should help reduce its size as well.

Related to: #540 

### DIFF
--- a/Dockerfile.linting
+++ b/Dockerfile.linting
@@ -1,0 +1,10 @@
+FROM quay.io/submariner/shipyard-linting:devel
+
+ENV DAPPER_ENV="GITHUB_SHA MAKEFLAGS" \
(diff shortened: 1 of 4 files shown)


mkolesnik:
Runs of linting on this patch are expected to be slower as the image isn't in Quay.io yet

dfarrell07 (approved):

tpantelis (approved):

skitt (approved):

mangelajo (approved):

mangelajo:
:tada:

mangelajo:
some jobs are queued but not running.

PR #564, opened 2021-05-24
Linting in a small image

### LINKED ISSUE #540: Break out linting commands to a separate image

mkolesnik:
Most linting commands (except golangci-lint) are self contained programs
with a relatively small footprint.
This change breaks them out to a separate Dockerfile which is based on
alpine and results in a tiny image (~80Mb on the filesystem) which
should make linting happen much faster (especially on the CI that has to
download the linting image).

If all consuming projects adopt this change, we can remove the linting
commands from the base image which should help reduce its size as well.

### DIFF
--- a/package/Dockerfile.shipyard-linting
+++ b/package/Dockerfile.shipyard-linting
@@ -1,0 +1,37 @@
+FROM alpine
+
+ENV DAPPER_HOST_ARCH=amd64 SHELL=/bin/bash \
+    SHIPYARD_DIR=/opt/shipyard
+ENV HOST_ARCH=${DAPPER_HOST_ARCH} ARCH=${DAPPER_HOST_ARCH} \
+    SCRIPTS_DIR=${SHIPYARD_DIR}/scripts
+
+# Requirements:
+# Component        | Usage
+# -------------------------------------------------------------------
+# bash             | Just your basic shell
+# findutils        | Finding executables to compress
+# gitlint          | Git commit message linting
+# grep             | For listing targets
+# make             | Running makefiles inside the container
+# markdownlint     | Markdown linting
+# nodejs           | Used by markdownlint
+# npm              | Installing markdownlint (Removed afterwards)
+# py3-pip          | Installing gitlint (Removed afterwards)
+# shellcheck       | Shell script linting
+# upx              | Compressing executables to get a smaller image
+# yamllint         | YAML linting
+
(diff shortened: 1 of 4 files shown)


mangelajo (approved):

package/Dockerfile.shipyard-linting
```diff
@@ -0,0 +1,37 @@
...
+# npm              | Installing markdownlint (Removed afterwards)
+# py3-pip          | Installing gitlint (Removed afterwards)
+# shellcheck       | Shell script linting
+# upx              | Compressing executables to get a smaller image
+# yamllint         | YAML linting
+
+RUN apk add --no-cache bash findutils git grep make nodejs shellcheck upx yamllint && \
+    apk add --no-cache --virtual installers npm py3-pip && \
+    npm install -g markdownlint-cli && \
+    pip install gitlint && \
```
dfarrell07, line 27: I think we're supposed to remove npm and pip after these installs?

https://github.com/submariner-io/shipyard/blob/devel/package/Dockerfile.shipyard-dapper-base#L45

dfarrell07 (dismissed):

package/Dockerfile.shipyard-linting
dfarrell07: Or I guess this is handled by `apk del installers`?

dfarrell07 (approved):

skitt (approved):

tpantelis (approved):

PR #568, opened 2021-05-26
clusters.sh: support short K8s versions

### LINKED ISSUE #567: clusters.sh should accept partial k8s_version arguments
To avoid issues when bumping kind, `clusters.sh` should take partial `--k8s_version` arguments (_e.g._ 1.18, 1.19, 1.20) and fill in the missing patch version based on the versions supported by kind.

skitt:
Instead of specifying the full K8s version all the time, support only
specifying the major and minor; clusters.sh then fills in the patch
version with whatever is supported by the current kind version.

Once downstreams are updated to use this, we will be able to bump kind
versions without breaking them.

Fixes: #567

### DIFF
--- a/scripts/shared/clusters.sh
+++ b/scripts/shared/clusters.sh
@@ -1,9 +1,17 @@
 #!/usr/bin/env bash
 
+## Kubernetes version mapping, as supported by kind ##
+# See the release notes of the kind version in use
+declare -A kind_k8s_versions
+kind_k8s_versions[1.17]=1.17.17
+kind_k8s_versions[1.18]=1.18.15
+kind_k8s_versions[1.19]=1.19.7
+kind_k8s_versions[1.20]=1.20.2
+
 ## Process command line flags ##
 
 source ${SCRIPTS_DIR}/lib/shflags
-DEFINE_string 'k8s_version' '1.20.2' 'Version of K8s to use'
+DEFINE_string 'k8s_version' '1.20' 'Version of K8s to use'
 DEFINE_string 'olm_version' '0.14.1' 'Version of OLM to use'
 DEFINE_boolean 'olm' false 'Deploy OLM'
 DEFINE_boolean 'prometheus' false 'Deploy Prometheus'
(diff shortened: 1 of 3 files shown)


scripts/shared/clusters.sh
```diff
@@ -1,9 +1,20 @@
 #!/usr/bin/env bash
 
+## Kubernetes version mapping, as supported by kind ##
+# See the release notes of the kind version in use
+declare -A kind_k8s_versions
+kind_k8s_versions[1.14]=1.14.10
+kind_k8s_versions[1.15]=1.15.12
+kind_k8s_versions[1.16]=1.16.15
```
mkolesnik:
```suggestion
```

Do we really need these old ones? We're not even supporting them..

scripts/shared/clusters.sh
```diff
@@ -16,6 +27,7 @@ eval set -- "${FLAGS_ARGV}"
 
 k8s_version="${FLAGS_k8s_version}"
 olm_version="${FLAGS_olm_version}"
+[[ -n "${kind_k8s_versions[$k8s_version]}" ]] && k8s_version="${kind_k8s_versions[$k8s_version]}"
```
mkolesnik, line 27:
```suggestion
[[ -z "${kind_k8s_versions[$k8s_version]}" ]] || k8s_version="${kind_k8s_versions[$k8s_version]}"
```

I think otherwise it breaks (bash is weird sometimes)

mkolesnik (changes requested):

scripts/shared/clusters.sh
skitt: Do you know under what circumstances?

skitt: If you’re referring to `set -e`, that doesn’t matter here — the `[[ -n ... ]]` test does produce a non-zero exit code, and then the overall pipeline has a non-zero exit code, but the `&&` causes it to be considered as consumed, and the shell script _doesn’t_ exit as a result.

dfarrell07 (approved):
Great idea @skitt, thank you!

tpantelis (approved):

scripts/shared/clusters.sh
```diff
@@ -1,9 +1,17 @@
...
+kind_k8s_versions[1.17]=1.17.17
+kind_k8s_versions[1.18]=1.18.15
+kind_k8s_versions[1.19]=1.19.7
+kind_k8s_versions[1.20]=1.20.2
+
 ## Process command line flags ##
 
 source ${SCRIPTS_DIR}/lib/shflags
-DEFINE_string 'k8s_version' '1.20.2' 'Version of K8s to use'
+DEFINE_string 'k8s_version' '1.20' 'Version of K8s to use'
```
mangelajo, line 14: Side comment to this patch, the change looks good:

Oh, I see this is 1.20 by default. I wonder if in the consuming projects we should leave the 1.20 version empty. At least that gives us the indication of which version was it tested with so may be it's better to keep it

mangelajo (approved):

scripts/shared/clusters.sh
mkolesnik: IDK had problems with that before but if it works fine with me :)

mkolesnik (approved):

PR #573, opened 2021-05-31
Shared Dockerfile.dapper

mkolesnik:
In light of Shipyard supplying shared infrastructure, it should be easy
to shared Dockerfile.dapper as it's mostly a copy pasted file.

In the case it already exists it shoudn't be re-downloaded by consuming
projects, and they can remove it gradually.


Makefile
```diff
@@ -1,10 +1,14 @@
 BASE_BRANCH ?= devel
+PROJECT := shipyard
```
skitt: This could be calculated as `$(notdir $(CURDIR))`, and declared in `Makefile.dapper` instead, so we don’t need to manually specify it everywhere.

skitt (changes requested):

skitt:
Yes, let’s get rid of copy-pasta!

Makefile
mkolesnik: Thanks :)

I had a feeling there's some automatic way to do this and that you'll suggest it ;)

mangelajo (approved):

skitt (approved):

PR #576, opened 2021-06-02
Copy Shipyard resources into consuming projects

### LINKED ISSUE #491: Changes to Makefile.dapper can’t be tested in consuming projects
When a Shipyard PR changes `Makefile.dapper`, it can be useful to run e2e in consuming projects; but the latter downloads the current `devel` `Makefile.dapper`, not the one to be tested (this was already a problem before projects were changed to download `Makefile.dapper`).

Ideally we should be able to override the downloaded `Makefile.dapper`, if only by copying the new version in — since it’s handled by a `Makefile` target, it shouldn’t be downloaded if it’s already present.

### LINKED ISSUE #575: When sharing Dockerfile.dapper, the tag must be overridable
For the consuming tests in Shipyard, we need to be able to [override the tag used for the Shipyard base image](https://github.com/submariner-io/shipyard/blob/devel/.github/workflows/consuming.yml#L57), even when there is no `Dockerfile.dapper` in the project. See https://github.com/submariner-io/shipyard/pull/566/checks?check_run_id=2728806310 for an example of a failure caused by this.

skitt:
For consuming jobs, the projects being checked should be updated to
use the Shipyard resources from the Shipyard PR being tested, not
those from the base branch.

Fixes: #491
Fixes: #575

### DIFF
--- a/.github/workflows/consuming.yml
+++ b/.github/workflows/consuming.yml
@@ -53,6 +53,9 @@ jobs:
           repository: submariner-io/${{ matrix.project }}
           path: ${{ matrix.project }}
 
+      - name: Copy Shipyard resources
+        run: cp -n Dockerfile.* Makefile.dapper ${{ matrix.project }}/
+
       - name: Make sure ${{ matrix.project }} is using the built Shipyard image
         run: sed -i 's/shipyard-dapper-base:*.*/shipyard-dapper-base:dev/' ${{ matrix.project }}/Dockerfile.dapper
 
@@ -89,12 +92,16 @@ jobs:
         uses: actions/checkout@v2
         with:
           repository: submariner-io/${{ matrix.project }}
+          path: ${{ matrix.project }}
+
+      - name: Copy Shipyard resources
+        run: cp -n Dockerfile.* Makefile.dapper ${{ matrix.project }}/
 
       - name: Make sure ${{ matrix.project }} is using the built Shipyard image
-        run: sed -i 's/shipyard-dapper-base:*.*/shipyard-dapper-base:dev/' Dockerfile.dapper
+        run: sed -i 's/shipyard-dapper-base:*.*/shipyard-dapper-base:dev/' ${{ matrix.project }}/Dockerfile.dapper
 
       - name: Run all linters
-        run: make lint
+        run: make -C ${{ matrix.project }} lint
 
   unit-consuming:
     name: Unit Tests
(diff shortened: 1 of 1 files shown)


.github/workflows/consuming.yml
```diff
@@ -53,6 +53,9 @@ jobs:
           repository: submariner-io/${{ matrix.project }}
           path: ${{ matrix.project }}
 
+      - name: Copy Dockerfile.dapper if necessary
+        run: test -f ${{ matrix.project }}/Dockerfile.dapper || cp Dockerfile.dapper ${{ matrix.project }}/
+
       - name: Make sure ${{ matrix.project }} is using the built Shipyard image
         run: sed -i 's/shipyard-dapper-base:*.*/shipyard-dapper-base:dev/' ${{ matrix.project }}/Dockerfile.dapper
```
mkolesnik, line 60: Just a question - why not have it as part of this step?

It seems closely tied, I'd put them both in the same step

dfarrell07 (approved):

.github/workflows/consuming.yml
```diff
@@ -89,12 +92,16 @@ jobs:
...
         with:
           repository: submariner-io/${{ matrix.project }}
+          path: ${{ matrix.project }}
+
+      - name: Copy Dockerfile.dapper if necessary
+        run: test -f ${{ matrix.project }}/Dockerfile.dapper || cp Dockerfile.dapper ${{ matrix.project }}/
 
       - name: Make sure ${{ matrix.project }} is using the built Shipyard image
-        run: sed -i 's/shipyard-dapper-base:*.*/shipyard-dapper-base:dev/' Dockerfile.dapper
+        run: sed -i 's/shipyard-dapper-base:*.*/shipyard-dapper-base:dev/' ${{ matrix.project }}/Dockerfile.dapper
```
mangelajo, line 101: Why not make this conditional instead?
```suggestion
        run: [ ! -f  ${{ matrix.project }}/Dockerfile.dapper ] || sed -i 's/shipyard-dapper-base:*.*/shipyard-dapper-base:dev/' ${{ matrix.project }}/Dockerfile.dapper
```

?

mangelajo (changes requested):

.github/workflows/consuming.yml
skitt: Because we want to push our `Dockerfile.dapper` into the project being tested, not rely on whatever version is downloaded — the latter gets two things wrong: it ends up relying on the `devel` image, not the new Shipyard image, and it uses the `devel` `Dockerfile.dapper`, not the one supposedly being tested.

.github/workflows/consuming.yml
skitt: I was looking at fixing #491 too, with a more generic step; does it make better sense like this?

.github/workflows/consuming.yml
skitt: Put another way, this line should never be conditional because we always want to replace the tag.

.github/workflows/consuming.yml
mkolesnik: Hmm yeah makes sense I guess.

Though what would make even more sense IMHO is to extract these all repetitive steps to it's own composite action and just call that, but it's a bit outside the scope of this specific PR :)

mkolesnik (approved):

.github/workflows/consuming.yml
skitt: Yup, a specific consuming composite action would be nice ;-).

.github/workflows/consuming.yml
mangelajo: Ah, right, got it.

mangelajo (approved):

PR #579, opened 2021-06-09
Store the hash of generated images

skitt:
This will help us check later on that we deployed the images we expect
to have.

### DIFF
--- a/scripts/shared/build_image.sh
+++ b/scripts/shared/build_image.sh
@@ -11,6 +11,7 @@ DEFINE_string 'dockerfile' '' "Dockerfile to build from" 'f'
 DEFINE_string 'buildargs' '' "Build arguments to pass to 'docker build'"
 DEFINE_boolean 'cache' true "Use cached layers from latest image"
 DEFINE_string 'platform' "${default_platform}" 'Platforms to target'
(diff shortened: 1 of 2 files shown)


mangelajo (approved):

scripts/shared/build_image.sh
```diff
@@ -11,6 +11,7 @@ DEFINE_string 'dockerfile' '' "Dockerfile to build from" 'f'
 DEFINE_string 'buildargs' '' "Build arguments to pass to 'docker build'"
 DEFINE_boolean 'cache' true "Use cached layers from latest image"
 DEFINE_string 'platform' "${default_platform}" 'Platforms to target'
+DEFINE_string 'hash' '' "File to write the hash to" 'h'
```
mangelajo, line 14: shall we may be using the same hash file always for all images?

Should make it easier to check on runtime later ? not sure how that works.

skitt: This PR writes each image’s hash to the `package/.image....` file which is used to indicate that the image has been built. That way, when we deploy, we can find out (a) which images have been built in the test run and (b) what their hashes are.

dfarrell07 (approved):

PR #588, opened 2021-06-28
Don't expand command substitution in debug

mkolesnik:
This should make sure command substitution isn't being expanded when
printing the debug messages, causing some command to be executed twice.

### DIFF
--- a/scripts/shared/lib/debug_functions
+++ b/scripts/shared/lib/debug_functions
@@ -9,9 +9,11 @@ NO_COLOR=$(echo -e '\e[0m')
 
 function trap_commands() {
     # Function to print each bash command before it is executed
-    trap '! [[ "$BASH_COMMAND" =~ ^(echo|read|\[|while|for|local) ]] &&
+    trap 'cmd="$BASH_COMMAND" &&
+          ! [[ "$cmd" =~ ^(echo|read|\[|while|for|local) ]] &&
+          { [[ ! "$cmd" =~ \$\( ]] || cmd="${cmd@Q}"; } &&
           ctxt="${cluster:+[${cluster}]}" &&
-          cmd=`eval echo "[${PWD##*/}]\$ $ctxt $BASH_COMMAND" 2>/dev/null` &&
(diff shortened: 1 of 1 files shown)


skitt (changes requested):
I might be missing something, but can’t we handle this by protecting `$BASH_COMMAND` directly with `@Q` in all cases in the existing code?

```
    trap '! [[ "$BASH_COMMAND" =~ ^(echo|read|\[|while|for|local) ]] &&
          ctxt="${cluster:+[${cluster}]}" &&
          cmd=`eval echo "[${PWD##*/}]\$ $ctxt ${BASH_COMMAND@Q}" 2>/dev/null` &&
          echo "${CYAN_COLOR}$cmd${NO_COLOR}" >&2; true' DEBUG
```

mkolesnik:
> I might be missing something, but can’t we handle this by protecting `$BASH_COMMAND` directly with `@Q` in all cases in the existing code?
> 
> ```
>     trap '! [[ "$BASH_COMMAND" =~ ^(echo|read|\[|while|for|local) ]] &&
>           ctxt="${cluster:+[${cluster}]}" &&
>           cmd=`eval echo "[${PWD##*/}]\$ $ctxt ${BASH_COMMAND@Q}" 2>/dev/null` &&
>           echo "${CYAN_COLOR}$cmd${NO_COLOR}" >&2; true' DEBUG
> ```

I tried that but then everything is escaped and debugging is much harder :/

This at least presereves most of the vars but avoids the "double running" problem

mangelajo (approved):

skitt:
> I tried that but then everything is escaped and debugging is much harder :/

Right, the reader needs to track all the assignments manually — it’s useful in some cases, but not so much in many others, and changes the behaviour compared to what we have currently.

> This at least presereves most of the vars but avoids the "double running" problem

+1

skitt (approved):

tpantelis (approved):

mangelajo:
I disabled the helm / globalnet combination while it's broken. (helm missing some RBAC)

mangelajo:
Backport in https://github.com/submariner-io/shipyard/pull/589

PR #592, opened 2021-07-01
Add a periodic check for updates

skitt:
This catches more pending updates than dependabot, in particular it
identifies updates in projects without releases.

See https://github.com/skitt/shipyard/pull/1/checks?check_run_id=2962467985 for an example run.


dfarrell07:
> See https://github.com/skitt/shipyard/pull/1/checks?check_run_id=2962467985 for an example run.

Maybe I'm missing something, but I don't see a sample run there and I don't know how there could be one since the job is being added to the periodic workflow. You could modify the periodic workflow to also run on PRs if you want to show it running there.

skitt:
> > See https://github.com/skitt/shipyard/pull/1/checks?check_run_id=2962467985 for an example run.
> 
> Maybe I'm missing something, but I don't see a sample run there and I don't know how there could be one since the job is being added to the periodic workflow. You could modify the periodic workflow to also run on PRs if you want to show it running there.

That’s what I’d done in my initial submission, to see the job run; the link above was the result. Unfortunately it seems old runs aren’t accessible, or am I missing something?

skitt:
Checks will appear here: https://github.com/skitt/shipyard/pull/2/checks

dfarrell07:
> Checks will appear here: https://github.com/skitt/shipyard/pull/2/checks

Oh wow, such nice output! 

Have you thought about somehow making this fail if there are pending updates (maybe that aren't whitelisted as okay to be behind, like the K8s API)? It's still very cool/useful as it is, just wondering what you're thinking.

dfarrell07 (approved):

skitt:
> > Checks will appear here: https://github.com/skitt/shipyard/pull/2/checks
> 
> Oh wow, such nice output!
> 
> Have you thought about somehow making this fail if there are pending updates (maybe that aren't whitelisted as okay to be behind, like the K8s API)? It's still very cool/useful as it is, just wondering what you're thinking.

Yeah, I’d thought about that but since we can’t easily ignore certain dependencies, I didn’t want to pursue that right away...

skitt:
This PR is really more about remembering how to do this, with some nice tooling, and having a central place to look for pending updates every now and then (as opposed to me running this locally).

dfarrell07:
Yep, makes sense, that's how I'm using it now (learning/reproducing the magic search you've been doing locally). :+1:

mkolesnik (commented):
This doesn't seem to have any ties to Shipyard, what's the significance of checking all the other projects specifically here?

skitt:
> This doesn't seem to have any ties to Shipyard, what's the significance of checking all the other projects specifically here?

It means we run the checks in a single job in a single project, instead of all the Go projects (including Shipyard). I can duplicate the PR across all projects if that makes more sense.

dfarrell07 (approved):

mkolesnik (approved):

dfarrell07:
Is this ready to merge?

skitt:
> Is this ready to merge?

As far as I’m concerned, it has been for a while.

dfarrell07:
The `Consuming Projects / E2E` job seems to be marked as required, but it shouldn't be I think.

dfarrell07:
All the required jobs are passing, but it's not merging.

skitt:
> All the required jobs are passing, but it's not merging.

According to the block message, this needs one more approving review (from a code owner, I imagine, since there are already two positive reviews).

dfarrell07:
> > All the required jobs are passing, but it's not merging.
> 
> According to the block message, this needs one more approving review (from a code owner, I imagine, since there are already two positive reviews).

I saw that, but GH shows Mike with a "code owner lock" by his name for me. I see now though he doesn't have rights to all the files touched by this PR, like tools.go. I guess that's the blocking UI-driven confusion, not the `Consuming Projects / E2E` being marked as required/skipped until I ran it and none of the individual jobs are marked as required.

tpantelis (approved):

PR #595, opened 2021-07-06
Add E2E framework functions to create Service w/o selector and Endpoints

mkimuram:
This PR is to add E2E framework functions to create Service w/o selector and Endpoints.
These functions are needed by https://github.com/submariner-io/submariner/pull/1455 and maybe be reused by other E2E tests.

### DIFF
--- a/test/e2e/framework/endpoints.go
+++ b/test/e2e/framework/endpoints.go
@@ -1,0 +1,78 @@
+/*
+SPDX-License-Identifier: Apache-2.0
+
+Copyright Contributors to the Submariner project.
+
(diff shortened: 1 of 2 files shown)


tpantelis (approved):

skitt (approved):

mkimuram:
As a result of testing with these functions, I found that `name` field in `Endpoints`'s port needs to match with `name` field in `Service`'s port. So, these functions are modified to take `portName` as an argument.

sridhargaddam (approved):

PR #596, opened 2021-07-07
Remove extraneous E2E args

### LINKED ISSUE #527: Remove script specific Makefile variables
There were some script specific Makefile variables added recently:

https://github.com/submariner-io/shipyard/blob/devel/Makefile.inc#L54-L69

We have, long ago, decided against having script specific make variables. In this regard, this change is a regression.
In addition, having these script options redefined as make variables, whereas other options aren't, is very inconsistent.

Instead of having and using these script specific variables, use the general `*_ARGS` variables i.e. `E2E_ARGS`

mkolesnik:
These shouldn't be `Makefile` args or GHA inputs, we already have a mechanism to specify such args with `using` and the various `*_E2E` args.

Resolves: #527 
Depends on: https://github.com/submariner-io/submariner/pull/1458


dfarrell07 (approved):

dfarrell07:
@manosnoam You're not using these Ginkgo options through this tooling, right?

dfarrell07:
@mkimuram Are you using these args?

mkimuram:
`plugin` and `testdir` are used [here](https://github.com/submariner-io/submariner/blob/devel/.github/workflows/e2e-external.yml#L18) in GHA. `focus` and `skip` are expected to be used manually on running `make e2e` from command line for local test/debug.

dfarrell07:
Thanks for the info @mkimuram! 

> `plugin` and `testdir` are used [here](https://github.com/submariner-io/submariner/blob/devel/.github/workflows/e2e-external.yml#L18) in GHA.

@mkolesnik has a PR to migrate these to `using=`: https://github.com/submariner-io/submariner/pull/1458

> `focus` and `skip` are expected to be used manually on running `make e2e` from command line for local test/debug.

Sounds like we should do your suggestion of adding them "as proper 'ginkgo args' should such a
feature be desired" @mkolesnik?

mkolesnik:
> `plugin` and `testdir` are used [here](https://github.com/submariner-io/submariner/blob/devel/.github/workflows/e2e-external.yml#L18) in GHA. `focus` and `skip` are expected to be used manually on running `make e2e` from command line for local test/debug.

Expected is one thing, question is if anyone is actually using them.

I prefer to add things that are actively used and not maintain them just for sake of "maybe they'll be used".

mkimuram:
> Expected is one thing, question is if anyone is actually using them.

Personally, I actually use them. It is helpful in testing only related tests or added tests, locally.

However, I will wait for reply from other people on whether using or planning to use it. Note that at least `focus` was in [submarier-io/submariner](https://github.com/submariner-io/submariner/commit/1b2511d66d511172266e2e3a8260a89e3ad37393) before my commit.

mkimuram:
FYI: ginkgo's skip and focus are effectively used in k8s's CI, by adding `[slow]` or `[disruptive]` tags in test specs and making them in separate CI jobs. These jobs run only on related PR or periodically, not on all PRs. It's not currently utilized in such a way in submariner's project, however it will be worth considering.

PR #599, opened 2021-07-11
test(deploy): add the bundle as a new deployment tool

SteveMattar:
Depends on https://github.com/submariner-io/submariner-operator/pull/1568
after adding commit 3add4f1af58fde8df69937dae7b59a74204259d5

### DIFF
--- a/scripts/shared/clusters.sh
+++ b/scripts/shared/clusters.sh
@@ -14,7 +14,7 @@ kind_k8s_versions[1.21]=1.21.1
 
 source ${SCRIPTS_DIR}/lib/shflags
 DEFINE_string 'k8s_version' "${DEFAULT_K8S_VERSION}" 'Version of K8s to use'
-DEFINE_string 'olm_version' '0.14.1' 'Version of OLM to use'
+DEFINE_string 'olm_version' '0.16.1' 'Version of OLM to use'
 DEFINE_boolean 'olm' false 'Deploy OLM'
 DEFINE_boolean 'prometheus' false 'Deploy Prometheus'
 DEFINE_boolean 'globalnet' false "Deploy with operlapping CIDRs (set to 'true' to enable)"
@@ -34,7 +34,7 @@ olm_version="${FLAGS_olm_version}"
 [[ "${FLAGS_registry_inmemory}" = "${FLAGS_TRUE}" ]] && registry_inmemory=true || registry_inmemory=false
 cluster_settings="${FLAGS_cluster_settings}"
 timeout="${FLAGS_timeout}"
-echo "Running with: k8s_version=${k8s_version}, olm_version=${olm_version}, olm=${olm}, globalnet=${globalnet}, registry_inmemory=${registry_inmemory}, cluster_settings=${cluster_settings}, timeout=${timeout}"
+echo "Running with: k8s_version=${k8s_version}, olm_version=${olm_version}, olm=${olm}, globalnet=${globalnet}, prometheus=${prometheus}, registry_inmemory=${registry_inmemory}, cluster_settings=${cluster_settings}, timeout=${timeout}"
 
 set -em
 
--- a/scripts/shared/clusters.sh
+++ b/scripts/shared/clusters.sh
@@ -56,10 +56,6 @@ EOM
 
 ### Functions ###
 
-function render_template() {
-    eval "echo \"$(cat $1)\""
-}
-
 function generate_cluster_yaml() {
     local pod_cidr="${cluster_CIDRs[${cluster}]}"
     local service_cidr="${service_CIDRs[${cluster}]}"
--- a/scripts/shared/lib/deploy_bundle
+++ b/scripts/shared/lib/deploy_bundle
@@ -1,0 +1,165 @@
+# shellcheck shell=bash
+# shellcheck source=scripts/shared/lib/source_only
+. "${BASH_SOURCE%/*}"/source_only
+
+### Constants ###
+
+# These variables are used elsewhere
+# shellcheck disable=SC2034
+readonly MARKETPLACE_NAMESPACE="olm"
+# shellcheck disable=SC2034
+readonly OPERATOR_NAME="submariner"
+# shellcheck disable=SC2034
+readonly INDEX_IMG=localhost:5000/submariner-operator-index:local
+
+# Variables
+
+BROKER_K8S_API_SERVER=""
+BROKER_K8S_API_SERVER_TOKEN=""
+BROKER_K8S_CA=""
+
+### Functions ###
+
+function deploytool_prereqs() {
+  [[ -z "${SUBM_NS}" ]] && (echo "[ERROR] Required environment variables SUBM_NS not loaded"; exit 3)
+
+  # Create new namespace
+  run_subm_clusters create_submariner_namespace
+
+  # Create the custom catalog source
+  run_subm_clusters create_catalog_source
+
+  # Install the submariner operator bundle
+  run_subm_clusters install_bundle
+}
+
+function create_submariner_namespace {
+  echo "[INFO](${cluster}) Create the namespace..."
+  kubectl create namespace "${SUBM_NS}" || :
+}
+
+function create_catalog_source() {
+  echo "[INFO](${cluster}) Create the catalog source..."
+
+  kubectl delete catalogsource/operatorhubio-catalog -n "${MARKETPLACE_NAMESPACE}" --wait --ignore-not-found
+  kubectl delete catalogsource/submariner-catalog-source -n "${MARKETPLACE_NAMESPACE}" --wait --ignore-not-found
+
+  # Create the CatalogSource
+  render_template "${RESOURCES_DIR}"/bundle/catalogSource.yaml | kubectl apply -f -
+
+  # Wait for the catalogSource Readiness
+  if ! (timeout 5m bash -c "until [[ $(kubectl --context="${cluster}" get catalogsource -n ${MARKETPLACE_NAMESPACE} submariner-catalog-source -o jsonpath='{.status.connectionState.lastObservedState}') -eq 'READY' ]]; do sleep 10; done"); then
+      echo "[ERROR] CatalogSource is not ready."
+      exit 1
+  fi
+
+  # Debug
+  kubectl -n ${MARKETPLACE_NAMESPACE} get catalogsource --ignore-not-found
+  kubectl -n ${MARKETPLACE_NAMESPACE} get pods --ignore-not-found
+  kubectl -n ${MARKETPLACE_NAMESPACE} get packagemanifests --ignore-not-found | grep 'Submariner Catalog Source' || true
+
+
+  # Create the OperatorGroup
+  render_template "${RESOURCES_DIR}"/bundle/operatorGroup.yaml | kubectl apply -f -
+
+  # Debug
+  kubectl get og -n "${SUBM_NS}" --ignore-not-found
+
+  echo "[INFO](${cluster}) Catalog source created"
+}
+
+function install_bundle() {
+  local installPlan
+
+  # Delete previous catalogSource and Subscription
+  kubectl delete sub/submariner-subscription -n "${SUBM_NS}" --wait --ignore-not-found
+
+  # Create the Subscription (Approval should be Manual not Automatic in order to pin the bundle version)
+  echo "[INFO](${cluster}) Install the bundle..."
+  render_template "${RESOURCES_DIR}"/bundle/subscription.yaml | kubectl apply -f -
+
+  # Manual Approve
+  echo "[INFO](${cluster}) Approve the installPlan..."
+  kubectl wait --for condition=InstallPlanPending --timeout=5m -n "${SUBM_NS}" subs/submariner-subscription || (echo "[ERROR](${cluster}) InstallPlan not found."; exit 1)
+  installPlan=$(kubectl get subscriptions.operators.coreos.com submariner-subscription -n "${SUBM_NS}" -o jsonpath='{.status.installPlanRef.name}')
+  if [ -n "${installPlan}" ]; then
+    kubectl patch installplan -n "${SUBM_NS}" "${installPlan}" -p '{"spec":{"approved":true}}' --type merge
+  fi
+
+  # Debug
+  kubectl get sub -n "${SUBM_NS}" --ignore-not-found
+  kubectl get installplan -n "${SUBM_NS}" --ignore-not-found
+  kubectl get csv -n "${SUBM_NS}" --ignore-not-found
+  kubectl get pods -n "${SUBM_NS}" --ignore-not-found
+
+  echo "[INFO](${cluster}) Bundle installed"
+}
+
+function setup_broker() {
+  local brokerClientSecret
+
+  if ! (timeout 5m bash -c "until kubectl --context=${cluster} get crds brokers.submariner.io > /dev/null 2>&1; do sleep 10; done"); then
+      echo "[ERROR] Broker CRD was not found."
+      exit 1
+  fi
+
+  # Create the broker Namespace & RBAC
+  echo "[INFO] Deploy the broker"
+  if [ -d "${DAPPER_SOURCE}"/config/broker ]; then
+    kubectl apply -k "${DAPPER_SOURCE}"/config/broker -n "${BROKER_NAMESPACE}"
+  else
+    [ -d /tmp/submariner-operator ] && rm -Rf /tmp/submariner-operator
+    git clone --depth 1 --single-branch --branch "${BASE_BRANCH}" https://github.com/submariner-io/submariner-operator /tmp/submariner-operator
+    kubectl apply -k /tmp/submariner-operator/config/broker -n "${BROKER_NAMESPACE}"
+  fi
+
+  # Enable the service-discovery component if defined
+  local components="connectivity"
+  components+=$([[ ${service_discovery} == "true" ]] && echo ", service-discovery" || echo "")
+  components+=$([[ ${globalnet} == "true" ]] && echo ", globalnet" || echo "")
+
+  ### Create the Broker instance
+  render_template "${RESOURCES_DIR}"/bundle/broker.yaml | kubectl apply -f -
+
+  echo "[INFO] Wait for the broker readiness..."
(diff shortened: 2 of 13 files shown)


dfarrell07 (commented):
This is awesome/exciting @SteveMattar, thank you! If we add "bundle" to the deploy script tests I think we'll start getting interesting test coverage. https://github.com/submariner-io/shipyard/blob/devel/.github/workflows/scripts.yml#L46

scripts/shared/lib/deploy_bundle
```diff
@@ -0,0 +1,249 @@
...
+
+  if ! (timeout 5m bash -c "until kubectl --context=${cluster} get crds brokers.submariner.io > /dev/null 2>&1; do sleep 10; done"); then
+      echo "[ERROR] Broker CRD was not found."
+      exit 1
+  fi
+
+  # Create the broker Namespace & RBAC
+  echo "[INFO] Deploy the broker"
+  [ -d /tmp/submariner-operator ] && rm -Rf /tmp/submariner-operator
+  git clone --depth 1 --single-branch --branch "${BASE_BRANCH}" https://github.com/submariner-io/submariner-operator /tmp/submariner-operator
```
skitt: This makes me somewhat uneasy, but I’m not sure it’s really any different than using the `devel` `subctl` when deploying using the latter...

skitt (approved):

scripts/shared/lib/deploy_bundle
SteveMattar: @skitt yes, there are some differences. `subctl` handles everything around the broker deployment whether it's RBAC or SCC. We would like to make sure we don't miss anything in our guide for users who are not using subctl.

Dockerfile.dapper
```diff
@@ -2,7 +2,7 @@ ARG BASE_BRANCH
 FROM quay.io/submariner/shipyard-dapper-base:${BASE_BRANCH}
 
 ARG PROJECT
-ENV DAPPER_ENV="QUAY_USERNAME QUAY_PASSWORD CLUSTERS_ARGS DEPLOY_ARGS CLEANUP_ARGS E2E_ARGS RELEASE_ARGS MAKEFLAGS FOCUS SKIP PLUGIN E2E_TESTDIR" \
+ENV DAPPER_ENV="QUAY_USERNAME QUAY_PASSWORD CLUSTERS_ARGS DEPLOY_ARGS CLEANUP_ARGS E2E_ARGS RELEASE_ARGS MAKEFLAGS FOCUS SKIP PLUGIN E2E_TESTDIR VERSION CHANNEL" \
```
mkolesnik: Why is there a need to pass the version? It is calculated by the `make` command

scripts/shared/lib/deploy_bundle
```diff
@@ -0,0 +1,249 @@
+# shellcheck shell=bash
+# shellcheck source=scripts/shared/lib/source_only
+. "${BASH_SOURCE%/*}"/source_only
+
+### Constants ###
+
+readonly CE_IPSEC_IKEPORT=500
+readonly CE_IPSEC_NATTPORT=4500
```
mkolesnik: Since these are also used in operator, they could be extracted to e.g. `deploy.sh` itself

scripts/shared/lib/deploy_bundle
```diff
@@ -0,0 +1,249 @@
...
+readonly SUBM_IMAGE_REPO=localhost:5000
```
mkolesnik: This one exists in operator as `SUBM_GATEWAY_IMAGE_REPO` and can be extracted to a common place and used here.

scripts/shared/lib/deploy_bundle
```diff
@@ -0,0 +1,249 @@
...
+readonly SUBM_IMAGE_TAG=${image_tag:-local}
+readonly MARKETPLACE_NAMESPACE="olm"
+readonly OPERATOR_NAME="submariner"
+readonly BROKER_NAMESPACE="submariner-k8s-broker"
+readonly BROKER_CLIENT_SA="submariner-k8s-broker-client"
+readonly INDEX_IMG=localhost:5000/submariner-operator-index:local
+readonly IPSEC_PSK="$(dd if=/dev/urandom count=64 bs=8 | LC_CTYPE=C tr -dc 'a-zA-Z0-9' | fold -w 64 | head -n 1)"
```
mkolesnik: I'd extract this one to a common function or constant as it's also used by helm deploy

scripts/shared/lib/deploy_bundle
```diff
@@ -0,0 +1,249 @@
...
+  echo "VERSION:        ${VERSION}"
+  echo "SUBM_NS:        ${SUBM_NS}"
+  echo "CHANNEL:        ${CHANNEL}"
+  if [[ -z "${VERSION}" ]] ||
+   [[ -z "${SUBM_NS}" ]] ||
+   [[ -z "${CHANNEL}" ]]; then
+    echo "[ERROR] Required environment variables not loaded"
+    echo "    VERSION"
+    echo "    SUBM_NS"
+    echo "    CHANNEL"
```
mkolesnik: I'd add the values here as it's not clear what isn't set, or alternatively print out a message for each missing expected var.

scripts/shared/clusters.sh
```diff
@@ -14,8 +14,8 @@ kind_k8s_versions[1.21]=1.21.1
 
 source ${SCRIPTS_DIR}/lib/shflags
 DEFINE_string 'k8s_version' "${DEFAULT_K8S_VERSION}" 'Version of K8s to use'
-DEFINE_string 'olm_version' '0.14.1' 'Version of OLM to use'
-DEFINE_boolean 'olm' false 'Deploy OLM'
+DEFINE_string 'olm_version' '0.16.1' 'Version of OLM to use'
+DEFINE_boolean 'olm' false 'Deploy OLM, create a catalogSource and install the operator bundle'
```
mkolesnik: Does this make sense? Seems `deploy.sh` creates the `catalogSource` and installs operator bundle, so this here seems unnecessary and misleading

scripts/shared/lib/deploy_bundle
mkolesnik: I'd clone only if it doesn't exist, as for operator you may want to test unmerged code and cloning from github would prohibit that.

scripts/shared/lib/deploy_bundle
```diff
@@ -0,0 +1,249 @@
...
+  image: ${INDEX_IMG}
+  displayName: Submariner Catalog Source
+  publisher: Submariner.io (Test)
+  updateStrategy:
+    registryPoll:
+      interval: 5m
+EOF
+
+  # Wait for the catalogSource Readiness
+  if ! (timeout 5m bash -c "until [[ $(kubectl --context="${cluster}" get catalogsource -n ${MARKETPLACE_NAMESPACE} submariner-catalog-source -o jsonpath='{.status.connectionState.lastObservedState}') -eq 'READY' ]]; do sleep 10; done"); then
```
mkolesnik, line 51: Can't you use `with_retries` for this retrying logic?

scripts/shared/lib/deploy_bundle
```diff
@@ -0,0 +1,249 @@
...
+  namespace: ${MARKETPLACE_NAMESPACE}
+spec:
+  sourceType: grpc
+  image: ${INDEX_IMG}
+  displayName: Submariner Catalog Source
+  publisher: Submariner.io (Test)
+  updateStrategy:
+    registryPoll:
+      interval: 5m
+EOF
```
mkolesnik: This would be better if it was in a real YAML file which gets validated. You can use `render_template` function from `clusters.sh`

scripts/shared/lib/deploy_bundle
```diff
@@ -0,0 +1,249 @@
...
+cat <<EOF | kubectl apply -f -
+apiVersion: operators.coreos.com/v1
+kind: OperatorGroup
+metadata:
+  name: submariner-group
+  namespace: ${SUBM_NS}
+spec:
+  targetNamespaces:
+    - ${SUBM_NS}
+EOF
```
mkolesnik: This would be better if it was in a real YAML file which gets validated. You can use `render_template` function from `clusters.sh`

scripts/shared/lib/deploy_bundle
```diff
@@ -0,0 +1,249 @@
...
+  name: submariner-subscription
+  namespace: ${SUBM_NS}
+spec:
+  channel: ${CHANNEL}
+  installPlanApproval: Manual
+  name: ${OPERATOR_NAME}
+  source: submariner-catalog-source
+  sourceNamespace: ${MARKETPLACE_NAMESPACE}
+  startingCSV: ${OPERATOR_NAME}.v${VERSION}
+EOF
```
mkolesnik: This would be better if it was in a real YAML file which gets validated. You can use `render_template` function from `clusters.sh`

scripts/shared/lib/deploy_bundle
```diff
@@ -0,0 +1,249 @@
...
+kind: Broker
+metadata:
+  name: submariner-broker
+  namespace: ${SUBM_NS}
+spec:
+  components: [${components}]
+  defaultGlobalnetClusterSize: 8192
+  globalnetEnabled: ${globalnet}
+  globalnetCIDRRange: 169.254.0.0/16
+EOF
```
mkolesnik: This would be better if it was in a real YAML file which gets validated. You can use `render_template` function from `clusters.sh`

scripts/shared/lib/deploy_bundle
```diff
@@ -0,0 +1,249 @@
...
+  ceIPSecIKEPort: ${CE_IPSEC_IKEPORT}
+  ceIPSecNATTPort: ${CE_IPSEC_NATTPORT}
+  namespace: "${SUBM_NS}"
+  repository: "${SUBM_IMAGE_REPO}"
+  version: "${SUBM_IMAGE_TAG}"
+  connectionHealthCheck:
+    enabled: true
+    intervalSeconds: 1
+    maxPacketLossCount: 5
+EOF
```
mkolesnik: This would be better if it was in a real YAML file which gets validated. You can use `render_template` function from `clusters.sh`

scripts/shared/lib/deploy_bundle
```diff
@@ -0,0 +1,249 @@
...
+  kubectl get csv -n "${SUBM_NS}" --ignore-not-found
+  kubectl get pods -n "${SUBM_NS}" --ignore-not-found
+
+  echo "[INFO](${cluster}) Bundle installed"
+}
+
+function setup_broker() {
+  local brokerClientSecret
+
+  if ! (timeout 5m bash -c "until kubectl --context=${cluster} get crds brokers.submariner.io > /dev/null 2>&1; do sleep 10; done"); then
```
mkolesnik, line 101: Can't you use `with_retries` for this retrying logic?

scripts/shared/lib/deploy_bundle
```diff
@@ -0,0 +1,249 @@
...
+  namespace: ${SUBM_NS}
+spec:
+  components: [${components}]
+  defaultGlobalnetClusterSize: 8192
+  globalnetEnabled: ${globalnet}
+  globalnetCIDRRange: 169.254.0.0/16
+EOF
+
+  echo "[INFO] Wait for the broker readiness..."
+  if ! (timeout 5m bash -c "until kubectl --context=${cluster} get brokers.submariner.io submariner-broker -n ${SUBM_NS} > /dev/null 2>&1; do sleep 10; done"); then
```
mkolesnik, line 125: Can't you use `with_retries` for this retrying logic?

scripts/shared/lib/deploy_bundle
```diff
@@ -0,0 +1,249 @@
...
+        echo "Skipping installation as requested in cluster settings"
+        return
+  fi
+
+  if kubectl wait --for=condition=Ready pods -l app=submariner-operator -n "${SUBM_NS}" --timeout=60s > /dev/null 2>&1; then
+        echo "[WARN](${cluster}) Submariner already installed, skipping installation..."
+        return
+  fi
+
+  if ! (timeout 5m bash -c "until kubectl --context=${cluster} get crds submariners.submariner.io > /dev/null 2>&1; do sleep 10; done"); then
```
mkolesnik, line 151: Can't you use `with_retries` for this retrying logic?

mkolesnik (changes requested):

Dockerfile.dapper
SteveMattar: The calculated version does not always follow the semver format.
This is how I intend to use it:
https://github.com/submariner-io/submariner-operator/pull/1476/commits/4b1ff16a91cf60f39e37d51bc57eb7e3b050e90c#diff-0a7959ee618128cab3d15fc4fbe17c29729ca35ebd6cbbf6596bf8753a6ad56fR41-R45

mkolesnik: The calculated version is not intended to be semver, it's for testing and/or publishing dev images and doesnt have semver semantics.
If you want semver you should calculate it yourself.

Anyhow, in your case no need to pass any new env vars anyhow, you can call it like this `make e2e using=olm VERSION=${{ env.VERSION }} FROM_VERSION=${{ env.FROM_VERSION }} CHANNEL=${{ env.CHANNEL }}`

Make makes sure to pass them to the command properly (you can try locally and see).

scripts/shared/clusters.sh
SteveMattar: Thanks @mkolesnik  I moved these operations to the `deploy.sh` script and forgot to remove the description.

scripts/shared/lib/deploy_bundle
SteveMattar: @dfarrell07 @mkolesnik you are right. The idea was to run this test periodically but it can definitely be part of the PR tests.

Dockerfile.dapper
SteveMattar: yes I'm already doing so please see this [part](https://github.com/submariner-io/submariner-operator/commit/4b1ff16a91cf60f39e37d51bc57eb7e3b050e90c#diff-0a7959ee618128cab3d15fc4fbe17c29729ca35ebd6cbbf6596bf8753a6ad56fR41-R45)
Unfortunately it does not work without updating the `DAPPER_ENV`.
Just in case you missed, the `Dockerfile.dapper` does not exist on the submariner-operator repo anymore. It is pulled directly from shipyard.

scripts/shared/lib/deploy_bundle
SteveMattar: `with_retries` does not support timeout does it?

Dockerfile.dapper
SteveMattar: @mkolesnik it works :+1:  I used these variables as arguments of `make` instead of environment variables.

scripts/shared/lib/deploy_bundle
skitt: > @skitt yes, there are some differences. `subctl` handles everything around the broker deployment whether it's RBAC or SCC. We would like to make sure we don't miss anything in our guide for users who are not using subctl.

Sorry, my comment wasn’t clear — I wasn’t commenting on differences between `subctl` and the bundle deployment, but rather the fact that we’re cloning the repo from the devel branch, which introduces a similar dependency as using devel `subctl`.

SteveMattar: @skitt thanks for your comment, so after the last change and as @mkolesnik was suggesting, I clone the repo only if it does not exist.

nyechiel:
ping @SteveMattar @skitt

SteveMattar:
I updated the branch
Please ignore the `Linting/Commit Size` as it was added way after I created this PR. It's very hard to break the commit now.

skitt:
> ping @SteveMattar @skitt

I approved the PR on July 16, and stand by that ;-).

@dfarrell07 and @mkolesnik could you update your review?

nyechiel:
> > ping @SteveMattar @skitt
> 
> I approved the PR on July 16, and stand by that ;-).
> 
> @dfarrell07 and @mkolesnik could you update your review?

Ah sorry, my bad ;)

dfarrell07:
> This is awesome/exciting @SteveMattar, thank you! If we add "bundle" to the deploy script tests I think we'll start getting interesting test coverage. https://github.com/submariner-io/shipyard/blob/devel/.github/workflows/scripts.yml#L46

I'm still hesitant to approve this without test coverage, given that we've recently agreed to [require test coverage for new features](https://submariner.io/development/code-review/#test-new-functionality) (as part of CNCF Sandbox process). I see https://github.com/submariner-io/submariner-operator/pull/1476/commits/144e0029cc2bb36afb7cae010292950a1be0bd72 makes various fixes to make testing possible, and also adds periodic jobs. Maybe the fixes and (at least) the deploy scripts job quoted above should be a part of this PR? Or if you don't want to mess with this PR, at least add the deploy scripts job as a dependent PR so we can see it working and have it ready to merge at the ~same time.

SteveMattar:
> 
> I'm still hesitant to approve this without test coverage, given that we've recently agreed to [require test coverage for new features](https://submariner.io/development/code-review/#test-new-functionality) (as part of CNCF Sandbox process). I see [submariner-io/submariner-operator@144e002](https://github.com/submariner-io/submariner-operator/commit/144e0029cc2bb36afb7cae010292950a1be0bd72) makes various fixes to make testing possible, and also adds periodic jobs. Maybe the fixes and (at least) the deploy scripts job quoted above should be a part of this PR? Or if you don't want to mess with this PR, at least add the deploy scripts job as a dependent PR so we can see it working and have it ready to merge at the ~same time.

@dfarrell07 sure, I can add the new deploytool [here](https://github.com/submariner-io/shipyard/blob/devel/.github/workflows/scripts.yml#L46)

SteveMattar:
@dfarrell07 the bundle is part of the deploy script tests but there are no tags for this branch so currently the bundle deployment tests are failing.

dfarrell07 (approved):
Thanks for adding those tests @SteveMattar! They are failing for now because of dependency spaghetti, not something missing/wrong. IMHO let's merge this and the related Operator PR at the ~same time, leaving them not-required, and see if that resolves the issue (as expected).

mkolesnik (approved):
Let's merge as is for now and @SteveMattar will open tech debt issues for unhandled parts

PR #600, opened 2021-07-12
Add GlobalPodIP to EndpointType

vthapar:
With support for Headless services in Globalnetv2,
we need to distinguish between GlobalIP for Service
and Pod in EndpointType to allow different E2E test
cases.

* Add `GlobalPodIP`
* Add `GlobalServiceIP = GlobalIP`
* Replace all references to `GlobalIP` with `GlobalServiceIP`

### DIFF
--- a/test/e2e/tcp/connectivity.go
+++ b/test/e2e/tcp/connectivity.go
@@ -33,7 +33,10 @@ type EndpointType int
 const (
 	PodIP EndpointType = iota
 	ServiceIP
+	// TODO: Remove GlobalIP once all consumer code switches to GlobalServiceIP
 	GlobalIP
+	GlobalPodIP
+	GlobalServiceIP = GlobalIP
 )
 
 type ConnectivityTestParams struct {
@@ -66,7 +69,7 @@ func RunConnectivityTest(p ConnectivityTestParams) (*framework.NetworkPod, *fram
 	Expect(connectorPod.TerminationMessage).To(ContainSubstring(listenerPod.Config.Data))
 
 	if p.Networking == framework.PodNetworking {
-		if p.ToEndpointType == GlobalIP {
+		if p.ToEndpointType == GlobalServiceIP {
 			// When Globalnet is enabled (i.e., remoteEndpoint is a globalIP) and POD uses PodNetworking,
 			// Globalnet Controller MASQUERADEs the source-ip of the POD to the corresponding global-ip
 			// that is assigned to the POD.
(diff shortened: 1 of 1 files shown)


test/e2e/tcp/connectivity.go
```diff
@@ -33,7 +33,10 @@ type EndpointType int
 const (
 	PodIP EndpointType = iota
 	ServiceIP
+	// TODO: Remove GlobalIP once all consumer code switches to GlobalServiceIP
 	GlobalIP
+	GlobalPodIP
```
mangelajo, line 38: But GlobalPodIP is not used anywhere yet, right?

sridhargaddam (approved):

test/e2e/tcp/connectivity.go
```diff
@@ -33,7 +33,10 @@ type EndpointType int
...
+	GlobalServiceIP = GlobalIP
```
mangelajo, line 39: Neither GlobalIP I see.

I think it's OK to break the code when somebody updates to the newer version of shipyard, we can update it to make an explicit choice of GlobalPodIP (let's add it only once implemented) or GlobalServiceIP.

mangelajo (commented):
Can we implement GlobalPodIP and remove GlobalIP in one go?

This is something that can be resolved once we switch projects to the new shipyard,
seems like an easy constant replacement.

mangelajo (changes requested):
See if my previous comment is ok.

test/e2e/tcp/connectivity.go
sridhargaddam: @vthapar will be adding e2e tests in Submariner code for GlobalPodIP once this is merged.

test/e2e/tcp/connectivity.go
sridhargaddam: GlobalIP is atleast used in Submariner e2e tests. If we remove the GlobalIP constant, it might break the dependent components. The idea was to have it in this manner until we migrate all the dependent components and remove it once all are updated.

mangelajo:
I'd be also fine with the current strategy if you @sridhargaddam @vthapar think it's better.

test/e2e/tcp/connectivity.go
skitt: I think Miguel’s point is that the upgrade won’t happen automatically, we’ll need to bump `go.mod` in the projects using this, and while we’re at it we can fix any breakage.

mangelajo (approved):
Ok, I'm fine to move on with the approach as we implement the pod IPs and do a transition.

Let's remember to remove the old const later.

test/e2e/tcp/connectivity.go
mangelajo: Yeah, that's it, at that point we are good to change the constant, and no risk of leaving an old constant lying around.

sridhargaddam: Okay, in that case, we can remove it now itself.

PR #624, opened 2021-08-10
Backport: Subctl fallback devel 0.10

skitt:
During releases from branches, the e2e jobs look for a subctl matching
the target release; but that doesn't exist yet, so the download fails.
To avoid that, if installing the target release version fails, fall
back to devel subctl (which should provide the behaviour we need for a
release, and is more appropriate than the default, which is the latest
release).

(cherry picked from commit d8ec5cea2f34cc654459615aaa04d3fb2e9d7475)

### DIFF
--- a/scripts/shared/deploy.sh
+++ b/scripts/shared/deploy.sh
@@ -48,7 +48,8 @@ declare_cidrs
 declare_kubeconfig
 
 # Always get subctl since we're using moving versions, and having it in the image results in a stale cached one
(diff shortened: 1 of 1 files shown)


dfarrell07 (approved):

tpantelis (approved):

dfarrell07:
> PR targets release-0.9

This will never pass, I guess we need a CI fix for release-0.10 branch.

PR #628, opened 2021-08-23
Fix yamllint strict config, fail on warnings

dfarrell07:
Setting yamllint to exit non-zero on warnings doesn't work from the
.yamllint config file. Set the --strict flag at CLI invocation.

Signed-off-by: Daniel Farrell <dfarrell@redhat.com>

PR #631, opened 2021-08-24
Implement .mdignore for ignoring Markdown files

dfarrell07:
Add logic to read Markdown files to ignore when running markdownlint
from an .mdignore file in the root of each repo. This allows
configuration of the linter close to the source, and config to be
adjusted per-repo.

The submariner-operator repo has an apis/vendor directory that contains
external repos with markdown files. They should not be linted, and are
causing the Consuming Projects/Lint (submariner-operator) CI to fail.

Go mod vendoring directories are ignored by default.

### DIFF
--- a/Makefile.inc
+++ b/Makefile.inc
@@ -142,7 +142,11 @@ lint: gitlint golangci-lint markdownlint yamllint
 
 # [markdownlint] validates Markdown files in the project
 markdownlint:
-	markdownlint markdownlint -c .markdownlint.yml -i vendor .
+	md_ignored=($(patsubst %/modules.txt,%,$(VENDOR_MODULES))); \
+	if [ -r .mdignore ]; then \
+		md_ignored+=($$(< .mdignore)); \
+	fi; \
+	markdownlint -c .markdownlint.yml $${md_ignored[@]/#/-i } .
 
 # [yamllint] validates YAML files in the project
 yamllint:


dfarrell07:
This will fix the YAML linting failure in Consuming Projects / Lint (submariner-operator): https://github.com/submariner-io/submariner-operator/pull/1551

skitt (changes requested):
`submariner` would also need an exception for `pkg/apis/vendor`, if we wanted to play it safe. Wouldn’t it be better to provide a way for projects to add their own exceptions?

dfarrell07:
> `submariner` would also need an exception for `pkg/apis/vendor`, if we wanted to play it safe. Wouldn’t it be better to provide a way for projects to add their own exceptions?

I sent a second commit with an attempt at this, but I'm struggling with the bash-inside-Makefile peculiarities. Suggestions/feedback appreciated @skitt. :pray: 

I'm also now seemingly blocked by some infra issue (for the last hour or so)

```
[+] Building 30.2s (3/3) FINISHED                                                                                                            
 => [internal] load build definition from Dockerfile.shipyard-linting                                                                   0.1s
 => => transferring dockerfile: 1.70kB                                                                                                  0.0s
 => [internal] load .dockerignore                                                                                                       0.1s
 => => transferring context: 117B                                                                                                       0.0s
 => ERROR [internal] load metadata for docker.io/library/alpine:latest                                                                 30.0s
------
 > [internal] load metadata for docker.io/library/alpine:latest:
------
Dockerfile.shipyard-linting:1
--------------------
   1 | >>> FROM alpine
   2 |     
   3 |     ENV DAPPER_HOST_ARCH=amd64 SHELL=/bin/bash \
--------------------
error: failed to solve: alpine: failed to do request: Head "https://registry-1.docker.io/v2/library/alpine/manifests/latest": dial tcp: i/o timeout
make: *** [Makefile.images:24: package/.image.shipyard-linting] Error 1
```

dfarrell07:
FYI, the job is passing but I don't think it's actually working:

```
[linting]$ make markdownlint
md_ignored=""; \
if [ -r .mdignore ]; then \
	while read -r ignore_line; do \
		echo  ; \
		if [ -r  ]; then \
			md_ignored=" --ignore "; \
		fi \
	done < .mdignore ; \
	echo  ; \
	markdownlint -c .markdownlint.yml  . ; \
else \
	markdownlint -c .markdownlint.yml . ; \
fi
```

Makefile.inc
```diff
@@ -142,7 +142,19 @@ lint: gitlint golangci-lint markdownlint yamllint
...
+			echo $(ignore_line) ; \
+			if [ -r $(ignore_line) ]; then \
+				md_ignored="$(md_ignored) --ignore $(line)"; \
+			fi \
+		done < .mdignore ; \
+		echo $(md_ignored) ; \
+		markdownlint -c .markdownlint.yml $(md_ignored) . ; \
+	else \
+		markdownlint -c .markdownlint.yml . ; \
+	fi
```
skitt: I haven’t run this through `shellcheck`, but I have checked it works:

```suggestion
	md_ignored=(vendor); \
	if [ -r .mdignore ]; then \
		md_ignored+=($$(< .mdignore)); \
	fi; \
	markdownlint -c .markdownlint.yml $${md_ignored[@]/#/-i }
```

skitt (changes requested):

Makefile.inc
dfarrell07: > 			if [ -r $(ignore_line) ]; then \

The thought with this extra bit was to verify that each chunk we read from .mdignore and pass to markdownlint is a valid path to a file, to avoid opening up a direct path to the shell and mitigate potential attacks with crafted .mdignore lines. That may not be worth the effort/complexity though.

skitt (approved):
Fair point about the security aspect, but attackers could just submit a PR with a GHA ;-). I haven’t been successful in trying to break this approach yet with malicious `.mdignore` content.

tpantelis (approved):

dfarrell07:
> but attackers could just submit a PR with a GHA ;-).

Yeah exactly, or replace the make target with `rm -rf /` to have an impact locally ;)

dfarrell07:
Something seems wrong, `Consuming Projects / Lint (submariner-operator)` shouldn't be passing yet. It should be failing in the same way as https://github.com/submariner-io/shipyard/pull/628#issuecomment-915257513 until we add an .mdignore file with `apis/vendor` to submariner-operator

dfarrell07:
Yeah, it seems like it's failing but (confirmed locally) `markdownlint garbage` exits 0 so the job isn't failing.

```
make markdownlint
md_ignored=(vendor); \
if [ -r .mdignore ]; then \
	md_ignored+=($(< .mdignore)); \
fi; \
markdownlint -c .markdownlint.yml ${md_ignored[@]/#/-i }
Usage: markdownlint [options] <files|directories|globs>

MarkdownLint Command Line Interface

Options:
  -V, --version                               output the version number
  -c, --config [configFile]                   configuration file (JSON, JSONC, JS, or YAML)
  -d, --dot                                   include files/folders with a dot (for example `.github`)
  -f, --fix                                   fix basic errors (does not work with STDIN)
  -i, --ignore [file|directory|glob]          file(s) to ignore/exclude (default: ["vendor","vendor"])
  -j, --json                                  write issues in json format
  -o, --output [outputFile]                   write issues to file (no console)
  -p, --ignore-path [file]                    path to file with ignore pattern(s)
  -r, --rules  [file|directory|glob|package]  custom rule files (default: [])
  -s, --stdin                                 read from STDIN (does not work with files)
  -h, --help                                  display help for command
```

Makefile.inc
```diff
@@ -142,7 +142,11 @@ lint: gitlint golangci-lint markdownlint yamllint
 
 # [markdownlint] validates Markdown files in the project
 markdownlint:
-	markdownlint markdownlint -c .markdownlint.yml -i vendor .
+	md_ignored=(vendor); \
+	if [ -r .mdignore ]; then \
+		md_ignored+=($$(< .mdignore)); \
+	fi; \
+	markdownlint -c .markdownlint.yml $${md_ignored[@]/#/-i }
```
skitt: Oops, sorry,

```suggestion
	markdownlint -c .markdownlint.yml $${md_ignored[@]/#/-i } .
```

skitt (changes requested):

Makefile.inc
dfarrell07: Gave me a chance to make some fixes to the commit message too. :wink: Thanks @skitt!

dfarrell07:
Perfect, it's now correctly failing. :partying_face:

```
apis/vendor/k8s.io/klog/v2/README.md:89:1 MD010/no-hard-tabs Hard tabs [Column: 1]
apis/vendor/k8s.io/klog/v2/README.md:91:1 MD010/no-hard-tabs Hard tabs [Column: 1]
apis/vendor/k8s.io/klog/v2/README.md:93 MD012/no-multiple-blanks Multiple consecutive blank lines [Expected: 1; Actual: 2]
make: *** [/opt/shipyard/Makefile.inc:145: markdownlint] Error 1
```

dfarrell07:
Here's a PR to add the required .mdignore to submariner-operator: https://github.com/submariner-io/submariner-operator/pull/1562

skitt:
Alternatively, instead of adding `vendor` exclusions everywhere, we could piggy-back on top of #634:

```
md_ignored=($(patsubst %/modules.txt,%,$(VENDOR_MODULES)))
```

dfarrell07:
Depends on https://github.com/submariner-io/shipyard/pull/634

.mdignore
```diff
@@ -0,0 +1 @@
+vendor
```
skitt: Since `vendor` is handled in the `markdownlint` target, this isn’t necessary any more (but worth keeping as an example?).

skitt (approved):

skitt:
The Submariner failure is weird; I can’t reproduce it locally...

skitt:
Oh wait it’s `yamllint`...

skitt:
Depends on https://github.com/submariner-io/submariner/pull/1547

dfarrell07:
We didn't actually need https://github.com/submariner-io/submariner-operator/pull/1562, given the smarter default ignore list, right? :sweat_smile:

skitt:
> We didn't actually need [submariner-io/submariner-operator#1562](https://github.com/submariner-io/submariner-operator/pull/1562), given the smarter default ignore list, right? sweat_smile

Yup, we didn’t!

PR #645, opened 2021-09-28
Fix service discovery deploy conditional

dfarrell07:
Invert the conditional logic to match the intention, fixing service
discovery deployments with the operator.

### DIFF
--- a/scripts/shared/lib/deploy_operator
+++ b/scripts/shared/lib/deploy_operator
@@ -20,7 +20,7 @@ function setup_broker() {
     local gn
     local sd
     [[ $globalnet != true ]] || gn="--globalnet"
-    if [[ $service_discovery != true ]]; then
+    if [[ $service_discovery == true ]]; then
         sd="--components service-discovery,connectivity"
     else
         sd="--components connectivity"


skitt (approved):

dfarrell07:
Is there a good reason to use `[[ $service_discovery != true ]]` vs `[[ $service_discovery == true ]]`? If not, it seems like the second option is less human-bug-prone.

I suspect it's just a legacy of how it was inlined a few commits back:

```
[[ $service_discovery != true ]] || sd="--components service-discovery,connectivity"
```

skitt:
> Is there a good reason to use `[[ $service_discovery != true ]]` vs `[[ $service_discovery == true ]]`? If not, it seems like the second option is less human-bug-prone.
> 
> I suspect it's just a legacy of how it was inlined a few commits back:
> 
> ```
> [[ $service_discovery != true ]] || sd="--components service-discovery,connectivity"
> ```

Right, I suspect the latter as well.

Whether one is better than the other depends on the exact goal (assuming the test is done the right way round): one approach excludes the named value, the other requires it. This isn’t visible with the current approach; but one can choose between `[[ $service_discovery != false ]]` and `[[ $service_discovery == true ]]` depending on what the default should be.

skitt (approved):

tpantelis (approved):

sridhargaddam (approved):

PR #654, opened 2021-10-25
API to fetch nodes in the clusters

sridhargaddam:

### DIFF
--- a/test/e2e/framework/framework.go
+++ b/test/e2e/framework/framework.go
@@ -260,6 +260,20 @@ func DetectGlobalnet() {
 	})
 }
 
+func InitNumClusterNodes() error {
+	TestContext.NumNodesInCluster = map[ClusterIndex]int{}
+	for i := range KubeClients {
+		nodes, err := KubeClients[i].CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{})
+		if err != nil {
+			return err
+		}
+
+		TestContext.NumNodesInCluster[ClusterIndex(i)] = len(nodes.Items)
+	}
+
+	return nil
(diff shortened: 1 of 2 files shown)


tpantelis:
So this fetches and globally stores the number of nodes. Can you explain a bit in the summary why this is needed and how will it be used?

sridhargaddam:
> So this fetches and globally stores the number of nodes. Can you explain a bit in the summary why this is needed and how will it be used?

I explained the use-case in the actual PR where this API is used. Please see - https://github.com/submariner-io/submariner/pull/1575

tpantelis (approved):

test/e2e/framework/framework.go
```diff
@@ -260,6 +260,20 @@ func DetectGlobalnet() {
 	})
 }
 
+func FetchClusterNodes() error {
```
mkolesnik: The name is a bit confusing as you're actually initializing number of nodes per cluster, I'd change it to something clearer

mkolesnik (changes requested):

test/e2e/framework/framework.go
sridhargaddam: Its called as part of init method as you can see here - https://github.com/submariner-io/submariner/pull/1575 
where it's updating the cluster nodes. What name would you prefer?

tpantelis:
```suggestion
func InitNumClusterNodes() error {
```

sridhargaddam: Done

mkolesnik (approved):

PR #682, opened 2021-11-18
Drop the requirement for GCC

skitt:
By disabling "go vet" we can remove the requirement for GCC when
testing with race detection.


skitt:
`golang` depends on `gcc` so this makes no difference.

PR #692, opened 2021-12-22
Handle local builds in our Dapper infrastructure 

skitt:
This is based on
<submariner-io/submariner#1607> but adapted so
that it can be shared across all projects. When LOCAL_BUILD is
defined, assume that Shipyard is available in ../shipyard (relative to
the top directory of the current project), and instead of running
Dapper, run Dapper-hosted commands directly.

This relies on DAPPER_ARGS for Dapper arguments throughout, instead of
specifying Dapper arguments in the Dapper invocation itself; that way,
local builds can ignore DAPPER_ARGS entirely. This produces the
expected behaviour when choosing a base image: the command to run
inside the chosen base image is run directly. It changes "make shell"
however, but in a sensible way; Make determines that there is nothing
to do and returns to whatever called it, typically a shell in the
right environment.

### DIFF
--- a/Makefile.dapper
+++ b/Makefile.dapper
@@ -9,6 +9,22 @@ NON_DAPPER_GOALS += .dapper shell targets $(LINTING_GOALS)
 export MAKEFLAGS
 export PROJECT
 
+# Define LOCAL_BUILD to build directly on the host and not inside a Dapper container
+ifdef LOCAL_BUILD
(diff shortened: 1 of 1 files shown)


dfarrell07 (approved):

mkolesnik (approved):

dfarrell07:
The failing bundle jobs are not related to this PR, have been showing up elsewhere.

PR #694, opened 2021-12-30
Added timestamp to echo of trapped commands 

davidohana:
This helps debugging timeout issues since now you can now when last command started.

Example:
image


davidohana:
Not clear to me how failed tests related to this small change. Could this be transient test failures due to timeouts etc.?

sridhargaddam:
> Not clear to me how failed tests related to this small change. Could this be transient test failures due to timeouts etc.?

The failure doesn't seem to be related to this issue. The Submariner pods are failing with ImagePullBackOff error

CC: @mkolesnik @skitt 
```
    Normal   Scheduled    4m21s                  default-scheduler  Successfully assigned submariner-operator/submariner-routeagent-bfjnr to cluster2-control-plane
    Warning  FailedMount  4m19s                  kubelet            MountVolume.SetUp failed for volume "submariner-routeagent-token-gjvng" : failed to sync secret cache: timed out waiting for the condition
    Normal   Pulling      2m49s (x4 over 4m19s)  kubelet            Pulling image "quay.io/submariner/submariner-route-agent:devel-f04b9b9"
    Warning  Failed       2m49s (x4 over 4m18s)  kubelet            Failed to pull image "quay.io/submariner/submariner-route-agent:devel-f04b9b9": rpc error: code = NotFound desc = failed to pull and unpack image "quay.io/submariner/submariner-route-agent:devel-f04b9b9": failed to resolve reference "quay.io/submariner/submariner-route-agent:devel-f04b9b9": quay.io/submariner/submariner-route-agent:devel-f04b9b9: not found
    Warning  Failed       2m49s (x4 over 4m18s)  kubelet            Error: ErrImagePull
    Normal   BackOff      2m35s (x6 over 4m18s)  kubelet            Back-off pulling image "quay.io/submariner/submariner-route-agent:devel-f04b9b9"
    Warning  Failed       2m35s (x6 over 4m18s)  kubelet            Error: ImagePullBackOff
```

scripts/shared/lib/debug_functions
```diff
@@ -13,7 +13,7 @@ function trap_commands() {
           ! [[ "$cmd" =~ ^(echo|read|\[|while|for|local) ]] &&
           { [[ ! "$cmd" =~ \$\( ]] || cmd="${cmd@Q}"; } &&
           ctxt="${cluster:+[${cluster}]}" &&
-          cmd=`eval echo "[${PWD##*/}]\$ $ctxt $cmd" 2>/dev/null` &&
+          cmd=`eval echo "[$(date -u)] [${PWD##*/}]\$ $ctxt $cmd" 2>/dev/null` &&
```
skitt: The default format only shows the time with second granularity, which might not be sufficient in all cases; it also shows the date, which isn’t useful here.

```suggestion
          cmd=`eval echo "[$(date +%R:%S:%N)] [${PWD##*/}]\$ $ctxt $cmd" 2>/dev/null` &&
```

skitt (changes requested):

scripts/shared/lib/debug_functions
davidohana: I think ms precision is sufficient, updated.
btw `%N` is not supported on macOS `date` command so we get something like `09:48:01:N`, but I think it's acceptable. We can also use Perl of `gdate` which is cross-platform but I guess we cannot rely on those being installed everywhere the scripts are intended to be executed.

skitt: Thanks, ms precision is indeed sufficient. The target platform is the Shipyard base image, non-container-based local builds are nice to have but I agree display differences such as this are acceptable. The Shipyard base image doesn’t include Perl.

skitt (approved):

tpantelis (approved):

PR #704, opened 2022-01-13
Ovn k kind updates

### LINKED ISSUE #703: Update OVN-K image, Upstream KIND hacks, Remove Submariner/ovn-kubernetes fork
The Image being used in shipyard for ovn-kubernetes KIND clusters is 

`local OVN_SRC_IMAGE="quay.io/vthapar/ovn-daemonset-f:latest"` 

Which is fairly stale and over a year old.  We also maintain a fork of the ovn-kubernetes repo to keep some kind script hacks that should be able to be up streamed in ovn-kubernetes. 

My goal with this issue is to accomplish three things 

- [x]  Upstream some work to setup a continuous build pipeline in upstream OVN-K the images will be updated on every push to master and stored automatically in the github container registy (ghcr.io) That work can be found here https://github.com/ovn-org/ovn-kubernetes/pull/2747
- [ ]  Upstream the submariner relevant KIND script hacks to OVN-K these include adding the ability to run the script from a container, explicitly supply a kubeconfig path, explicitly name the cluster, provide a custom DNSdomain, and to integrate with a local docker registry 
- [ ] Ultimately remove the submariner/ovn-kubernetes fork so we have one less thing to maintain

astoycos:
This PR works to close #703 

It fixes the makefile target `make clusters using=ovn` and then ensures we use the work upstream in ovn-k. This cannot be merged until https://github.com/ovn-org/ovn-kubernetes/pull/2747 goes in 

Depends on https://github.com/ovn-org/ovn-kubernetes/pull/2747


astoycos:
Still waiting on https://github.com/ovn-org/ovn-kubernetes/pull/2747 to merge :(

astoycos:
Note: I still need to update the git submodule to point to upstream ovn-k rather than the submariner fork here.

skitt:
> Note: I still need to update the git submodule to point to upstream ovn-k rather than the submariner fork here.

If we can use upstream ovn-k directly now, do we still need a submodule?

astoycos:
Well we still need to source the ovn-k kind script @skitt, what's the "most official" way to do that? It's either a submodule or download [the script](https://github.com/ovn-org/ovn-kubernetes/blob/master/contrib/kind.sh) manually every time we need it right? (unfortunately it's not packaged in the ovn-k image)

skitt:
> Well we still need to source the ovn-k kind script @skitt, what's the "most official" way to do that? It's either a submodule or download [the script](https://github.com/ovn-org/ovn-kubernetes/blob/master/contrib/kind.sh) manually every time we need it right? (unfortunately it's not packaged in the ovn-k image)

I vote for downloading it when we need it; that seems preferable to cloning a 103MiB repository even when we don’t need OVN.

astoycos:
++ I agree will implement this here,  thanks!

Makefile.inc
```diff
@@ -1,5 +1,4 @@
 , := ,
-_using = $(subst $(,), ,$(using))
```
skitt, line 2: `Makefile.inc` is shared across all projects, this will break any project relying on `_using` (`submariner`, `submariner-charts`, `submariner-operator`).

package/Dockerfile.shipyard-dapper-base
```diff
@@ -5,10 +5,10 @@ FROM fedora:35
 ARG UPX_LEVEL=-5
 ENV DAPPER_HOST_ARCH=amd64 SHIPYARD_DIR=/opt/shipyard SHELL=/bin/bash \
     DAPPER_RUN_ARGS="--net=kind"
-ENV HOST_ARCH=${DAPPER_HOST_ARCH} ARCH=${DAPPER_HOST_ARCH} PATH=/go/bin:/usr/local/go/bin:$PATH \
+ENV HOST_ARCH=${DAPPER_HOST_ARCH} ARCH=${DAPPER_HOST_ARCH} PATH=/root/.local/bin:/go/bin:/usr/local/go/bin:$PATH \
```
skitt, line 8: Do we need this? `kind.sh` is executed without relying on `PATH`; is anything else added to `/root/.local/bin`?

skitt (changes requested):

package/Dockerfile.shipyard-dapper-base
astoycos: > kind.sh is executed without relying on PATH 

Is it? When I remove that dir from PATH the kind script cannot use the `j2` binary which is installed by DNF at `/root/.local/bin/j2`  in the dapper build container.

Makefile.inc
astoycos: Ok so I guess what I probably need to do is just ensure that in the MAKEFILE we still `include MAKEFILE.inc` so that we still parse the `using=` flag before firing up the clusters, let me make sure this works

astoycos: Fixed

package/Dockerfile.shipyard-dapper-base
skitt: `kind.sh` is run as `./kind.sh` so there’s no `PATH` lookup for that; but `j2` is the answer to my question.

Makefile
```diff
@@ -13,6 +13,8 @@ ifneq (,$(DAPPER_HOST_ARCH))
 
 # Running in Dapper
 
+include Makefile.inc
```
skitt: If we need this here, line 26 should be removed (but I’m not sure why it would make a difference).

astoycos: Ah my bad.... And I think it makes the difference since currently we decide on the settings yaml (OVN `.shipyard.e2e.ovn.yml` / Standard `$(DAPPER_SOURCE)/.shipyard.e2e.yml`)  BEFORE we actually import the value of `using` from `Makefile.inc`....

skitt: Ah yes, I missed the reference to `_using` just below. So we need this line here, and the `include` further down removed.

astoycos: Fixed!

skitt (approved):

astoycos:
Huh @skitt That ^ breaks things because now we go and call `./clusters.sh` before populating the SETTINGS var  :( . 

We have a bit of a circular dependency going on here, why can't we just move 
```
ifneq (,$(filter ovn,$(_using)))
SETTINGS ?= $(DAPPER_SOURCE)/.shipyard.e2e.ovn.yml
else
SETTINGS ?= $(DAPPER_SOURCE)/.shipyard.e2e.yml
endif
``` 

to `makefile.inc`, since that's where we actually call the clusters script AND parse the `using` flag?

astoycos:
The `[Deploy (ovn)]` Job will fail until the dependent PR merges

Makefile.inc
```diff
@@ -59,6 +59,9 @@ endif
 
 ifneq (,$(filter ovn,$(_using)))
 PRELOAD_IMAGES += submariner-networkplugin-syncer
+SETTINGS ?= $(DAPPER_SOURCE)/.shipyard.e2e.ovn.yml
+else
+SETTINGS ?= $(DAPPER_SOURCE)/.shipyard.e2e.yml
```
mkolesnik: Please keep them where they were, there's no guarantee consuming projects will have exact same file names

mkolesnik (changes requested):

Makefile.inc
astoycos: Thanks for the look @mkolesnik! Just a quick question, If you look at my comment [here](https://github.com/submariner-io/shipyard/pull/704/commits/7ba768f0453ecf3827ae400ff2815986ae713811) you can see I tried to change this bit because  it seemed to break `make clusters using=ovn` since we don't populate the `using` flag internally until after picking which settings file to use....Do you have any recs on how to best fix this?

astoycos: I fixed this by parsing the `using` flag in both makefiles so that we can ensure we're always up to date. I am definitely not a MAKE expert so please let me know if this makes sense.

astoycos:
Dependent OVN-K PR has merged!

astoycos:
So in the OVN test the cluster now deploys successfully in CI, however the connectivity tests still fail, this should be fixed by https://github.com/submariner-io/submariner/pull/1689

Makefile
```diff
@@ -15,6 +15,10 @@ ifneq (,$(DAPPER_HOST_ARCH))
 
 # Running in Dapper
 
+# parse `using` flag here and in makefile.inc or the incorrct kind config yaml
```
mkolesnik:
```suggestion
# parse `using` flag here and in makefile.inc or the incorrect kind config yaml
```

nit

scripts/shared/clusters.sh
```diff
@@ -304,6 +302,16 @@ function warn_inotify() {
     fi
 }
 
+function download_ovnk() {
+    if [[ "${cluster_cni[$cluster]}" == "ovn" ]]; then
+        rm -rf ovn-kubernetes/
+        # To deploy ovn-kubernetes properly it's easiest to just clone the
```
mkolesnik:
```suggestion

        # To deploy ovn-kubernetes properly it's easiest to just clone the
```
nit

scripts/shared/clusters.sh
```diff
@@ -304,6 +302,16 @@ function warn_inotify() {
...
+        # ovn-kubernetes repo from master on the fly in order to get the required
+        # kind scripts, and manfest generation templates.
```
mkolesnik:
```suggestion
        # kind scripts, and manifest generation templates.
```
nit

scripts/shared/clusters.sh
```diff
@@ -304,6 +302,16 @@ function warn_inotify() {
     fi
 }
 
+function download_ovnk() {
+    if [[ "${cluster_cni[$cluster]}" == "ovn" ]]; then
```
mkolesnik: This won't work properly since you're not calling it in a context where `$cluster` is set.

You should either call it from `deploy_kind_ovn` and risk parallel process forks overwriting each other, or check if any of the clusters needs ovn (basically if `${cluster_cni[*]}` contains `ovn`).

Also fail fast is easier to read so if no cluster needs ovn you can `return` from the function, otherwise do what you need.

mkolesnik (changes requested):
Thanks for you work on this, some nits and one big issue :)

Makefile
astoycos: Done

scripts/shared/clusters.sh
astoycos: right so calling it in deploy_kind_ovn creates issues, I went ahead and checked if ${cluster_cni[*]} contains "ovn" and made it so that we fail fast if possible

scripts/shared/clusters.sh
astoycos: Done

scripts/shared/clusters.sh
astoycos: Updated this comment

astoycos:
I think this _should_ be ready to go @mkolesnik, PTAL if you get the chance :)

scripts/shared/clusters.sh
```diff
@@ -304,13 +302,28 @@ function warn_inotify() {
     fi
 }
 
+# If any of the clusters use OVN-K as the CNI then clone the
+# ovn-kubernetes repo from master in order to access the required
+# kind scripts, and manifest generation templates.
+function download_ovnk() {
+    if [[ ${cluster_cni[*]} != *"ovn"* ]]
+    then
```
mkolesnik:
```suggestion
    if [[ "${cluster_cni[*]}" != *"ovn"* ]]; then
```

Some nits :)

scripts/shared/clusters.sh
```diff
@@ -304,13 +302,28 @@ function warn_inotify() {
...
+        return
+    fi
+
+    echo "Cloning ovn-kubernetes from source"
+    rm -rf ovn-kubernetes/
```
mkolesnik: Should probably add `ovn-kubernetes` to `.gitignore`?

scripts/shared/clusters.sh
```diff
@@ -304,13 +302,28 @@ function warn_inotify() {
...
+    git clone https://github.com/ovn-org/ovn-kubernetes.git
```
mkolesnik:
```suggestion
    git clone https://github.com/ovn-org/ovn-kubernetes.git \
        || { git -C ovn-kubernetes fetch && git -C ovn-kubernetes reset --hard origin/master; }
```

On local envs this might be preferrable instead of cloning each time?

We could also use `git clone https://github.com/ovn-org/ovn-kubernetes.git || git -C ovn-kubernetes pull` for simplicity but if the remote branch diverges (e.g. force push) then it would fail, though on GH it's unlikely but is a **very remote** possibility.

mkolesnik (commented):
Thanks, some thoughts and some nits :)

Makefile.inc
mkolesnik: Yes there's a bit of a "chicken and egg" situation here since the `_using` part is traditionally parsed in `Makefile.inc` but `SETTINGS` is expected to be set before including it, so we were actually silently not using `ovn` anywhere for a while.

Perhaps the best solution is to move `_using` parsing into `Makefile.dapper` as that one gets loaded way before, or just create some dedicated `Makefile.variables` file (to load any such shared variables).

astoycos: Done, I added the parsing of the `using` flag to `Makefile.dapper` I think it makes it much easier to understand

scripts/shared/clusters.sh
astoycos: +1 done

scripts/shared/clusters.sh
astoycos: Yep :)

scripts/shared/clusters.sh
astoycos: Yep, went with the first option, I doubt it will break anytime soon

scripts/shared/clusters.sh
```diff
@@ -304,13 +302,28 @@ function warn_inotify() {
...
+# If any of the clusters use OVN-K as the CNI then clone the
+# ovn-kubernetes repo from master in order to access the required
+# kind scripts, and manifest generation templates.
+function download_ovnk() {
+    if [[ ${cluster_cni[*]} != *"ovn"* ]]; then
+        return
+    fi
+
+    echo "Cloning ovn-kubernetes from source"
+    rm -rf ovn-kubernetes/
```
skitt: I think part of @mkolesnik’s comment was that we shouldn’t unconditionally delete `ovn-kubernetes`, to avoid having to re-download the full git repository all the time when building locally. That’s how the suggested `git clone ... || git reset` works: if `ovn-kubernetes` already exists, `git clone` will fail, and instead `git fetch && git reset` will update the existing clone.

scripts/shared/clusters.sh
```diff
@@ -304,13 +302,28 @@ function warn_inotify() {
     fi
 }
 
+# If any of the clusters use OVN-K as the CNI then clone the
+# ovn-kubernetes repo from master in order to access the required
+# kind scripts, and manifest generation templates.
+function download_ovnk() {
+    if [[ ${cluster_cni[*]} != *"ovn"* ]]; then
```
skitt, line 309: I’m not sure this does what is needed, I’m checking.

skitt (changes requested):

scripts/shared/clusters.sh
astoycos: Ah yep of course, sorry about this, fixed!

skitt (approved):

dfarrell07:
I see the dependent PR is merged. Do we expect `make deploy using=ovn` to work at this point? It seems to be consistently failing.

I see you mention this getting `make clusters using=ovn` working. We could add a test for that here if you think that would be helpful: https://github.com/submariner-io/shipyard/blob/devel/.github/workflows/scripts.yml#L19

astoycos:
So I believe the deployment is coming up correctly @dfarrell07 However the connection test is failing 

```
Max attempts reached, failed to run 'test_connection netshoot-664cc4c98b-56nb8 100.1.48.194'!
``` 

This test will not pass until https://github.com/submariner-io/submariner/pull/1689 has merged. 

> I see you mention this getting make clusters using=ovn working. We could add a test for that here if you think that would be helpful: https://github.com/submariner-io/shipyard/blob/devel/.github/workflows/scripts.yml#L19 

100% I'll look at making a test for that, verifying that the actual CNI in use is OVN-K not weave

dfarrell07:
> This test will not pass until https://github.com/submariner-io/submariner/pull/1689 has merged.

Okay thanks for clarifying. I marked this PR as dependent on that one for clarity, hope that's right.

astoycos:
So @dfarrell07 sorry let me try and clarify...It's a bit of a circular dependency here, we actually need this PR to land before https://github.com/submariner-io/submariner/pull/1689 because it allows the submariner+ovn-K kind setup to come up with new OVN-K images, those of which are required for https://github.com/submariner-io/submariner/pull/1689's CI to pass

dfarrell07:
Okay thanks for clarifying! So it seems like we need to have both ready to merge before merging either?

astoycos:
> Okay thanks for clarifying! So it seems like we need to have both ready to merge before merging either?

For CI to be green everywhere yes.... however I think the only reason CI is even passing now with OVN is because it's running at an old/fixed ovn-k version so it's kind of like a false positive since on real OCP+ovn-k clusters it won't work. 

I don't know what the protocol is in the submariner community is on this but I'd rather see this get merged first to ensure that CI looks good on https://github.com/submariner-io/submariner/pull/1689.  Maybe we make the OVN job not required until 1689 lands to allow this to go in first, or something along those lines?

astoycos:
Further proof of the above issue -> https://github.com/submariner-io/submariner/pull/1689#issuecomment-1090887976

dfarrell07:
@astoycos Okay that makes sense, agreed. Let's get this in and let the CI correctly fail, make the OVN jobs not-required.

@tpantelis That's possible with the GitHub required job config knobs, right? Basically the params to the job count as part of the job name we can filter on? This new issue with workflow names not being able to be filtered on has me doubting. Maybe it would be good to go ahead and make that config change?

@mkolesnik Please re-review :pray:

Makefile.dapper
```diff
@@ -3,11 +3,13 @@
 
 PROJECT ?= $(notdir $(CURDIR))
 BASE_DAPPER := Dockerfile.dapper
+USING = $(subst $(,), ,$(using))
```
mkolesnik, line 6: Interesting approach, though I wonder if you also pass `USING` when calling from outside dapper if this will hold or break :)

(Not an issue just an observation)

Makefile.inc
mkolesnik, line 2: This still can;t be removed as it;s being used by consuming projects.

They will have to be updated to use `USING` and only then can this be removed

mkolesnik (changes requested):
Still have to keep `_using` until it's not used in consuming projects (no pun intended)

Makefile.dapper
astoycos: > Perhaps the best solution is to move _using parsing into Makefile.dapper as that one gets loaded way before

So I believe this is the only way to parse that flag within `Makefile.dapper` and pass it into the dapper build environment, at least from what I've seen. Unfortunately the naming `_using` makes it really hard to pass the var into the build environment  since we can only do so with environment vars

Makefile.inc
astoycos: But then we're just parsing the flag twice again..... Can we simply "translate" the variable here? i.e 

_using = ${USING} 

To ensure we only actually parse the flag a single time?

astoycos: Or we need to step way back and re-evaluate again how we populate `SETTINGS` i.e Re-verting / re-working 
https://github.com/submariner-io/shipyard/commit/a1ab011f2271478a694aeba864099f09f37ff93d

Please let me know what the correct way forward is here

dfarrell07: It makes sense to me to keep the var as a "shadow" (`_using = ${USING}`) until it can be removed.

dfarrell07: Also

> /opt/shipyard/scripts/deploy.sh --cable_driver libreswan --deploytool helm --settings /go/src/github.com/submariner-io/submariner/.shipyard.e2e.yml
2022-04-12T01:42:59.0911727Z Running with: globalnet='false', deploytool='helm', deploytool_broker_args='', deploytool_submariner_args='', settings='/go/src/github.com/submariner-io/submariner/.shipyard.e2e.yml', timeout=5m, image_tag=local, cable_driver=libreswan, service_discovery=false

https://github.com/submariner-io/shipyard/runs/5982624162?check_suite_focus=true

mkolesnik: Problem is all of these "using" cases in the consuming e2e are actually parsed on the Shipyard level, e.g. `using=lighthouse` in Submariner is not tested and that would break:
https://github.com/submariner-io/submariner-operator/blob/devel/Makefile#L61-L63

Makefile.dapper
mkolesnik: Yeah no that's not what I meant but OK :)

I was talking about what would happen if someone would theoretically try running `make e2e USING="bla bla"` since it would be both an env var and a make flag but it was just a random thought about the weirdness of Make ;)

Makefile.inc
dfarrell07: Yeah that makes sense, unfortunately the code path for `using` in the consuming tests isn't the same as if we were running in a consuming project directly.

Makefile.dapper
astoycos: Ah I gotcha, yeah I'm not sure about that, we set and export USING within the Makefile so hopefully anything user provided would be overridden!

Makefile.inc
astoycos: > It makes sense to me to keep the var as a "shadow" (_using = ${USING}) until it can be removed. 

Done and tested manually, all seems to work!

mkolesnik (approved):

dfarrell07 (approved):

dfarrell07:
IMHO we should make this not-required here and the main repo, merge, and then review/merge the fix in the main repo. Need help from an Owner.

Thanks @astoycos! :partying_face:

PR #720, opened 2022-02-04
Use Buildah to build and release images

skitt:
This replaces Docker buildx, inside the Dapper image for now.
Multi-arch builds are supported using manifests.

The default platform is now the runtime platform, rather than
"linux/amd64"; this should make it simpler to build local images for
local consumption on ARM systems in particular.

The containerd and selinux* packages are no longer installed, so we no
longer need to remove them. Explicitly installing nodejs means we can
also remove npm and python3-pip using dnf.

Multiarch images need to be explicitly declared as such, since some
images (e.g. shipyard-linting) can't be built on all arches.

Depends on https://github.com/submariner-io/submariner/pull/1676

### DIFF
--- a/gh-actions/release-images/action.yaml
+++ b/gh-actions/release-images/action.yaml
@@ -14,15 +14,12 @@ inputs:
 runs:
   using: "composite"
   steps:
-    - shell: bash
-      env:
-        IMAGES_ARGS: --nocache ${{ inputs.image_args }}
-      run: |
-        echo "::group::Build new images"
-        make images
-        echo "::endgroup::"
-
-    - name: Release newly built images
+    - name: Set up QEMU
+      uses: docker/setup-qemu-action@27d0a4f181a40b142cce983c5393082c365d1480
+      with:
+        image: tonistiigi/binfmt:latest
+        platforms: all
+    - name: Build and release images
       shell: bash
       env:
         QUAY_USERNAME: ${{ inputs.username }}
--- a/scripts/shared/build_image.sh
+++ b/scripts/shared/build_image.sh
@@ -2,15 +2,13 @@
 
 ## Process command line flags ##
 
-default_platform=linux/amd64
 source ${SCRIPTS_DIR}/lib/shflags
 DEFINE_string 'tag' "${DEV_VERSION}" "Tag to set for the local image"
 DEFINE_string 'repo' 'quay.io/submariner' "Quay.io repo to use for the image"
 DEFINE_string 'image' '' "Image name to build" 'i'
 DEFINE_string 'dockerfile' '' "Dockerfile to build from" 'f'
 DEFINE_string 'buildargs' '' "Build arguments to pass to 'docker build'"
-DEFINE_boolean 'cache' true "Use cached layers from latest image"
-DEFINE_string 'platform' "${default_platform}" 'Platforms to target'
+DEFINE_string 'platform' '' 'Platforms to target'
 DEFINE_string 'hash' '' "File to write the hash to" 'h'
 FLAGS "$@" || exit $?
 eval set -- "${FLAGS_ARGV}"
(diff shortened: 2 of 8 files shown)


skitt:
I need to adjust `.dapper` to take all these changes into account; however I’m still keen to get some reviews.

gh-actions/release-images/action.yaml
```diff
@@ -14,15 +14,12 @@ inputs:
 runs:
   using: "composite"
   steps:
-    - shell: bash
-      env:
-        IMAGES_ARGS: --nocache ${{ inputs.image_args }}
-      run: |
-        echo "::group::Build new images"
-        make images
-        echo "::endgroup::"
```
mkolesnik, line 23: So the separation here was for ease of debugging in case of a failure. Separating the `build` step from the `release` step makes it easier to figure out which failed.

Is it necessary to remove this separate step or can it be kept?

scripts/shared/build_image.sh
```diff
@@ -2,15 +2,13 @@
...
 ## Process command line flags ##
 
-default_platform=linux/amd64
 source ${SCRIPTS_DIR}/lib/shflags
 DEFINE_string 'tag' "${DEV_VERSION}" "Tag to set for the local image"
 DEFINE_string 'repo' 'quay.io/submariner' "Quay.io repo to use for the image"
 DEFINE_string 'image' '' "Image name to build" 'i'
 DEFINE_string 'dockerfile' '' "Dockerfile to build from" 'f'
 DEFINE_string 'buildargs' '' "Build arguments to pass to 'docker build'"
-DEFINE_boolean 'cache' true "Use cached layers from latest image"
```
mkolesnik, line 12: We had the cache flag to make sure we always pull fresh images when releasing, and cache when developing and running E2E to conserve time. Is this handled automatically now with buildah?

scripts/shared/build_image.sh
```diff
@@ -34,30 +31,41 @@ set -e
...
-        docker pull ${cache_image} || :
-    fi
-    for parent in $(grep FROM ${dockerfile} | cut -f2 -d' ' | grep -v scratch); do
-        cache_flag+=" --cache-from ${parent}"
-        docker pull ${parent} || :
-    done
-fi
+# We always use manifest builds
+buildah manifest rm "${image}" || :
+buildah manifest create "${image}" || :
```
mkolesnik, line 36: Do we want to ignore the failure in this case?

mkolesnik (commented):
Some questions

gh-actions/release-images/action.yaml
skitt: Yes, we can keep them — we can use the same approach as that used with buildx (export multi-arch images as tarballs).

scripts/shared/build_image.sh
skitt: Buildah doesn’t support `--cache-from`, basically, which is why I dropped this option.

scripts/shared/build_image.sh
skitt: When doing local development, if an image is in use in a given manifest, that manifest can’t be deleted and created again. Ideally we should figure out a way of dealing with that...

gh-actions/release-images/action.yaml
mkolesnik: I think we should if it's not a hassle, since it'll be easier to understand if the image building itself failed (e.g. due to some error downloading/installing) or if the upload failed (e.g. wrong creds, service unavailable, etc)

scripts/shared/build_image.sh
mkolesnik: OK but how does it handle caching, would it just reuse whatever layer is available in the local registry or would it always check remote registry if the layer is up to date or not?

If it's smart enough to cache when working locally as needed, while rebuilding the image when necessary, when building it for release in CI, then we can live without explicitly specifying caching behavior.

scripts/shared/build_image.sh
mkolesnik: Well an easy way is setting some env var on CI so we definitely know it's CI (or use existing GITHUB vars)

PR #721, opened 2022-02-08
Build multi-arch images with buildx

skitt:
The default platform is now the runtime platform, rather than
"linux/amd64"; this should make it simpler to build local images for
local consumption on ARM systems in particular.

Multiarch images need to be explicitly declared as such, since some
images (e.g. shipyard-linting) can't be built on all arches.

This currently relies on emulating each target architecture, using
QEMU. Some builds will be able to use cross-compilation only instead,
that will come later.

### DIFF
--- a/.github/workflows/multiarch.yml
+++ b/.github/workflows/multiarch.yml
@@ -1,0 +1,19 @@
+---
+name: Multi-arch builds
+
+on:
+  pull_request:
+
+jobs:
+  apply-suggestions-commits:
+    name: Check the multi-arch builds
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out the repository
+        uses: actions/checkout@5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f
+      - name: Set up QEMU (to support building on non-native architectures)
+        uses: docker/setup-qemu-action@27d0a4f181a40b142cce983c5393082c365d1480
+      - name: Set up buildx
+        uses: docker/setup-buildx-action@94ab11c41e45d028884a99163086648e898eed25
+      - name: Build the multi-arch images
+        run: make multiarch-images
--- a/Makefile.images
+++ b/Makefile.images
@@ -23,12 +23,24 @@ docker_deps = $(shell files=($(1) $$(awk '/COPY/ && substr($$2, 1, 7) != "--from
 package/.image.%: $$(call docker_deps,package/Dockerfile.$$*) $$(call force_image_rebuild,$$*)
 	$(SCRIPTS_DIR)/build_image.sh -i $(lastword $(subst ., ,$@)) -f $< -h $@ $(IMAGES_ARGS)
 
+# Build an OCI tarball from a Dockerfile
+.SECONDEXPANSION:
+package/%.tar: $$(call docker_deps,package/Dockerfile.$$*) $$(call force_image_rebuid,$$*)
+	image=$(firstword $(subst ., ,$(@F))) && \
+	$(SCRIPTS_DIR)/build_image.sh -i $$image -f $< --oci $@ -h package/.image.$$image $(IMAGES_ARGS)
+
 # [images] builds all the container images for the project
 # Default target to build images based on IMAGES variable
 images: $(foreach image,$(IMAGES),package/.image.$(image))
+
+# [multiarch-images] builds all multi-arch container images
+# Default target to build images based on the MULTIARCH_IMAGES variable
+# Default platforms based on the PLATFORMS variable
+multiarch-images: override IMAGES_ARGS += --platform $(PLATFORMS)
+multiarch-images: $(foreach image,$(MULTIARCH_IMAGES),package/$(image).tar)
 
 # [release-images] uploads the project's images to a public repository
 release-images:
 	$(SCRIPTS_DIR)/release_images.sh $(IMAGES) $(RELEASE_ARGS)
 
-.PHONY: images release-images
+.PHONY: images multiarch-images release-images
--- a/scripts/shared/build_image.sh
+++ b/scripts/shared/build_image.sh
@@ -2,7 +2,6 @@
 
 ## Process command line flags ##
 
-default_platform=linux/amd64
 source ${SCRIPTS_DIR}/lib/shflags
 DEFINE_string 'tag' "${DEV_VERSION}" "Tag to set for the local image"
 DEFINE_string 'repo' 'quay.io/submariner' "Quay.io repo to use for the image"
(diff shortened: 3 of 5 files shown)


scripts/shared/build_image.sh
```diff
@@ -22,19 +22,24 @@ dockerfile="${FLAGS_dockerfile}"
 buildargs="${FLAGS_buildargs}"
 platform="${FLAGS_platform}"
 hashfile="${FLAGS_hash}"
+ocifile="${FLAGS_oci}"
 [[ "${FLAGS_cache}" = "${FLAGS_TRUE}" ]] && cache=true || cache=false
 
 [[ -n "${image}" ]] || { echo "The image to build must be specified!"; exit 1; }
 [[ -n "${dockerfile}" ]] || { echo "The dockerfile to build from must be specified!"; exit 1; }
 [[ -n "${hashfile}" ]] || { echo "The file to write the hash to must be specified!"; exit 1; }
+if [[ "${platform}" =~ , && -z "${ocifile}" ]]; then
```
dfarrell07, line 31: What does regex-comparing to a comma (`=~ ,`) do? Is it something like a parameter expansion and nil check in one?

dfarrell07: We can also ask/answer this on Stackoverflow if you want, with how I currently know to describe it I don't see any existing question.

skitt: This is already covered on SE ;-): https://unix.stackexchange.com/q/340440/86440. `=~` is a regex comparison, it allows substring matches; so this is looking for commas anywhere in `"${platform}"`, which indicates that multiple platforms have been requested.

dfarrell07: > This is already covered on SE

Thanks! (I deleted my comment a few moments after posting, before seeing your reply, for fear it was there and I was just searching poorly)

dfarrell07: Ah okay, yeah I read exactly that SE post and already knew `=~`, but I was over-thinking the comma. I thought you were using it like https://unix.stackexchange.com/questions/500274/what-do-the-commas-in-var-mean, not a literal.

dfarrell07 (approved):

Makefile.images
```diff
@@ -23,12 +23,24 @@ docker_deps = $(shell files=($(1) $$(awk '/COPY/ && substr($$2, 1, 7) != "--from
...
 
+# [multiarch-images] builds all multi-arch container images
+# Default target to build images based on the MULTIARCH_IMAGES variable
+# Default platforms based on the PLATFORMS variable
+multiarch-images: override IMAGES_ARGS += --platform $(PLATFORMS)
+multiarch-images: $(foreach image,$(MULTIARCH_IMAGES),package/$(image).tar)
+
 # [release-images] uploads the project's images to a public repository
-release-images:
+release-images: multiarch-images images
```
mkolesnik: We have separated the targets on purpose, please see https://github.com/submariner-io/shipyard/blob/devel/gh-actions/release-images/action.yaml

Is there a specific reason to re-couple them?

scripts/shared/build_image.sh
```diff
@@ -47,17 +52,31 @@ if [[ "$cache" = true ]]; then
     done
 fi
 
+output_flag=--load
+[[ -z "${ocifile}" ]] || output_flag="--output=type=oci,dest=${ocifile}"
+
+# Default to linux/amd64 (for CI); platforms match Go OS/arch
+if command -v "${GO:-go}" >/dev/null; then
+    default_platform="$(${GO:-go} env GOOS)/$(${GO:-go} env GOARCH)"
```
mkolesnik, line 69: Why use default `go` if it already defaults to `go` in the Makefile? https://github.com/submariner-io/shipyard/blob/devel/Makefile.inc#L87

scripts/shared/build_image.sh
```diff
@@ -2,16 +2,16 @@
...
 DEFINE_string 'tag' "${DEV_VERSION}" "Tag to set for the local image"
 DEFINE_string 'repo' 'quay.io/submariner' "Quay.io repo to use for the image"
 DEFINE_string 'image' '' "Image name to build" 'i'
 DEFINE_string 'dockerfile' '' "Dockerfile to build from" 'f'
 DEFINE_string 'buildargs' '' "Build arguments to pass to 'docker build'"
 DEFINE_boolean 'cache' true "Use cached layers from latest image"
-DEFINE_string 'platform' "${default_platform}" 'Platforms to target'
+DEFINE_string 'platform' '' 'Platforms to target'
 DEFINE_string 'hash' '' "File to write the hash to" 'h'
+DEFINE_string 'oci' '' 'File to write an OCI tarball to instead of an image in the local registry'
```
mkolesnik, line 14: I'm guessing we'll be doing something with these tar files in a follow up PR?

mkolesnik (commented):
Couple of questions

scripts/shared/build_image.sh
mkolesnik: Ah just saw https://github.com/submariner-io/shipyard/pull/723 :)

Makefile.images
skitt: Thanks, this is a left-over from early iterations, I’ll drop it.

scripts/shared/build_image.sh
skitt: So that the script can be run with non-default `go` without necessarily relying on the `Makefile`.

.github/workflows/multiarch.yml
```diff
@@ -0,0 +1,19 @@
...
+  pull_request:
+
+jobs:
+  apply-suggestions-commits:
+    name: Check the multi-arch builds
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out the repository
+        uses: actions/checkout@5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f
+      - name: Set up QEMU
```
mkolesnik: Can we add documentation why QEMU is needed?

scripts/shared/build_image.sh
mkolesnik: I don't believe we run these scripts outside of `make`, seems like an unnecessary complication IMHO

skitt: I run these scripts outside of `make` ;-).

.github/workflows/multiarch.yml
skitt: I’ve added a brief mention in the step name, and in the commit message (and PR description).

mkolesnik (approved):

PR #725, opened 2022-02-17
Automated backport of #718: Refactor inotify E2E GHA step to be more standard

dfarrell07:
Backport of #718 on release-0.11.

#718: Refactor inotify E2E GHA step to be more standard

For details on the backport process, see the [backport requests](https://submariner.io/development/backports/) page.

### DIFF
--- a/gh-actions/e2e/action.yaml
+++ b/gh-actions/e2e/action.yaml
@@ -56,9 +56,11 @@ runs:
         free -h
         echo "::endgroup::"
 
-    - name: Set up kernel, increase the inotify settings
-      shell: bash
-      run: sudo sysctl -w fs.inotify.max_user_watches=524288 fs.inotify.max_user_instances=512
+    - shell: bash
+      run: |
+        echo "::group::Increase inotify settings"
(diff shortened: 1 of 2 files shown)


skitt (approved):

tpantelis (approved):

dfarrell07:
What do we want the branch check job to do in this situation, where we intend to target non-devel? Should we add a "release-*" check to the GHA? Or make it not-required?

tpantelis:
> What do we want the branch check job to do in this situation, where we intend to target non-devel? Should we add a "release-*" check to the GHA? Or make it not-required?

I tried to remove the job from the branch settings but it had no effect, ie it was still listed as required and pending. That's odd.

skitt:
> > What do we want the branch check job to do in this situation, where we intend to target non-devel? Should we add a "release-*" check to the GHA? Or make it not-required?
> 
> I tried to remove the job from the branch settings but it had no effect, ie it was still listed as required and pending. That's odd.

We need to fix the job so it has the same name in all branches.

PR #726, opened 2022-02-21
Mount ~/.docker/config.json in Dapper containers

skitt:
This passes through authentication tokens, e.g. for Docker Hub or
Quay. This avoids having to try "docker login" inside the running
containers.

### DIFF
--- a/.dapper
+++ b/.dapper
@@ -100,6 +100,7 @@ fi
 docker run -i --rm --security-opt seccomp=unconfined \
        $(printf -- " -e %s" $DAPPER_ENV) -e "DAPPER_UID=$DAPPER_UID" -e "DAPPER_GID=$DAPPER_GID" \
        -v "${DAPPER_CP}:${DAPPER_SOURCE}${suffix}" \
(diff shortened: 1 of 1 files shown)


Billy99:
@skitt Pulled the change in and ran `docker info | grep "Username"` from build_image.sh and cluster.sh, the two places I found I needed to be logged in. My docker login was found in build_image.sh but not cluster.sh. So the Weave and KIND image downloads (and maybe other) in cluster.sh are still from "anonymous".
**CORRECTION:**
When running `make images`, my docker login was found in build_image.sh. However, when running `make deploy`, my docker login was NOT found in either build_image.sh or cluster.sh.
**ADDITIONAL CORRECTION:**
When running `make deploy` from submariner/submariner-operator, I remember to delete .dapper but forgot to copy over my updated copy. So it was re-pulling the upstream version. Retested with the updated .dapper file and worked fine. Sorry for the noise. I definitely like this solution better than passing the credentials in environment variables.

dfarrell07 (approved):
Very cool fix, thanks @skitt!

Billy99 (approved):

tpantelis (approved):

PR #730, opened 2022-02-23
Reinstall .shflags if it's empty

dfarrell07:
We currently download and install .shflags if it's missing from the root
of a repo that uses Shipyard. This will cause us to additionally
reinstall .shflags if the file exists but is empty and clean up
.shflags if an attempt to update it fails.

If .shflags is empty, almost all Shipyard commands fail with cryptic
errors about failing to be able to source vars.

The .shflags file does actually end up empty if something is wrong with
container networking on a dev's computer. The very first steps of all
Shipyard shared scripts include sourcing the lib that first makes these
shell features available, so it's basically the first network usage from
our shared dev env. If networking ever fails, .shflags will be
overwritten and made empty, and without this change it will keep causing
failures even after networking is fixed.

Container networking can fail while still being able to run containers.
My docker daemon sometimes fails the following test after running for a
long time. I've been seeing this bug for most of the life of Submariner,
across two laptops.

docker run --rm alpine sh -c 'wget -q -O- https://docs.docker.com \
  | grep "<title"'

Container networking starts working again after `sudo systemctl restart
docker` or restarting my computer.

Also switch some tabs to spaces for consistency. Raised #731 to track
adding linting to prevent mixed tabs/spaces in shell scripts.

### DIFF
--- a/scripts/shared/lib/shflags
+++ b/scripts/shared/lib/shflags
@@ -4,13 +4,14 @@
 
 ensure_shflags() {
     if [ -f /usr/share/shflags/shflags ]; then
-	. /usr/share/shflags/shflags
-	return
+      . /usr/share/shflags/shflags
+      return
     fi
 
-    if [ ! -f .shflags ]; then
+    if [ ! -s .shflags ]; then
         echo "Downloading shflags ${SHFLAGS_VERSION}"
         if ! curl -L "https://raw.githubusercontent.com/kward/shflags/${SHFLAGS_VERSION}/src/shflags" > .shflags; then
+            rm -f .shflags
             echo "$0 needs shflags in /usr/share/shflags/shflags or locally as .shflags." 1>&2
             echo "It was unable to download it; please install it and try again." 1>&2
             exit 1


scripts/shared/lib/shflags
```diff
@@ -8,7 +8,7 @@ ensure_shflags() {
 	return
     fi
 
-    if [ ! -f .shflags ]; then
+    if [ ! -s .shflags ]; then
         echo "Downloading shflags ${SHFLAGS_VERSION}"
         if ! curl -L "https://raw.githubusercontent.com/kward/shflags/${SHFLAGS_VERSION}/src/shflags" > .shflags; then
             echo "$0 needs shflags in /usr/share/shflags/shflags or locally as .shflags." 1>&2
```
skitt, line 15: For extra bonus points, we could delete `.shflags` if `curl` fails. The shell creates `.shflags` because of the redirection, and a failed `curl` leaves it empty (or containing nonsense).

```suggestion
            rm -f .shflags
            echo "$0 needs shflags in /usr/share/shflags/shflags or locally as .shflags." 1>&2
```

skitt (changes requested):

scripts/shared/lib/shflags
dfarrell07: Done, thanks!

scripts/shared/lib/shflags
```diff
@@ -4,13 +4,14 @@
...
-	return
+      . /usr/share/shflags/shflags
+      return
     fi
 
-    if [ ! -f .shflags ]; then
+    if [ ! -s .shflags ]; then
         echo "Downloading shflags ${SHFLAGS_VERSION}"
         if ! curl -L "https://raw.githubusercontent.com/kward/shflags/${SHFLAGS_VERSION}/src/shflags" > .shflags; then
+            rm -rf .shflags
```
skitt:
```suggestion
            rm -f .shflags
```

skitt (changes requested):
Nearly ;-).

scripts/shared/lib/shflags
dfarrell07: I debated with myself about using the GH way to accept your patch and squash it locally. Turns out the answer was "yes, do that!". :stuck_out_tongue_winking_eye:

skitt (approved):

tpantelis (approved):

PR #734, opened 2022-02-24
Update golangci-lint 1.43.0 to 1.44.2

dfarrell07:
Depends on submariner-io/shipyard#738
Depends on submariner-io/submariner#1708
Depends on submariner-io/submariner-operator#1879
Depends on submariner-io/admiral#323
Depends on submariner-io/lighthouse#695
Depends on https://github.com/submariner-io/admiral/pull/324
Depends on https://github.com/submariner-io/cloud-prepare/pull/211
Depends on https://github.com/submariner-io/cloud-prepare/pull/214
Depends on https://github.com/submariner-io/submariner-operator/pull/1892
Depends on https://github.com/submariner-io/coastguard/pull/58
Depends on https://github.com/submariner-io/lighthouse/pull/700

### DIFF
--- a/package/Dockerfile.shipyard-dapper-base
+++ b/package/Dockerfile.shipyard-dapper-base
@@ -52,7 +52,7 @@ RUN dnf -y install --nodocs --setopt=install_weak_deps=False \
     find /usr/bin /usr/lib/golang /usr/libexec -type f -executable -newercc /proc -size +1M ! -name hyperkube \( -execdir upx ${UPX_LEVEL} {} \; -o -true \) && \
     ln -f /usr/bin/kubectl /usr/bin/hyperkube
 
-ENV LINT_VERSION=v1.43.0 \
+ENV LINT_VERSION=v1.44.2 \
     HELM_VERSION=v3.8.0 \
     KIND_VERSION=v0.11.1 \
     BUILDX_VERSION=v0.7.1 \
--- a/test/e2e/framework/util.go
+++ b/test/e2e/framework/util.go
@@ -22,7 +22,6 @@ import (
 	"strings"
 
 	. "github.com/onsi/gomega"
-
 	// We need GCP authentication.
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
 	restclient "k8s.io/client-go/rest"


tpantelis (approved):

dfarrell07:
We need to fix the new issues in the projects this highlights (see Consuming Projects / Lint jobs) before we start using the new version.

skitt (approved):

sridhargaddam (approved):

nyechiel:
@dfarrell07 all dependent PRs are now merged; I think you can fix the conflict and merge. Thanks!

dfarrell07:
In addition to the "Depends on"-linked PRs, I need to figure out fixes for these errors in the operator, as mentioned at https://github.com/submariner-io/submariner-operator/pull/1879#issuecomment-1051356125 and failing in CI here.

```
pkg/subctl/cmd/diagnose/cni.go:219:17: dynamicFmtString: use errors.New(message) or fmt.Errorf("%s", message) instead (gocritic)
		return false, fmt.Errorf(message)
		              ^
pkg/subctl/cmd/diagnose/firewall.go:111:3: deferInLoop: Possible resource leak, 'defer' is called in the 'for' loop (gocritic)
		defer sPod.Delete()
		^
```

Any help with these, @skitt?

Regarding deferInLoop, I found https://stackoverflow.com/questions/45617758/defer-in-the-loop-what-will-be-better and messed around with using an anonymous function to cause the defer to happen per-loop-cycle, but the returns in the usage of the pod make it more complex it seems.

Also, should we document this "Sniffer" pod in https://submariner.io/development/security/containers/? I guess `subctl diagnose` could be used in production and run into this condition.

skitt:
> ```
> pkg/subctl/cmd/diagnose/cni.go:219:17: dynamicFmtString: use errors.New(message) or fmt.Errorf("%s", message) instead (gocritic)
> 		return false, fmt.Errorf(message)
> 		              ^
> ```

This one should be

```
		return false, fmt.Errorf("%s status not found for IPPool %q", key, pool.GetName())
```

(replacing both lines).

> ```
> pkg/subctl/cmd/diagnose/firewall.go:111:3: deferInLoop: Possible resource leak, 'defer' is called in the 'for' loop (gocritic)
> 		defer sPod.Delete()
> 		^
> ```

For this one, we could construct a slice of pods in the loop, and before the loop, `defer` a function which iterates over the slice of pods and deletes them.

dfarrell07:
I removed the `Depends on #744` because the current state of consuming project linting with golangci-lint is maximally strict, already runs in all Shipyard repos with Go files.

dfarrell07:
Note that this still creates new failures that need fixing per https://github.com/submariner-io/shipyard/pull/734#issuecomment-1056789850, and shouldn't be merged until those issues are fixed and the Consuming Lint / submariner-operator job is passing.

tpantelis:
> Note that this still creates new failures that need fixing per [#734 (comment)](https://github.com/submariner-io/shipyard/pull/734#issuecomment-1056789850), and shouldn't be merged until those issues are fixed and the Consuming Lint / submariner-operator job is passing.

Submitted https://github.com/submariner-io/submariner-operator/pull/1954

PR #739, opened 2022-02-26
Run all linter make targets in `make lint` alias

dfarrell07:
Run all linters that can be run via make in the `make lint` make target,
which is meant to be an alias to run all linters.

This will cause more linters to be run in Shipyard's consuming-lint
tests, but other CI will not change as they invoke linters directly.

### DIFF
--- a/Makefile.inc
+++ b/Makefile.inc
@@ -177,7 +177,7 @@ packagedoc-lint:
 	done 2>/dev/null; \
 	exit $$result
 
-lint: gitlint golangci-lint markdownlint yamllint
+lint: gitlint golangci-lint markdownlint packagedoc-lint shellcheck yamllint
 
 # [markdownlint] validates Markdown files in the project
 markdownlint:


mkolesnik:
I suggest merging #738 first and applying the label to see how it fares

mkolesnik (commented):
Seems it's much harder to figure out which linting failed, not directly related to this specific patch but since you`re already touching it I suggest to change the consuming job to run each linter separately instead.

That seems to be the only place where `make lint` is used, so we can then remove `make lint`, unless someone uses it out of convenience from shell.

dfarrell07:
Yeah agreed, it is hard to read that failure with all the reports smashed together at the end, would be easier if they were split per-job. I don't use the `make lint` target, and failing to keep it up-to-date is what lead to this gap in test coverage. Adding explicit per-linter tests in the Consuming workflow would be more obvious for folks to know to extend when adding linters. So yeah, I'm fine with getting rid of it.

dfarrell07:
> I don't use the make lint target

Mike asked on Slack and no said they use it, two other devs confirmed they don't use it.

https://kubernetes.slack.com/archives/C010RJV694M/p1646120917562169

dfarrell07:
> I suggest to change the consuming job to run each linter separately instead.

@mkolesnik done in https://github.com/submariner-io/shipyard/pull/744

PR #741, opened 2022-02-28
Revert "Update shflags 1.0.3 to 1.2.3"

dfarrell07:
This reverts commit 264c1e7dad11d271c60e7a8f4368076cec881334.

PR #743, opened 2022-02-28
Update shflags 1.0.3 to 1.2.3

dfarrell07:
Also update the path from which it's installed, as it has changed
between releases.


dfarrell07:
This breaks things and needs special attention, see https://github.com/submariner-io/shipyard/pull/741.

nyechiel:
@dfarrell07 @skitt is this still relevant?

skitt:
Redone with an additional required fix in #854.

PR #744, opened 2022-03-01
Disaggregate consumer linting for full coverage

dfarrell07:
For better coverage, clearer maintenance, and more readable log output.

This also removes the `make lint` target. Per discussion on #739, it
isn't used by devs and causes unclear maintenance that can result in
gaps in test coverage.

I looked at doing this by using per-project linter jobs and a custom
GitHub Action for the logic that can be extracted. Because we can't use
other GHAs in custom GHAs, we can't extract actions/checkout or would
have to create custom git logic to extract the two checkout steps.
Splitting each desired linter/project combo into a job results in lots
of duplicated GHA boilerplate. It also wastes resources re-doing things
like building the Shipyard image. Splitting each project into a parallel
job and clearly separating the linting runs in the output using steps
seems like the best solution overall.

Also reorder markdownlint make target for consistent ABC ordering.

Depends on https://github.com/submariner-io/shipyard/pull/746
Depends on https://github.com/submariner-io/submariner-charts/pull/206


.github/workflows/consuming.yml
```diff
@@ -102,8 +102,29 @@ jobs:
...
+      - name: Run golangci-lint
+        if: ${{ always() && matrix.project != 'submariner-charts' }}
+        run: make -C ${{ matrix.project }} golangci-lint
+
+      - name: Run markdownlint
+        if: always()
+        run: make -C ${{ matrix.project }} markdownlint
+
+      - name: Run packagedoc-lint
+        if: ${{ always() && matrix.project != 'cloud-prepare' && matrix.project != 'coastguard' && matrix.project != 'lighthouse' && matrix.project != 'submariner-charts' && matrix.project != 'submariner-operator' }}
```
dfarrell07: I gave up trying to wrap this, if you can get it working/wrapped I'd love to use that.

dfarrell07:
```
[22:18:18.] [linting]$ make yamllint
yamllint --strict .
./.github/workflows/consuming.yml
  118:141   error    line too long (216 > 140 characters)  (line-length)
  122:141   error    line too long (281 > 140 characters)  (line-length)
```

Seems like the linter is going to insist, keep us honest.

.github/workflows/consuming.yml
```diff
@@ -102,8 +102,47 @@ jobs:
...
 
-      - name: Run all linters
-        run: make -C ${{ matrix.project }} lint
+      - name: Run gitlint
+        if: always()
+        run: make -C ${{ matrix.project }} gitlint
+
+      - name: Run golangci-lint
+        if: ${{ always() && matrix.project != 'submariner-charts' }}
+        run: make -C ${{ matrix.project }} golangci-lint
```
skitt, line 109: We should make the linting targets exit without error if there’s nothing to do, instead of hard-coding lists of exceptions here.

.github/workflows/consuming.yml
```diff
@@ -102,8 +102,47 @@ jobs:
       - name: Make sure ${{ matrix.project }} is using the built Shipyard image
         run: sed -i 's/shipyard-dapper-base:*.*/shipyard-dapper-base:dev/' ${{ matrix.project }}/Dockerfile.dapper
 
-      - name: Run all linters
-        run: make -C ${{ matrix.project }} lint
+      - name: Run gitlint
+        if: always()
```
skitt, line 104: Do we need this?

skitt (changes requested):

.github/workflows/consuming.yml
```diff
@@ -102,8 +102,47 @@ jobs:
...
+
+      - name: Run packagedoc-lint
+        if: >-
+          ${{
+          always()
+          && matrix.project != 'cloud-prepare'
+          && matrix.project != 'coastguard'
+          && matrix.project != 'lighthouse'
+          && matrix.project != 'submariner-charts'
+          && matrix.project != 'submariner-operator'
```
skitt: `packagedoc-lint` has no effect on `submariner-charts`; the rest should be fixed.

.github/workflows/consuming.yml
dfarrell07: There's an implicit `success()` if you don't override it with `always()`, if that's what you mean. Maybe I should document that inline? Or at least in the commit message.

In one sense we don't need it for this particular linter because it's the first to run, so nothing should fail before it that we want to continue running in spite of. That would maybe be a bit brittle to maintain though, if someone adds a new linter that comes first in ABC order but doesn't add an `if: always()` to the now-second linter, it would be a bug.

skitt: That was indeed the information I was after, thanks!

.github/workflows/consuming.yml
```diff
@@ -102,8 +102,47 @@ jobs:
...
+        if: >-
+          ${{
+          always()
+          && matrix.project != 'admiral'
+          && matrix.project != 'cloud-prepare'
+          && matrix.project != 'coastguard'
+          && matrix.project != 'lighthouse'
+          && matrix.project != 'submariner'
+          && matrix.project != 'submariner-charts'
+          && matrix.project != 'submariner-operator'
```
mkolesnik: This is almost all the projects, isnt it easier to have a whitelist?

.github/workflows/consuming.yml
```diff
@@ -102,8 +102,47 @@ jobs:
...
         run: sed -i 's/shipyard-dapper-base:*.*/shipyard-dapper-base:dev/' ${{ matrix.project }}/Dockerfile.dapper
 
-      - name: Run all linters
-        run: make -C ${{ matrix.project }} lint
+      - name: Run gitlint
+        if: always()
+        run: make -C ${{ matrix.project }} gitlint
+
+      - name: Run golangci-lint
+        if: ${{ always() && matrix.project != 'submariner-charts' }}
```
mkolesnik: Is always() necessary in this context? I'd imagine it isnt.

.github/workflows/consuming.yml
```diff
@@ -102,8 +102,47 @@ jobs:
...
+        run: make -C ${{ matrix.project }} golangci-lint
+
+      - name: Run markdownlint
+        if: always()
+        run: make -C ${{ matrix.project }} markdownlint
+
+      - name: Run packagedoc-lint
+        if: >-
+          ${{
```
mkolesnik: You don't have to specify `${{` in the `if:` AFAIK

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idif

mkolesnik (changes requested):

.github/workflows/consuming.yml
skitt: It’s easier to not need a list of projects at all, just run the targets everywhere, and change the targets so that they don’t do anything if they’re not supposed to.

.github/workflows/consuming.yml
dfarrell07: That's already answered here I think: https://github.com/submariner-io/shipyard/pull/744#discussion_r817890307

.github/workflows/consuming.yml
dfarrell07: Yep, seems true. https://github.com/dfarrell07/shipyard/pull/38

dfarrell07:
This is ready for re-review I think.

skitt (approved):

dfarrell07:
@mkolesnik Can you re-review this one?

.github/workflows/consuming.yml
```diff
@@ -100,8 +100,29 @@ jobs:
       - name: Make sure ${{ matrix.project }} is using the built Shipyard image
         run: sed -i 's/shipyard-dapper-base:*.*/shipyard-dapper-base:dev/' ${{ matrix.project }}/Dockerfile.dapper
 
-      - name: Run all linters
-        run: make -C ${{ matrix.project }} lint
+      - name: Run gitlint
+        if: always()
+        run: make -C ${{ matrix.project }} gitlint
```
mkolesnik, line 105: Since they all basically perform the same action now `make -C ${{ matrix.project }} <linter>` can't we just add to the matrix: `linter: ['gitlint', 'golangci-lint', 'markdownlint', 'packagedoc-lint', 'shellcheck', 'yamllint']`?

Then we can simply call `make -C ${{ matrix.project }} ${{ matrix.linter }}` :)

This also avoids the need for awkward `if: always()` ;)

mkolesnik (changes requested):

.github/workflows/consuming.yml
mkolesnik: You know what I thought about it some more and adding it to the matrix is much nicer and cleaner BUT it would mean we would rebuild the shipyard image and insane amount of times so maybe its best to leave as is for sake of less long running jobs.

WDYT?

dfarrell07: Yeah agreed, I noted something like that in the commit message:
 
> Splitting each desired linter/project combo into a job...wastes resources re-doing things like building the Shipyard image.

mkolesnik (approved):

PR #746, opened 2022-03-02
Make golangci-lint a no-op if unconfigured

skitt:
golangci-lint only make sense in projects with Go files; attempting to
run it on projects without any Go content causes errors. This turns
"make golangci-lint" into a no-op if there are no Go files to process.

### DIFF
--- a/Makefile.inc
+++ b/Makefile.inc
@@ -165,8 +165,12 @@ VENDOR_MODULES := $(shell find . -name vendor -prune -o -name go.mod -printf "%h
 
 # [golangci-lint] validates Go code in the project
 golangci-lint: $(VENDOR_MODULES)
+ifneq (,$(shell find . -name '*.go'))
 	golangci-lint linters
 	golangci-lint run --timeout 10m
+else
+	@echo 'There are no Go files to lint.'
+endif
 
 # [packagedoc-lint] checks that the package docs don’t include the SPDX header
 packagedoc-lint:


dfarrell07 (approved):
```
[~/go/src/submariner-io/submariner-charts]$ [ ! -r .golangci.yml ] || golangci-lint linters
[~/go/src/submariner-io/submariner-charts]$ echo $?
0
```

mkolesnik (changes requested):
What's the purpose of this change?

`golangci-lint` can run perfectly fine in projects that don't have the config file.

For example, I deleted the file and ran it:
```bash
golangci-lint run --timeout 10m
Successfully remade target file 'golangci-lint'.
[13:10:45.618] [cloud-prepare]$ make --debug=b golangci-lint
[mkolesni@mkolaptop cloud-prepare (devel *%=)]$ git st
On branch devel
Your branch is up to date with 'origin/devel'.

Changes not staged for commit:
  (use "git add/rm <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	deleted:    .golangci.yml
```

skitt:
> What's the purpose of this change?

Thanks, the purpose wasn’t very clear in my commit message. Hopefully the updated commit message and PR description explain things better!

mkolesnik:
> > What's the purpose of this change?
> 
> Thanks, the purpose wasn’t very clear in my commit message. Hopefully the updated commit message and PR description explain things better!

Thanks for clarifying, I think a better approach is to `find` go files and if none found skip the action. This should be more robust as the file seems not to be necessary in all the projects, and if its removed we don't want this linting to generate false negatives.

Makefile.inc
```diff
@@ -165,8 +165,13 @@ VENDOR_MODULES := $(shell find . -name vendor -prune -o -name go.mod -printf "%h
 
 # [golangci-lint] validates Go code in the project
 golangci-lint: $(VENDOR_MODULES)
+ifneq (,$(shell find . -name '*.go'))
```
mkolesnik, line 168: I guess you could've used `wildcard` here as well?

mkolesnik (approved):

Makefile.inc
```diff
@@ -165,8 +165,13 @@ VENDOR_MODULES := $(shell find . -name vendor -prune -o -name go.mod -printf "%h
...
+	echo $(wildcard **/*.go)
 	golangci-lint linters
 	golangci-lint run --timeout 10m
+else
+	@echo 'There are no Go files to lint; if this is incorrect, write some Go.'
```
dfarrell07:
```suggestion
	@echo 'There are no Go files to lint.'
```

If you end up pushing another revision, I'd get rid of this.

Makefile.inc
skitt: `wildcard` doesn’t match inside subdirectories, and we have a number of Go projects with no `*.go` files in their root directory. (The `echo` underneath was a remnant of my tests...)

PR #747, opened 2022-03-02
Make shellcheck a no-op if unconfigured

skitt:
Only run shellcheck if SHELLCHECK_ARGS (i.e., the list of files to
check) is set.

### DIFF
--- a/Makefile.inc
+++ b/Makefile.inc
@@ -191,7 +191,12 @@ markdownlint:
 
 # [shellcheck] validates your shell files
 shellcheck:
+# Only run shellcheck if there are files to check
+ifneq (,$(SHELLCHECK_ARGS))
 	shellcheck -x -P $${SCRIPTS_DIR} $(SHELLCHECK_ARGS)
+else
+	@echo 'There are no shell scripts to check; if this is incorrect, specify them in SHELLCHECK_ARGS.'
+endif
 
 # [yamllint] validates YAML files in the project
 yamllint:


dfarrell07 (approved):
Patching /opt/shipyard/Makefile.inc in a `make shell` with the same:

```
[root@5487338d2c20 submariner-charts]# make shellcheck
/opt/shipyard/Makefile.inc:7: Your Makefile.dapper and Shipyard's are different,
/opt/shipyard/Makefile.inc:8: you might run into issues. To avoid this, ensure
/opt/shipyard/Makefile.inc:9: your Shipyard image is up-to-date and delete the
/opt/shipyard/Makefile.inc:10: local Makefile.dapper.
[ -z "" ] || shellcheck -x -P ${SCRIPTS_DIR} 
[root@5487338d2c20 submariner-charts]# echo $?
0
```

Makefile.inc
```diff
@@ -189,7 +189,7 @@ markdownlint:
 
 # [shellcheck] validates your shell files
 shellcheck:
-	shellcheck -x -P $${SCRIPTS_DIR} $(SHELLCHECK_ARGS)
+	[ -z "$(SHELLCHECK_ARGS)" ] || shellcheck -x -P $${SCRIPTS_DIR} $(SHELLCHECK_ARGS)
```
mkolesnik: Would it be better to use `ifdef` ?

Makefile.inc
```diff
@@ -191,7 +191,10 @@ markdownlint:
 
 # [shellcheck] validates your shell files
 shellcheck:
+# Only run shellcheck if there are files to check
+ifneq (,$(SHELLCHECK_ARGS))
 	shellcheck -x -P $${SCRIPTS_DIR} $(SHELLCHECK_ARGS)
+endif
```
mkolesnik, line 199: Probably should print a warning to stdout clarifying it was skipped because the ARGS is not defined, otherwise it might be confusing

mkolesnik (changes requested):

Makefile.inc
skitt: Isn’t

    make: Nothing to be done for 'shellcheck'.

clear enough?

mkolesnik: That's `make` implicit, explicit is better than implicit.

Not everyone is a make guru or understands the intricacies.. Besides if someone did run `make shellcheck` on a project not supporting it I'd expect a clear message such as `Shellcheck is not configured to run on this project` or something like that

skitt: If we start going down that path, we’ll need to add messages for every similar situation to be consistent. *E.g.*

    $ make bin/subctl
    make: 'bin/subctl' is up to date.

etc. Or am I misunderstanding?

skitt: I don’t agree that one needs to be a Make guru to understand what “Nothing to be done” means, but whatever, I’ve added a message. I’ve updated the `golangci-lint` patch along similar lines, and I’ll submit other patches to do this on other targets where necessary.

dfarrell07: > make: Nothing to be done for 'shellcheck'.

This is more clear to me, as I'm accustomed to seeing it from Make and I don't have to think about `if this is incorrect, specify them in SHELLCHECK_ARGS`.

Either way is fine with me.

mkolesnik (approved):

PR #751, opened 2022-03-08
Include cloud-prepare in `make prune-images`

dfarrell07:
Delete cloud-prepare container images in the helper command meant to
delete all Submariner container images. Cloud prepare does produce
images (demo: `make shell`/`docker images`), so this fixes the ~bug of
the leftover images.

### DIFF
--- a/Makefile.dapper
+++ b/Makefile.dapper
@@ -57,7 +57,7 @@ $(LINTING_GOALS): .dapper $(LINTING_DAPPER)
 # Copied from Makefile to provide this everywhere (until we can share
 # non-Dapper goals across projects)
 prune-images:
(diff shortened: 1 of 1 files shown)


skitt (approved):

mkolesnik (approved):
The solution is correct though the explanation isn't :)

The "image" is created by running dapper and defaults (AFAICT) to the name of the folder, hence "cloud-prepare" images appear.

Since we're using our own "dapper" now perhaps we can change that convention to something that doesn't require us to keep track of the folder names in this command?

dfarrell07:
> The "image" is created by running dapper and defaults (AFAICT) to the name of the folder, hence "cloud-prepare" images appear.

Yeah I was wondering about that, didn't expect their would be an image until I noticed it dangling. Thanks for the explanation Mike!

PR #216, opened 2020-06-20
remove default reporter and file path since ginkgo creates them

pinikomarov:
Signed-off-by: pinikomarov <pkomarov@redhat.com>

PR #755, opened 2022-03-17
Add JUnit report generation to "subctl verify"

MaxBab:
The E2E tests should be able to generate report at the end of the
execution.
After the following ginkgo patch - [1], creation of the test directory
(if required) and generation of the report handled by ginkgo after the
input parameters such as directory and file name.

Replacing existing parameters which are no longer working after the
following patch - [2].

[1] - https://github.com/onsi/ginkgo/pull/601
[2] - https://github.com/submariner-io/shipyard/pull/216

Depends On: https://github.com/submariner-io/submariner-operator/pull/1961

### DIFF
--- a/test/e2e/e2e.go
+++ b/test/e2e/e2e.go
@@ -91,5 +91,10 @@ func RunE2ETests(t *testing.T) bool {
 		config.DefaultReporterConfig.SlowSpecThreshold = 45.0
 	}
 
+	// Set junit report path and filename if required by user
+	// The default of the parameter is set to empty, so if not used
+	// creation of report will be skipped
+	config.DefaultReporterConfig.ReportFile = framework.TestContext.JunitReport
+
 	return ginkgo.RunSpecs(t, "Submariner E2E suite")
 }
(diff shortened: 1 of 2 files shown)


test/e2e/e2e.go
```diff
@@ -91,5 +91,11 @@ func RunE2ETests(t *testing.T) bool {
 		config.DefaultReporterConfig.SlowSpecThreshold = 45.0
 	}
 
+	// Set junit report path and filename if required by user
+	junitReport := framework.TestContext.JunitReport
+	if junitReport != "" {
+		config.DefaultReporterConfig.ReportFile = junitReport
+	}
```
skitt: We could skip the additional variable; but we can even skip the conditional entirely, since if `JUnitReport` is empty, we just set `ReportFile` to empty:

```suggestion
	config.DefaultReporterConfig.ReportFile = framework.TestContext.JunitReport
```

skitt (changes requested):

test/e2e/e2e.go
MaxBab: That's a good point.
Thanks.

skitt (approved):

dfarrell07 (approved):

PR #761, opened 2022-03-23
Allow comments in CODEOWNERS.in

skitt:
This is being used (currently incorrectly) in feature branches; it is
useful to document different scopes of code ownership. This isn't a
concept as such in our documented processes, but we want to keep
"core" code owners alongside feature branch code owners, and it is
helpful to document which is which.


tpantelis (approved):

sridhargaddam (approved):
Thanks for fixing @skitt

PR #773, opened 2022-04-04
Only use buildx for multi-arch image builds

### LINKED ISSUE #771: Multi arch images release broke the CI
I noticed CI started failing with this error:
```
  error: failed to solve: failed to get reader: content digest sha256:40e059520d199e1a1a259089077f2a0c879951c9a4540490bad3a0d7714c6ae7: not found
```

This seems to be related to the latest merge which published multi arch images.

We're either not consuming them correctly or not publishing them correctly (e.g. should have a default platform?).

For example, see https://github.com/submariner-io/shipyard/runs/5805414807?check_suite_focus=true

skitt:
We are seeing issues with buildx when building a single-arch image
using a multi-arch manifest. This changes our image build script to
only use buildx when building a multi-arch image.

Fixes: #771

### DIFF
--- a/scripts/shared/build_image.sh
+++ b/scripts/shared/build_image.sh
@@ -76,7 +76,7 @@ fi
 # Rebuild the image to update any changed layers and tag it back so it will be used.
 buildargs_flag="--build-arg BUILDKIT_INLINE_CACHE=1 --build-arg BASE_BRANCH=${BASE_BRANCH}"
 [[ -z "${buildargs}" ]] || buildargs_flag="${buildargs_flag} --build-arg ${buildargs}"
-if docker buildx version > /dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


mkolesnik (approved):
Kind of weird, should we track it long term?

We were using buildx before the multi-arch builds and it worked quite well

skitt:
> Kind of weird, should we track it long term?
> 
> We were using buildx before the multi-arch builds and it worked quite well

I think we’re running into an edge-case: the `buildx` messages suggest it thinks it’s doing an OCI build, despite the `--load` setting. The failure mode is very strange since the missing object is always the first layer that was pulled, so it really is present.

I’m not sure it’s worth spending the time to investigate why `buildx` is failing here. I’d switched to it a long time ago as a pre-requisite for multi-arch builds, I think it’s in keeping with that to restrict it to multi-arch builds.

dfarrell07 (approved):

PR #774, opened 2022-04-05
Don't release multi-arch images if there are none

skitt:
release-images always tries to release multi-arch images, even when
there aren't any; see
https://github.com/submariner-io/submariner-operator/runs/5831545335?check_suite_focus=true
for one instance. This doesn't break the release of existing images
but causes the job to report failure.


dfarrell07 (approved):

dfarrell07:
@tpantelis We briefly talked about a failing job involving an image push, I was thinking about this as the possible root cause.

tpantelis (approved):

PR #775, opened 2022-04-08
Bump to Fedora 36

skitt:
Fedora 36 isn't quite released yet, but it provides Go 1.18, and we
need Go 1.17 for current Kubernetes dependencies (see
https://github.com/submariner-io/lighthouse/pull/658 for example).

(Fedora 35 has Go 1.16.)

Depends on #776
Depends on #802 
Depends on #803
Depends on https://github.com/submariner-io/submariner-operator/pull/2016

### DIFF
--- a/package/Dockerfile.shipyard-dapper-base
+++ b/package/Dockerfile.shipyard-dapper-base
@@ -1,4 +1,4 @@
-FROM fedora:35
+FROM fedora:36
 
 # Unless specified otherwise, compress to a medium level which gives (from experemintation) a
 # good balance between compression time and resulting image size.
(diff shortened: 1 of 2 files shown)


mkolesnik (approved):
Generally makes sense though I wonder if it's not better to wait for official F36 release?

skitt:
> Generally makes sense though I wonder if it's not better to wait for official F36 release?

The packages we care about are unlikely to change much before the F36 release at this point.

dfarrell07 (approved):

package/Dockerfile.shipyard-dapper-base
```diff
@@ -1,4 +1,4 @@
-FROM fedora:35
+FROM fedora:36
```
dfarrell07, line 1: (tl;dr this works both pre and post-release)
It seems like this tag is maintained at the latest build from the 36 branch, which is currently pre-release but will be updated to 36 once it's released, and will continue to be updated after release (based on 34/35 tags I see) 


https://quay.io/repository/fedora/fedora?tab=tags

PR #781, opened 2022-04-21
Fix Nginx Image

astoycos:
quay.io/bitnami/nginx:latest was removed use
quay.io/testing-farm/nginx instead since it is
just a quay.io mirror of the official docker
image.

We verified the SHAs of the containers here match
the official ones

https://quay.io/repository/testing-farm/nginx/manifest/sha256:61face6bf030edce7ef6d7dd66fe452298d6f5f7ce032afdd01683ef02b2b841

https://hub.docker.com/layers/nginx/library/nginx/latest/images/sha256-61face6bf030edce7ef6d7dd66fe452298d6f5f7ce032afdd01683ef02b2b841

Also https://gitlab.com/testing-farm/docs/root seems
to be CopyWrited/Owned by Redhat.

The new image exposes port 80 instead of 8080,
so update the `nginx-demo` service+deployment
to also reflect this.

PR #782, opened 2022-04-21
Use Quay mirror of official nginx container images

dfarrell07:
The current nginx container images from bitnami are failing to pull.
They seem to no longer be provided as a part of bitnami changes.

This Quay nginx source seems to automatically mirror the official nginx
builds on DockerHub. Just like us, they want to avoid the rate limiting.

I verified the SHAs of the containers here match the official ones

https://quay.io/repository/testing-farm/nginx/manifest/sha256:61face6bf030edce7ef6d7dd66fe452298d6f5f7ce032afdd01683ef02b2b841

https://hub.docker.com/layers/nginx/library/nginx/latest/images/sha256-61face6bf030edce7ef6d7dd66fe452298d6f5f7ce032afdd01683ef02b2b841

### DIFF
--- a/scripts/shared/resources/nginx-demo.yaml
+++ b/scripts/shared/resources/nginx-demo.yaml
@@ -15,7 +15,7 @@ spec:
     spec:
       containers:
         - name: nginx-demo
-          image: quay.io/bitnami/nginx:latest
+          image: quay.io/testing-farm/nginx:latest
           ports:
             - containerPort: 8080
 ---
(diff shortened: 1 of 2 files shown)


dfarrell07:
We might need to re-work the ports here, the container seems to actually use 80 where we expect 8080. I think @astoycos is sending a PR with that.

astoycos:
Yeah I like this better since it's just a total mirror @dfarrell07, I wonder who "owns" this though All I could find on git was https://github.com/testing-farm

dfarrell07:
This has been incorporated in to https://github.com/submariner-io/shipyard/pull/781

PR #786, opened 2022-04-26
Use openshifttest/nginx-alpine for rootless E2E

dfarrell07:
To fix support for running E2E tests using an nginx web server container
in OpenShift, which only allows running rootless containers.

The quay.io/bitnami/nginx image we used for a long time was rootless but
has been removed. The quay.io/testing-farm/nginx mirror of the official
nginx containers on DockerHub is not rootless. The official rootless
container nginxinc/nginx-unprivileged is on DockerHub but has no mirror
to Quay. The agnhost K8s CI tool's netexec or nettest commands might be
useful, but would require more retooling.

### DIFF
--- a/scripts/shared/resources/nginx-demo.yaml
+++ b/scripts/shared/resources/nginx-demo.yaml
@@ -15,9 +15,9 @@ spec:
     spec:
       containers:
         - name: nginx-demo
-          image: quay.io/testing-farm/nginx:latest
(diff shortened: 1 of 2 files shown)


dfarrell07:
I tried this with port 80 here, same results: https://github.com/dfarrell07/shipyard/pull/40

dfarrell07:
Trying to do this with https://github.com/submariner-io/shipyard/pull/787 instead.

PR #787, opened 2022-04-26
Add `simpleserver` for testing

mkolesnik:
Instead of relying on an external webserver, we can use `nc` in a simple
wrapper script to serve a simple endpoint for our tests.

This should avoid future breakage since we're shipping it in our own
image, and it's a very simple image with simple requirements (rootless).

### DIFF
--- a/scripts/nettest/simpleserver
+++ b/scripts/nettest/simpleserver
@@ -1,0 +1,7 @@
+#!/bin/bash
+set -e
+
+while true
+do
+    echo -e "HTTP/1.1 200 OK\r\n\r\nHello World" | nc -l -p 8080
+done
--- a/package/Dockerfile.nettest
+++ b/package/Dockerfile.nettest
@@ -24,5 +24,6 @@ RUN apk add --no-cache \
 	tcpdump
 
 COPY --from=0 /usr/local/bin/net* /usr/local/bin/
+COPY scripts/nettest/* /app/
 
 CMD ["/bin/bash","-l"]
--- a/scripts/shared/lib/deploy_funcs
+++ b/scripts/shared/lib/deploy_funcs
@@ -75,7 +75,7 @@ function test_connection() {
 
 function connectivity_tests() {
     target_cluster="$1"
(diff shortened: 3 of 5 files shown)


scripts/nettest/simpleserver
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+set -e
+
+while true
+do
+    echo -e "HTTP/1.1 200 OK\r\n\r\nHello World" | nc -l -p 8080
```
dfarrell07, line 6: This will peg a thread at 100%, right? Is that okay, or should we use some small delay in the loop or something?

dfarrell07 (approved):

scripts/nettest/simpleserver
dfarrell07: I think we should go ahead and merge this to un-break rootless E2E, but still wondering if this should be edited.

dfarrell07:
I think this should be backported to fix `release-0.12`.

scripts/nettest/simpleserver
tpantelis: I tried it locally and don't see it taking much CPU. I think it just waits for a connection on the port.

tpantelis (approved):

scripts/nettest/simpleserver
mkolesnik: `nc` will wait for a connection which won't take much CPU resources.
Once it gets a connection it responds and exits, so the loop just makes sure it is re-run again (and listens again)

mkolesnik:
Upgrade was failing because it was pulling `nettest:latest` which hasn't been tagged for quite a while now. This was fine before since we didn't need anything new in the image, but it makes more sense to pull the base image anyhow instead of `latest`.

scripts/nettest/simpleserver
dfarrell07: Nice!

mkolesnik:
Backported to 0.12 as #792

PR #789, opened 2022-04-28
Make sure to pull `devel` nettest image

mkolesnik:
The tests are pulling `latest` which is actually tagged but is very
old and unmaintained.

### DIFF
--- a/test/e2e/framework/deployments.go
+++ b/test/e2e/framework/deployments.go
@@ -66,7 +66,7 @@ func (f *Framework) NewNetShootDeployment(cluster ClusterIndex) *corev1.PodList 
 					Containers: []corev1.Container{
 						{
 							Name:            "netshoot",
-							Image:           "quay.io/submariner/nettest",
+							Image:           "quay.io/submariner/nettest:devel",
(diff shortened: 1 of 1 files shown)


mkolesnik:
There is a caveat that `nettest` is an internal image, and one could argue that on stable branches we would like to pull the image for that branch rather than the `devel` one, however this isn't much different than deploying any 3rd party image (as we used to deploy `nginx`) from `latest` in stable branch.

Therefore, it seems fine to me to deploy `devel` for these purposes, but we can of course discuss this if it doesn't make sense.

mkolesnik:
I ran LH E2E locally this change and all the tests pass

vthapar (approved):

mkolesnik:
An alternative solution (which I actually prefer) is to move these functions to LH as they're onlyh used there, and remove it from Shipyard entirely.

If we're OK with moving them and merge https://github.com/submariner-io/lighthouse/pull/738 I'll send a PR to remove them from here.

mkolesnik:
We went with https://github.com/submariner-io/lighthouse/pull/738

PR #802, opened 2022-05-03
Mark the source directory as safe

skitt:
See
https://github.blog/2022-04-12-git-security-vulnerability-announced/
for context. git now refuses to handle repositories which don't belong
to the current user by default; such repositories need to be
explicitly marked as safe, in the global configuration for the current
user.


skitt:
This restores the “Commit Message(s)” job, and prevents future failures when we upgrade `git` (in the Fedora bump, see #775).

tpantelis (approved):

dfarrell07 (approved):

PR #807, opened 2022-05-05
Check that the multiarch image builds work

skitt:
make multiarch-images can succeed without actually producing
multiarch-images; it warns about this but doesn't fail. This adds a
check that we actually produced multiarch images.

### DIFF
--- a/.github/workflows/multiarch.yml
+++ b/.github/workflows/multiarch.yml
@@ -5,7 +5,7 @@ on:
   pull_request:
 
 jobs:
-  apply-suggestions-commits:
+  check-multiarch:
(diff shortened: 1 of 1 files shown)


tpantelis (approved):

.github/workflows/multiarch.yml
```diff
@@ -5,7 +5,7 @@ on:
   pull_request:
 
 jobs:
-  apply-suggestions-commits:
+  check-multiarch:
```
dfarrell07, line 8: Good catch

dfarrell07 (approved):

PR #813, opened 2022-05-12
Disable VCS stamping in go compile

dfarrell07:
Local builds in Lighthouse (at least) fail with:

$ make deploy using=globalnet
...
[16:19:11.121] [lighthouse]$ CGO_ENABLED=0 go build -trimpath -ldflags
  -s -w  -o bin/lighthouse-agent ./pkg/agent
error obtaining VCS status: exit status 128
	Use -buildvcs=false to disable VCS stamping.
make: *** [Makefile:30: bin/lighthouse-agent] Error 1

We have go 1.18 in the dev container (after make shell in LH):

[root@5dd87becc9c2 lighthouse]# go version
go version go1.18 linux/amd64

But use go 1.17 in CI, which seems like the relevant difference issue.

kind v0.12.0 go1.17.8 linux/amd64

### DIFF
--- a/scripts/shared/compile.sh
+++ b/scripts/shared/compile.sh
@@ -34,6 +34,6 @@ if [ "$build_debug" = "false" ]; then
     ldflags="-s -w ${ldflags}"
 fi
 
-CGO_ENABLED=0 ${GO:-go} build -trimpath -ldflags "${ldflags}" -o $binary $source_file
+CGO_ENABLED=0 ${GO:-go} build -buildvcs=false -trimpath -ldflags "${ldflags}" -o $binary $source_file
 [[ "$build_upx" = "false" ]] || [[ "$build_debug" = "true" ]] || upx $binary
 


dfarrell07:
Probably @skitt should validate this before merging, I'm not very confident that this is the ultimately-correct fix.

dfarrell07:
Running `make images` with this change in Shipyard did result in my LH compile picking up the new vcs=false flag, and the resulting compile/deploy did pass.

skitt:
> We have go 1.18 in the dev container (after make shell in LH):
> 
> [root@5dd87becc9c2 lighthouse]# go version go version go1.18 linux/amd64
> 
> But use go 1.17 in CI, which seems like the relevant difference issue.
> 
> kind v0.12.0 go1.17.8 linux/amd64

The latter only indicates the version used to build kind; we download kind binaries, we don’t rebuild it ourselves.

skitt:
The correct fix is to delete your local `Dockerfile.dapper` in your `lighthouse` repo.

PR #814, opened 2022-05-12
make deploy fails to provision Weave images

### LINKED ISSUE #769: make deploy fails to provision Weave images if they are locally tagged
When deploying with `make deploy`, Weave fails to start because the images aren’t pushed to the local registry if they already exist with conflicting names and tags (_e.g._ `localhost:5000/weave-kube:2.8.1`).

It might not be worth fixing this; but it might explain issues various developers have run into over the past few weeks, and if we don’t fix it we should document it. We could fix it by untagging the images before starting the local registry...

Billy99:
Resolves: #769

When deploying with make deploy, Weave fails to start because the images
aren’t pushed to the local registry if they already exist with conflicting
names and tags (e.g. localhost:5000/weave-kube:2.8.1).

Add code to run_local_registry() to push images to local registry if the
registry is stored in memory and the local volume mount directory is empty.

I was able to reproduce by rebooting the host. When the host comes back
up (a VM in my testing), the local registry is empty, but the `localhost:5000`
images are still tagged.

Signed-off-by: Billy McFall <22157057+Billy99@users.noreply.github.com>

### DIFF
--- a/scripts/shared/clusters.sh
+++ b/scripts/shared/clusters.sh
@@ -252,11 +252,26 @@ function run_local_registry() {
         echo "Deploying local registry $KIND_REGISTRY to serve images centrally."
         local volume_flag
         if [[ $registry_inmemory = true ]]; then
-            volume_flag="-v /dev/shm/${KIND_REGISTRY}:/var/lib/registry"
+            volume_dir="/var/lib/registry"
+            volume_flag="-v /dev/shm/${KIND_REGISTRY}:${volume_dir}"
             selinuxenabled && volume_flag="${volume_flag}:z" 2>/dev/null
         fi
         docker run -d $volume_flag -p 127.0.0.1:5000:5000 --restart=always --name $KIND_REGISTRY registry:2
         docker network connect kind $KIND_REGISTRY || true
+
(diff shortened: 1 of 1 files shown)


Billy99:
@skitt Let me know if you think this is a valid alternative. In #769 you suggested "untagging the images before starting the local registry." The problem with that approach is that local Submariner images are not copied and there is nothing in the scripts that would repair it. This approach gets the Submariner images as well as the Weave images copied to the local registry.

The only scenario that still doesn't work is if a set of clusters is up and running with images in the local registry (KIND clusters in my testing) and the host reboots. When the host comes back up, the KIND cluster is restarted, but the local registry is empty. But I was having trouble with certs and accessing the cluster so I think there are other issues with that scenario.

skitt (approved):

skitt:
> @skitt Let me know if you think this is a valid alternative. In #769 you suggested "untagging the images before starting the local registry." The problem with that approach is that local Submariner images are not copied and there is nothing in the scripts that would repair it. This approach gets the Submariner images as well as the Weave images copied to the local registry.

Thanks, this does seem like a better approach!

tpantelis (approved):

PR #825, opened 2022-05-18
Bump to kind 0.14.0

skitt:
This allows us to test against the current release of Kubernetes,
1.24. Support for 1.17 is no longer provided by default with kind, so
the corresponding entry is removed.

See https://github.com/kubernetes-sigs/kind/releases/tag/v0.13.0 and
https://github.com/kubernetes-sigs/kind/releases/tag/v0.14.0 for
details.

Depends on https://github.com/kubernetes-sigs/kind/issues/2803
Depends on https://github.com/submariner-io/subctl/issues/102

### DIFF
--- a/scripts/shared/clusters.sh
+++ b/scripts/shared/clusters.sh
@@ -2,7 +2,7 @@
 
 ## Kubernetes version mapping, as supported by kind ##
 # See the release notes of the kind version in use
-DEFAULT_K8S_VERSION=1.23
+DEFAULT_K8S_VERSION=1.24
 declare -A kind_k8s_versions
 kind_k8s_versions[1.18]=1.18.20@sha256:738cdc23ed4be6cc0b7ea277a2ebcc454c8373d7d8fb991a7fcdbd126188e6d7
 kind_k8s_versions[1.19]=1.19.16@sha256:d9c819e8668de8d5030708e484a9fdff44d95ec4675d136ef0a0a584e587f65c
--- a/scripts/shared/clusters.sh
+++ b/scripts/shared/clusters.sh
@@ -4,13 +4,13 @@
 # See the release notes of the kind version in use
 DEFAULT_K8S_VERSION=1.23
 declare -A kind_k8s_versions
-kind_k8s_versions[1.17]=1.17.17@sha256:e477ee64df5731aa4ef4deabbafc34e8d9a686b49178f726563598344a3898d5
-kind_k8s_versions[1.18]=1.18.20@sha256:e3dca5e16116d11363e31639640042a9b1bd2c90f85717a7fc66be34089a8169
-kind_k8s_versions[1.19]=1.19.16@sha256:81f552397c1e6c1f293f967ecb1344d8857613fb978f963c30e907c32f598467
-kind_k8s_versions[1.20]=1.20.15@sha256:393bb9096c6c4d723bb17bceb0896407d7db581532d11ea2839c80b28e5d8deb
-kind_k8s_versions[1.21]=1.21.10@sha256:84709f09756ba4f863769bdcabe5edafc2ada72d3c8c44d6515fc581b66b029c
-kind_k8s_versions[1.22]=1.22.7@sha256:1dfd72d193bf7da64765fd2f2898f78663b9ba366c2aa74be1fd7498a1873166
-kind_k8s_versions[1.23]=1.23.4@sha256:0e34f0d0fd448aa2f2819cfd74e99fe5793a6e4938b328f657c8e3f81ee0dfb9
+kind_k8s_versions[1.18]=1.18.20@sha256:738cdc23ed4be6cc0b7ea277a2ebcc454c8373d7d8fb991a7fcdbd126188e6d7
+kind_k8s_versions[1.19]=1.19.16@sha256:d9c819e8668de8d5030708e484a9fdff44d95ec4675d136ef0a0a584e587f65c
+kind_k8s_versions[1.20]=1.20.15@sha256:6f2d011dffe182bad80b85f6c00e8ca9d86b5b8922cdf433d53575c4c5212248
+kind_k8s_versions[1.21]=1.21.12@sha256:f316b33dd88f8196379f38feb80545ef3ed44d9197dca1bfd48bcb1583210207
+kind_k8s_versions[1.22]=1.22.9@sha256:8135260b959dfe320206eb36b3aeda9cffcb262f4b44cda6b33f7bb73f453105
+kind_k8s_versions[1.23]=1.23.6@sha256:b1fa224cc6c7ff32455e0b1fd9cbfd3d3bc87ecaa8fcb06961ed1afb3db0f9ae
+kind_k8s_versions[1.24]=1.24.2@sha256:a3220cefdf4f9be6681c871da35521eaaf59fadd7d509613a9e1881c5f74b587
 
 ## Process command line flags ##
 
--- a/package/Dockerfile.shipyard-dapper-base
+++ b/package/Dockerfile.shipyard-dapper-base
@@ -55,7 +55,7 @@ RUN dnf -y install --nodocs --setopt=install_weak_deps=False \
 
 ENV LINT_VERSION=v1.46.0 \
     HELM_VERSION=v3.8.0 \
-    KIND_VERSION=v0.12.0 \
+    KIND_VERSION=v0.14.0 \
     BUILDX_VERSION=v0.8.2 \
     GH_VERSION=2.5.1 \
     YQ_VERSION=4.20.2


tpantelis (approved):

dfarrell07:
It seems like we need to increase the inotify settings in the shared script tests. I can take a look.

dfarrell07:
Testing a fix here: https://github.com/dfarrell07/shipyard/pull/41

dfarrell07:
Seems like the inotify settings were not the actual issue. Might still be good to bump them in these tests to follow our warn_inotify function's advice and not making misleading logs.

dfarrell07:
Digging through logs, this seems to be the actual failure:

```
2022-05-18T15:01:25.4137367Z [36m[15:00:24.753] [shipyard]$ [cluster1] command kubectl --context=cluster1 -n kube-system rollout status deploy/coredns --timeout=1m[0m
2022-05-18T15:01:25.4137817Z [cluster1] Waiting for deployment "coredns" rollout to finish: 0 of 2 updated replicas are available...
2022-05-18T15:01:25.4138189Z [cluster1] error: timed out waiting for the condition
2022-05-18T15:01:25.4138529Z [cluster1] Failed to deploy cluster capabilities, removing the cluster
```

skitt:
I’m going to split the kind bump from the k8s bump to see if that changes anything.

skitt:
The CoreDNS pods aren’t starting because of

```
  Warning  FailedCreatePodSandBox  73s   kubelet            Failed to create pod sandbox: rpc error: code = Unknown desc = failed to setup network for sandbox "552c0f5ea87fbd12566
5c018ec47e264d352e965ef61bffbb0fc9303b1f953f5": plugin type="portmap" failed (add): failed to parse config: could not parse prevResult: could not parse prevResult: result type sup
ports [0.3.0 0.3.1 0.4.0] but unmarshalled CNIVersion is "0.1.0"
```

dfarrell07:
```
[~/go/src/submariner-io]$ grep -rniI --exclude-dir=vendor CNIVersion .
[~/go/src/submariner-io]$ echo $?
1
```

:sweat_smile:

dfarrell07:
https://github.com/containernetworking/cni/blob/main/pkg/types/020/types.go#L58

dfarrell07:
> Seems like the inotify settings were not the actual issue.

Made the logs more clear here: https://github.com/submariner-io/shipyard/pull/826

skitt:
The breakage is [caused by the bump to containerd 1.6.4](https://github.com/weaveworks/weave/issues/3942): it expects a `CNIVersion` field in the information provided by the CNI, Weave doesn’t supply this, and the fall-back value is 0.1.0 which isn’t supported (Weave supports 0.3.0 but doesn’t advertise this correctly). This might end up being [fixed in Weave ](https://github.com/weaveworks/weave/pull/3939) and is likely to be [fixed in containerd 1.6.5](https://github.com/containerd/containerd/issues/6921) (through https://github.com/containernetworking/cni/pull/896).

mkolesnik (approved):

scripts/shared/clusters.sh
```diff
@@ -4,13 +4,13 @@
...
 DEFAULT_K8S_VERSION=1.23
 declare -A kind_k8s_versions
-kind_k8s_versions[1.17]=1.17.17@sha256:e477ee64df5731aa4ef4deabbafc34e8d9a686b49178f726563598344a3898d5
-kind_k8s_versions[1.18]=1.18.20@sha256:e3dca5e16116d11363e31639640042a9b1bd2c90f85717a7fc66be34089a8169
-kind_k8s_versions[1.19]=1.19.16@sha256:81f552397c1e6c1f293f967ecb1344d8857613fb978f963c30e907c32f598467
-kind_k8s_versions[1.20]=1.20.15@sha256:393bb9096c6c4d723bb17bceb0896407d7db581532d11ea2839c80b28e5d8deb
-kind_k8s_versions[1.21]=1.21.10@sha256:84709f09756ba4f863769bdcabe5edafc2ada72d3c8c44d6515fc581b66b029c
-kind_k8s_versions[1.22]=1.22.7@sha256:1dfd72d193bf7da64765fd2f2898f78663b9ba366c2aa74be1fd7498a1873166
-kind_k8s_versions[1.23]=1.23.4@sha256:0e34f0d0fd448aa2f2819cfd74e99fe5793a6e4938b328f657c8e3f81ee0dfb9
+kind_k8s_versions[1.18]=1.18.20@sha256:738cdc23ed4be6cc0b7ea277a2ebcc454c8373d7d8fb991a7fcdbd126188e6d7
```
sridhargaddam, line 7: Just curious, why is the `sha` changing even when the image version is the same?

skitt: The images were rebuilt for 0.13.0, with newer versions of `containerd` and the CNI plugins.

sridhargaddam (approved):

skitt:
There is now a kind image for K8s 1.24 with `containerd` 1.6.6, but not for earlier releases yet.

vthapar:
https://github.com/submariner-io/subctl/pull/125 should get CI to pass. Someone should add it as a Depends On

skitt:
Note that the dependency on https://github.com/submariner-io/subctl/issues/102 is temporary; this PR currently changes the default K8s version to 1.24, but that is only because we can test kind 0.14 only with 1.24 for the time being. Before this PR can be merged, the commit changing the default should be removed, which will also remove the dependency.

dfarrell07:
All of the Helm jobs are failing with

```
[cluster1] Installing Submariner...
[16:28:33.529] [shipyard]$ [cluster1] helm --kube-context cluster1 install submariner-operator submariner-latest/submariner-operator --create-namespace --namespace submariner-operator --set ipsec.psk=lWWBU8Qe7sTBCUtmiGj06gdvvsGpmuJkUccILJvfhveWoiOAoJcBy6KSu1NXOcHc --set broker.server=172.18.0.6:6443 --set broker.token= --set broker.namespace=submariner-k8s-broker --set broker.ca= --set broker.globalnet=false --set submariner.serviceDiscovery=false --set submariner.cableDriver=libreswan --set submariner.clusterId=cluster1 --set submariner.clusterCidr=10.1.0.0/16 --set submariner.serviceCidr=100.1.0.0/16 --set submariner.globalCidr= --set serviceAccounts.globalnet.create=false --set serviceAccounts.lighthouseAgent.create=false --set serviceAccounts.lighthouseCoreDns.create=false --set submariner.natEnabled=false --set operator.image.repository=localhost:5000/submariner-operator --set operator.image.tag=local --set operator.image.pullPolicy=IfNotPresent --set submariner.images.repository=localhost:5000 --set submariner.images.tag=local --set brokercrds.create=false
[cluster1] Error: INSTALLATION FAILED: unable to build kubernetes objects from release manifest: error validating "": error validating data: [ValidationError(Submariner.spec): missing required field "brokerK8sApiServerToken" in io.submariner.v1alpha1.Submariner.spec, ValidationError(Submariner.spec): missing required field "brokerK8sCA" in io.submariner.v1alpha1.Submariner.spec]
```

The main parts being `missing required field "brokerK8sApiServerToken"` and `missing required field "brokerK8sCA"`.

dfarrell07:
> [submariner-io/subctl#125](https://github.com/submariner-io/subctl/pull/125) should get CI to pass. Someone should add it as a Depends On

The backport of this (https://github.com/submariner-io/submariner-operator/pull/2121) should get the upgrade job passing once 0.12.2 is released.

dfarrell07:
One interesting thing about the failing Helm jobs is that they constantly fail here but don't fail on any other runs.

Another is that the invocation that fails here passes `--set broker.ca=` and `set broker.token=` as null, whereas the passing runs elsewhere supply actual values.

dfarrell07:
Those missing values come from:

```
    submariner_broker_ca=$(kubectl -n "${BROKER_NAMESPACE}" get secrets -o jsonpath="{.items[?(@.metadata.annotations['kubernetes\.io/service-account\.name']=='${BROKER_CLIENT_SA}')].data['ca\.crt']}")
    submariner_broker_token=$(kubectl -n "${BROKER_NAMESPACE}" get secrets -o jsonpath="{.items[?(@.metadata.annotations['kubernetes\.io/service-account\.name']=='${BROKER_CLIENT_SA}')].data.token}"|base64 --decode)
```

This seems like it's related to the recent upstream secrets/tokens changes and our adjustments.

nyechiel:
As per @aswinsuryan, this should also resolve the OVN kind issues we are seeing in CI.

dfarrell07:
I raised a bug to properly track this Helm failure: https://github.com/submariner-io/submariner-charts/issues/236

scripts/shared/clusters.sh
```diff
@@ -2,15 +2,15 @@
 
 ## Kubernetes version mapping, as supported by kind ##
 # See the release notes of the kind version in use
-DEFAULT_K8S_VERSION=1.23
+DEFAULT_K8S_VERSION=1.24
```
sridhargaddam, line 5: This change seems to be part of the following PR - https://github.com/submariner-io/shipyard/pull/859/files

mkolesnik:
We already merged #859 and #852

Issue #731, opened 2022-02-24
Add linting for mixed tabs/spaces in shell scripts


dfarrell07:
We currently have a gap in our linting that allows mixed tabs and spaces in shell scripts (at least). 

We should avoid mixed tabs and spaces for consistency and avoiding bugs.

We currently mostly (at least) use spaces. The easiest answer might just be failing on hard tabs in *.sh files. We could extend that to other types of files if we check and find other gaps (.md?).

mkolesnik:
This can be interesting but two questions:
1. Does `shellcheck` support it?
2. Doesn't `markdownlint` check for no tabs?

For the first I found this open issue, which seems to imply there isnt a check for tabs (nor is there consensus which to use) - https://github.com/koalaman/shellcheck/issues/1659
It seems though that if we need this in shellcheck it would make sense to contribute it there directly

dfarrell07:
Ran into this again, quite annoying gap in linting. I chimed in on the upstream issue noting that I think there may be consensus that mixed tabs and spaces are not good.

PR #826, opened 2022-05-18
Clarify logged inotify warning

dfarrell07:
As-is, this results in logs that seem to imply the automation knows this
is an issue with inotify settings. Make it clear that we're only saying
they are below a recommendation, not that we know they caused a failure.

### DIFF
--- a/scripts/shared/clusters.sh
+++ b/scripts/shared/clusters.sh
@@ -310,7 +310,9 @@ function deploy_cluster_capabilities() {
 
 function warn_inotify() {
     if [[ "$(cat /proc/sys/fs/inotify/max_user_instances)" -lt 512 ]]; then
-        echo "Please increase your inotify settings (currently $(cat /proc/sys/fs/inotify/max_user_watches) and $(cat /proc/sys/fs/inotify/max_user_instances)):"
+        echo "Your inotify settings are lower than our recommendation."
+        echo "This may cause failures in large deployments, but we don't know if it caused this failure."
+        echo "You may need to increase your inotify settings (currently $(cat /proc/sys/fs/inotify/max_user_watches) and $(cat /proc/sys/fs/inotify/max_user_instances)):"
         echo sudo sysctl fs.inotify.max_user_watches=524288
         echo sudo sysctl fs.inotify.max_user_instances=512
         echo 'See https://kind.sigs.k8s.io/docs/user/known-issues/#pod-errors-due-to-too-many-open-files'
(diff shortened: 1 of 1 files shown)


scripts/shared/clusters.sh
```diff
@@ -295,7 +295,9 @@ function deploy_cluster_capabilities() {
 
 function warn_inotify() {
     if [[ "$(cat /proc/sys/fs/inotify/max_user_instances)" -lt 512 ]]; then
-        echo "Please increase your inotify settings (currently $(cat /proc/sys/fs/inotify/max_user_watches) and $(cat /proc/sys/fs/inotify/max_user_instances)):"
+        echo "Your inotify settings are lower than our recommendation"
+        echo "This may cause failures in large deployments, but we don't know if it caused this failure"
```
skitt:
```suggestion
        echo "Your inotify settings are lower than our recommendation."
        echo "This may cause failures in large deployments, but we don't know if it caused this failure."
```

scripts/shared/clusters.sh
```diff
@@ -329,6 +331,7 @@ declare_cidrs
 # Run in subshell to check response, otherwise `set -e` is not honored
 ( run_all_clusters with_retries 3 create_kind_cluster; ) &
 if ! wait $!; then
+		echo "Failed to create kind clusters"
```
skitt:
```suggestion
		echo "Failed to create kind clusters."
```

skitt (changes requested):
Current practice is inconsistent in `clusters.sh`, but let’s end our sentences.

dfarrell07:
Yeah I went back and forth on that, ending them properly sounds good. :+1: :heavy_check_mark:

tpantelis (approved):

scripts/shared/clusters.sh
```diff
@@ -344,6 +346,7 @@ declare_cidrs
 # Run in subshell to check response, otherwise `set -e` is not honored
 ( run_all_clusters with_retries 3 create_kind_cluster; ) &
 if ! wait $!; then
+		echo "Failed to create kind clusters."
```
dfarrell07: I seem to have a hard tab here, fixing...

skitt (approved):

scripts/shared/clusters.sh
dfarrell07: https://github.com/submariner-io/shipyard/issues/731#issuecomment-1131715084

dfarrell07:
The OVN job should not be required. I thought it wasn't? We need https://github.com/submariner-io/submariner/pull/1689 to land to fix OVN CI.

tpantelis:
> The OVN job should not be required. I thought it wasn't? We need [submariner-io/submariner#1689](https://github.com/submariner-io/submariner/pull/1689) to land to fix OVN CI.

Right.  I'll change it.

PR #830, opened 2022-05-25
Quote parameter expansions

skitt:
The ${var:+"$var"} variant expands var, properly quoted, only if it's defined
and non-empty; this avoids placing empty values in lists of
arguments.

Signed-off-by: Stephen Kitt <skitt@redhat.com>

PR #831, opened 2022-05-27
Fix unit test failures

tpantelis:
Introduced by https://github.com/submariner-io/shipyard/pull/830. See example [run](https://github.com/submariner-io/admiral/runs/6630852183?check_suite_focus=true).


mkolesnik (commented):
While this will fix the base failure, expanding folder names like this is error prone.

I introduced a more robust way to do it in #832

PR #834, opened 2022-05-30
Add a GHA to check that unit test failures are caught

skitt:
Depends on #833.

### DIFF
--- a/.github/workflows/unit.yml
+++ b/.github/workflows/unit.yml
@@ -1,0 +1,15 @@
+---
+name: Unit Test Validation
+
+on:
+  pull_request:
+
+jobs:
+  check-unit-tests-fail:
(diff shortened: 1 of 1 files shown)


skitt:
The failure is expected here; currently failing unit tests are mis-detected. #833 will fix that.

.github/workflows/unit.yml
```diff
@@ -0,0 +1,15 @@
+---
+name: Unit test validation
```
dfarrell07:
```suggestion
name: Unit Test Validation
```

This feels like quite a nit, but it would make the workflow name consistent with all the others ones in this repo (and I think other SubM repos).

dfarrell07 (approved):

tpantelis (approved):

PR #840, opened 2022-06-13
Make weave deploy more robust

mkolesnik:
Sometimes we try to await the weave pods before any of them are ready.
This fails due to [1], so in the meantime implementing this workaround
makes deploying with weave much more robust (and stops wasting time
trying to redeploy).

[1]: https://github.com/kubernetes/kubectl/issues/1516

### DIFF
--- a/scripts/shared/clusters.sh
+++ b/scripts/shared/clusters.sh
@@ -215,9 +215,19 @@ function deploy_weave_cni(){
 
     echo "${WEAVE_YAML}" | kubectl apply -f -
     echo "Waiting for weave-net pods to be ready..."
+    with_retries 5 ensure_weave_pods
     kubectl wait --for=condition=Ready pods -l name=weave-net -n kube-system --timeout="${timeout}"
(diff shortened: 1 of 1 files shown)


skitt:
> Sometimes we try to await the weave pods before any of them are ready.

Would

> Sometimes we try to await the weave pods before any of them have been created.

not be more accurate? The overall goal is to wait for the pods to be ready, but the failure addressed here occurs when we start waiting before any of them have been created.

skitt:
> Sometimes we try to await the weave pods before any of them are ready.

Would

> Sometimes we try to await the weave pods before any of them have been created.

not be more accurate? The overall goal is to wait for the pods to be ready, but the failure addressed here occurs when we start waiting before any of them have been created.

tpantelis (approved):

mkolesnik:
@skitt sure, I've updated the commit message to reflect that

skitt (approved):

PR #842, opened 2022-06-14
Skip releasing images if there are none

skitt:
Some projects ship all their images as multi-arch images; in such
cases, the first release step (non-multi-arch images) must be
skipped.

This fixes the Submariner multi-arch release job.


mkolesnik (approved):

nyechiel (approved):

nyechiel:
@skitt do we also need to cherry-pick it to release-0.12?

tpantelis (approved):

skitt:
> @skitt do we also need to cherry-pick it to release-0.12?

FTR, we don’t need this in `release-0.12` which doesn’t push multi-arch images.

PR #853, opened 2022-06-30
Remove `deploytool_.*_args`

mkolesnik:
These aren't really used and are just dead code now.


skitt (approved):

skitt:
Nice, I wondered if anyone was going to follow up on the `using` cleanups ;-).

mkolesnik:
> Nice, I wondered if anyone was going to follow up on the `using` cleanups ;-).

there's technically usage of them on subm-charts (read only, no assign) for e2e, but the whole thing there is copying the `deploy_helm` file which just seems like a bad approach.

anyway as it's bash there's no issue trying to access undeclared variables, so it should be fine, but ill see what to do to more properly do it there (that file shouldnt be duplicated there)

dfarrell07 (approved):

tpantelis (approved):

PR #856, opened 2022-07-05
Adding basic scale testing capability

### LINKED ISSUE #655: Develop ability to run scale tests: `make scale`
We would like to create tooling that allows running multiple kind clusters in scale testing on one or more machines.

The tool should:
 * Receive amount of clusters to create and an optional broker info file, and bring up these kind clusters.
 * Detect and adjust environment settings (open files limit, etc)
 * Bring up the clusters in a roll-out fashion to avoid slow down due to contention.

We could also use the tool to run longevity testing on the scale environment.

Depends on: #582

mkolesnik:
Adding `make scale` command which will bring up multiple KIND servers
locally to be used for scale testing.

This was part of the scale work, pushing it for posterity in case we want to continue with that work.

Resolves #655 


Makefile
```diff
@@ -55,6 +57,19 @@ include Makefile.dapper
...
 
+scale-prereqs:
+	sudo sysctl -w fs.inotify.max_user_watches=1073741824
+	sudo sysctl -w fs.inotify.max_user_instances=524288
+	sudo sysctl -w kernel.pty.max=524288
+	sudo sysctl -w kernel.pid_max=524288
+	sudo sysctl -w vm.swappiness=5
+	sudo sysctl net.ipv4.neigh.default.gc_thresh1=2048
+	sudo sysctl net.ipv4.neigh.default.gc_thresh2=4096
+	sudo sysctl net.ipv4.neigh.default.gc_thresh3=8092
```
skitt: For consistency:

```suggestion
	sudo sysctl -w net.ipv4.neigh.default.gc_thresh1=2048
	sudo sysctl -w net.ipv4.neigh.default.gc_thresh2=4096
	sudo sysctl -w net.ipv4.neigh.default.gc_thresh3=8192
```

Makefile
```diff
@@ -55,6 +57,19 @@ include Makefile.dapper
 # Make sure linting goals have up-to-date linting image
 $(LINTING_GOALS): package/.image.shipyard-linting
 
+scale-prereqs:
+	sudo sysctl -w fs.inotify.max_user_watches=1073741824
+	sudo sysctl -w fs.inotify.max_user_instances=524288
+	sudo sysctl -w kernel.pty.max=524288
+	sudo sysctl -w kernel.pid_max=524288
```
skitt: This is smaller than the default in RHEL 8, Fedora 35/36, Debian 11 etc. Is that intentional?

More generally, could you describe why each setting is used? Some of them (`vm.swappiness` in particular) will have a lasting effect on the system’s behaviour, it would be good to warn the user about that.

skitt (changes requested):

Makefile
mkolesnik: Thanks for the review, I updated it hope its clearer now

dfarrell07:
Needs a rebase @mkolesnik and re-review @skitt.

skitt (approved):

dfarrell07 (approved):

dfarrell07:
Needs a manual rebase.

PR #857, opened 2022-07-05
Check freshness of global `Dockerfile.*` files

### LINKED ISSUE #815: Shipyard doesn’t flag outdated Dockerfile.dapper
Shipyard should flag outdated Dockerfile.dapper files, as it does for Makefile.dapper, otherwise we end up with problems like https://github.com/submariner-io/shipyard/pull/813 with no indication of what’s really happening.

We need to handle projects which have their own `Dockerfile.dapper`.

mkolesnik:
Add a check for freshness of these files as well, as they could go
stale.

Resolves #815 

### DIFF
--- a/Makefile.inc
+++ b/Makefile.inc
@@ -3,13 +3,27 @@ USING = $(subst $(,), ,$(using))
 _using = ${USING}
 PRELOAD_IMAGES ?= submariner-gateway submariner-operator submariner-route-agent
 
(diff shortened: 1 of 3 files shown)


skitt (approved):

dfarrell07 (approved):
Nice, this would have saved me some debugging frustration from time to time.

tpantelis (approved):

Issue #887, opened 2022-07-26
Stop using `shflags` in scripts, use environment variables directly


mkolesnik:
Initially the flags were added as a mechanism to make sure scripts are being called correctly, but we never actually call the scripts directly and rather used commands such as `make deploy using=globalnet`.
Another goal of using the flags was avoiding overloading the same variable for different scripts that use it differently (a problem we had way back with the first scripts).

Nowadays, these flags are rather useless as no one runs the scripts directly (as mentioned previously), and all interaction with the scripts is done via `make` which already has logic to handle different deployment options.
Additionally, our scripts are quite stable now, so there's no problem of overloaded variables used differently in different scripts.

Moreover, we already have a few instances where the scripts interact with environment variables directly and avoid the flags altogether, which leads to a lack of consistency.

Removing the flags and using the variables directly would simplify our code, while lowering the barrier to learn and maintain the project.

From a UX perspective little if anything will change, as the existing commands should all work as previously. If anything, this'll make overriding scripts behavior directly easier e.g. `make deploy TIMEOUT=1m` vs `make deploy DEPLOY_ARGS="--timeout 1m"`

Any CI caller that does use flags will be updated to use the variables instead.

We should retain the informative printouts (eg `echo "Running with: globalnet=${globalnet@Q}, deploytool=${deploytool@Q}, settings=...`) for each script which would list any user specified variables that affect it's behavior.

mkolesnik:
This is mostly done, we can clean up staggering places as we come across them

PR #890, opened 2022-07-27
Consume ENV vars where possible

mkolesnik:
Instead of consuming flags, consume environment variables where
possible.
Some flags are retained for backwards compatibility not to break
consumers, and will be removed once their consumers are updated.

This is the first step in resolving #887 

Depends on #886 


.github/workflows/linting.yml
```diff
@@ -40,7 +40,7 @@ jobs:
       - name: Make sure each commit in the PR is within reviewable size
         uses: ./gh-actions/commit-size
         with:
-          size: 200
+          size: 250
```
skitt, line 43: If this gets updated for any reason, could you mention this change in the corresponding commit message?

Makefile.inc
```diff
@@ -31,73 +31,58 @@ include $(SHIPYARD_DIR)/Makefile.versions
 # Process extra flags from the `using=a,b,c` optional flag
 
 ifneq (,$(filter libreswan,$(_using)))
-override DEPLOY_ARGS += --cable_driver libreswan
+export CABLE_DRIVER = libreswan
```
skitt: Since these are exported further down, how about

    CABLE_DRIVER = lighthouse

Same for the other variables.

skitt (changes requested):

scripts/shared/lib/utils
```diff
@@ -189,3 +189,10 @@ function load_settings_from_yaml() {
         cluster_subm["${cluster}"]=$(_yq ".submariner")
     done
 }
+
+function print_env() {
+   echo "Running with environment variables:"
+   for var; do
+       eval echo "${var}=\${${var}@Q}"
+   done
```
skitt: The safe way of doing this is

```
    for var; do
        declare -n value=$var
        printf '%s=%q\n' "$var" "$value"
    done
```

skitt (changes requested):
Regarding the cache-removing commit, note that we only use `buildx` for multi-arch images.

scripts/shared/lib/utils
mkolesnik: Cool, good to know :D

mkolesnik:
> Regarding the cache-removing commit, note that we only use `buildx` for multi-arch images.

Yes you're right of course, it's BuildKit that has the improvements (well probably buildx too as BuildKit is it's backed).
I updated the commit message.

skitt (approved):

tpantelis (approved):

PR #893, opened 2022-07-31
Extract `get-subctl.sh` to easily fetch `subctl`

mkolesnik:
This will be useful both for scripts and for `make shell` users who need
to quickly get subctl for various operations

### DIFF
--- a/package/Dockerfile.shipyard-dapper-base
+++ b/package/Dockerfile.shipyard-dapper-base
@@ -79,6 +79,9 @@ RUN curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/i
     find /go/bin /usr/local/libexec/docker/cli-plugins -type f -executable -newercc /proc \( -execdir upx ${UPX_LEVEL} {} \; -o -true \) && \
     go clean -cache -modcache
 
+# Link get-subctl script so it can be easily run inside a shell
+RUN ln -s $SCRIPTS_DIR/get-subctl.sh /root/.local/bin/subctl
+
 # Copy kubecfg to always run on the shell
 COPY scripts/shared/lib/kubecfg /etc/profile.d/kubecfg.sh
 
--- a/scripts/shared/get-subctl.sh
+++ b/scripts/shared/get-subctl.sh
@@ -1,0 +1,17 @@
+#!/usr/bin/env bash
+
+set -e
+
+# In case we're pretending to be `subctl`
+if [[ "${0##*/}" = subctl ]] && [[ -L "$0" ]]; then
+    run_subctl=true
+
+    # Delete ourselves to ensure we don't run into issues with the new subctl
+    rm -f "$0"
+fi
+
+# Default to devel if we don't know what base branch were on
+curl -Ls --retry 3 https://get.submariner.io | VERSION="${SUBCTL_VERSION:-${BASE_BRANCH:-devel}}" bash
+
+# If we're pretending to be subctl, run subctl with any given arguments
+[[ -z "${run_subctl}" ]] || subctl "$@"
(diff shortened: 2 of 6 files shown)


package/Dockerfile.shipyard-dapper-base
```diff
@@ -79,6 +79,9 @@ RUN curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/i
     find /go/bin /usr/local/libexec/docker/cli-plugins -type f -executable -newercc /proc \( -execdir upx ${UPX_LEVEL} {} \; -o -true \) && \
     go clean -cache -modcache
 
+# Link get-subctl script so it can be easily run inside a shell
+RUN ln -s $SCRIPTS_DIR/get-subctl.sh /root/.local/bin/subctl
```
skitt, line 83: Because of the way `getsubctl.sh` (get.submariner.io) works, this can’t work — `getsubctl.sh` overwrites the target file instead of moving it into place. In this scenario, the `get-subctl.sh` script changes while it’s running, which results in

```
Installing subctl version devel
  OS detected:           linux
  Architecture detected: amd64
  Download URL:          https://github.com/submariner-io/subctl/releases/download/subctl-devel/subctl-devel-linux-amd64.tar.xz

Downloading...
subctl-devel-linux-amd64 has been installed as /root/.local/bin/subctl
This provides subctl version: devel
/opt/shipyard/scripts/get-subctl.sh: line 14: syntax error near unexpected token `$'\201\024\002''
```

_e.g._ in https://github.com/submariner-io/shipyard/runs/7605145306?check_suite_focus=true#step:7:1482

https://github.com/submariner-io/get.submariner.io/pull/30 will fix this.

mkolesnik: Hence why I added the line https://github.com/submariner-io/shipyard/pull/893/files#diff-b57b55d512de87ffb6a16ac5dbb0069de967a4bc2630f9ba5ca03f41355a05c0R10

But  seems that it didn't work for the consuming job for some reason, so I'm fine with getting the `get.submariner` fix first and then this one.

skitt: Ah yes, that should have done the trick! But `command -v subctl` in the `subctl` repository will point to the `subctl` binary built in the repository, not `get-subctl.sh`-as-`subctl`.

scripts/shared/get-subctl.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+
+set -e
+
+# In case we're pretending to be `subctl`
+[[ -L "$0" ]] && run_subctl=true
+
+# Try to unlink original "subctl" stub in case it still exists
+orig_subctl="$(command -v subctl)"
+[[ -L "${orig_subctl}" ]] && unlink "${orig_subctl}"
```
skitt: Let’s only delete the link if we’re running as `subctl`, and delete the actual link rather than whatever is first on the `PATH`.

```suggestion
# In case we're pretending to be `subctl`
if [[ "${0##*/}" = subctl ]] && [[ -L "$0" ]]; then
    run_subctl=true

    # Delete ourselves to ensure we don't run into issues with the new subctl
    rm -f "$0"
fi
```

skitt (changes requested):

package/Dockerfile.shipyard-dapper-base
mkolesnik: Yes that must be why it failed, well it should be working now.

I see you suggested a better way :)

mkolesnik:
Thanks @skitt for the help.

I also added a release note to clarify this.

skitt (approved):

tpantelis (approved):

PR #895, opened 2022-07-31
Support multi arch compile inside `compile.sh`

mkolesnik:
Instead of relying on callers of the script to properly set the
`GOARCH`, detect it inside the script.

Callers can still set `GOARCH` if they choose to.

### DIFF
--- a/scripts/shared/compile.sh
+++ b/scripts/shared/compile.sh
@@ -27,8 +27,21 @@ source "${SCRIPTS_DIR}/lib/utils"
 print_env BUILD_DEBUG BUILD_UPX LDFLAGS
 source "${SCRIPTS_DIR}/lib/debug_functions"
 
+### Functions ###
+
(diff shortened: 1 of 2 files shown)


scripts/shared/compile.sh
```diff
@@ -25,8 +25,20 @@ set -e
 
 source "${SCRIPTS_DIR}/lib/debug_functions"
 
+### Functions ###
+
+function determine_goarch() {
```
skitt, line 33: To clarify what this is supposed to do:

```suggestion
# Determine GOARCH based on the last component of the target directory, if any
function determine_goarch() {
```

skitt (changes requested):

skitt (approved):

tpantelis (approved):

dfarrell07 (approved):

PR #909, opened 2022-08-07
Remove E2E_ARGS, get from settings

mkolesnik:
We use Shipyard's settings capability everywhere, which already dictates
the clusters deployment topology.
Instead of forcing the users to explicitly state the clusters for E2E,
deduce them from the settings.

Special care is given to place the biggest cluster first, as some tests
rely on that.


skitt (approved):

dfarrell07 (approved):

dfarrell07:
Needs a manual rebase.

PR #914, opened 2022-08-10
Gather and restore GW nodes state

MaxBab:
Add functions that will gather the state of the gateway nodes on all
clusters and restore it when required.
Will be used in redundancy tests.

- The "GatherClustersGatewaysState" function will execute during
  "BeforeSuite" execution and store the GW state globally.
- The "RestoreClustersGatewaysState" will be called by redundancy tests
  and will restore the state of GW nodes to the initial state.

The change will replace a static definition of GW state during restore.

### DIFF
--- a/test/e2e/framework/framework.go
+++ b/test/e2e/framework/framework.go
@@ -105,9 +105,10 @@ type Framework struct {
 var (
 	beforeSuiteFuncs []func()
 
-	RestConfigs []*rest.Config
-	KubeClients []*kubeclientset.Clientset
-	DynClients  []dynamic.Interface
+	RestConfigs           []*rest.Config
+	KubeClients           []*kubeclientset.Clientset
+	DynClients            []dynamic.Interface
+	ClustersGatewaysState map[int][]string
 )
 
 // NewBareFramework creates a test framework, without ginkgo dependencies.
@@ -183,6 +184,7 @@ func BeforeSuite() {
 	}
 
 	fetchClusterIDs()
+	GatherClustersGatewaysState()
 
 	err := mcsv1a1.AddToScheme(scheme.Scheme)
 	Expect(err).NotTo(HaveOccurred())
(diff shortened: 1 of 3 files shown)


tpantelis (changes requested):
From my understanding of slack discussions, simulating node fail-over via changing GW labeling is not the approach we want to take. We talked about using another node label/affinity to effectively disable a node from becoming the active GW. Another approach is to figure out a way to actually "crash" the node.

MaxBab:
@tpantelis 

You're right.
But this change relevant for the local (kind) environment or any other where the MachineSet is not applied.
And it reflects the current failure of the kind e2e execution where an environment of two gateway nodes is created (the default is one single gateway node), the gateway restore tries to make a restore of a single gateway node only.
And in case the cluster contains two GW nodes, the restore will fail.
https://github.com/submariner-io/shipyard/blob/devel/test/e2e/framework/gateways.go#L189

Beside the gateway node fail-over test the we spoke about in the slack thread, the gateway label change is made in "[expansion] Test expanding/shrinking an existing cluster fleet" test, which I'm not sure I'm going to refactor.
https://github.com/submariner-io/submariner/blob/devel/test/e2e/cluster/add_remove_cluster.go#L90

What do you think?

tpantelis:
>But this change relevant for the local (kind) environment or any other where the MachineSet is not applied.
And it reflects the current failure of the kind e2e execution where an environment of two gateway nodes is created (the default is one single gateway node), the gateway restore tries to make a restore of a single gateway node only.
And in case the cluster contains two GW nodes, the restore will fail.
https://github.com/submariner-io/shipyard/blob/devel/test/e2e/framework/gateways.go#L189

I'm not sure I follow. I thought we agreed on slack that we don't want to modify GW labels to effect node fail-over. But this PR is related to that. I proposed for 0.14 that we introduce a new `submariner.io/gateway-disabled` label to prevent a node from becoming a GW. As part of test cleanup, we would just ensure no node has the label. This way we don't touch the `submariner.io/gateway` label and conflict with the ACM controller. This approach may not work when `MachineSets` are used though.

> Beside the gateway node fail-over test the we spoke about in the slack thread, the gateway label change is made in "[expansion] Test expanding/shrinking an existing cluster fleet" test, which I'm not sure I'm going to refactor.
https://github.com/submariner-io/submariner/blob/devel/test/e2e/cluster/add_remove_cluster.go#L90

Those tests were pretty much experimental and don't run (ie they're pending).

MaxBab:
Ok, if the second test that is using the labeling is experimental and not really executed, I'll close this PR and start working on a flow we discussed on slack.

PR #921, opened 2022-08-19
Only check out ovn-kubernetes contrib and dist

skitt:
We only need the contrib and dist directories from ovn-kubernetes, so
only check those out. This doesn't save any download time, but it
avoids build-tool confusion due to the introduction of new Go modules
in ovn-kubernetes.

### DIFF
--- a/scripts/shared/clusters.sh
+++ b/scripts/shared/clusters.sh
@@ -317,8 +317,20 @@ function download_ovnk() {
     fi
 
     echo "Cloning ovn-kubernetes from source"
-    git clone https://github.com/ovn-org/ovn-kubernetes.git \
-        || { git -C ovn-kubernetes fetch && git -C ovn-kubernetes reset --hard origin/master; }
+    mkdir -p ovn-kubernetes
+    # We only need the contrib directory, use a sparse checkout
+    cd ovn-kubernetes
+    git init
+    git config core.sparseCheckout true
+    echo contrib/ > .git/info/sparse-checkout
+    echo dist/ >> .git/info/sparse-checkout
+    if git remote add -f origin https://github.com/ovn-org/ovn-kubernetes.git
+    then
+        git pull origin master
+    else
+        git fetch && git reset --hard origin/master
+    fi
+    cd ..
 }
 
 ### Main ###


tpantelis (approved):

scripts/shared/clusters.sh
```diff
@@ -317,8 +317,19 @@ function download_ovnk() {
     fi
 
     echo "Cloning ovn-kubernetes from source"
-    git clone https://github.com/ovn-org/ovn-kubernetes.git \
-        || { git -C ovn-kubernetes fetch && git -C ovn-kubernetes reset --hard origin/master; }
+    mkdir -p ovn-kubernetes
+    # We only need the contrib directory, use a sparse checkout
+    cd ovn-kubernetes
+    git init
+    if git remote add -f origin https://github.com/ovn-org/ovn-kubernetes.git
```
dfarrell07, line 327: If anyone else is wondering, this conditional is being used to check if the ovn-k repo already exists locally. `git remote add` fails if the remote already exists, `git init` doesn't fail if the dir is already a repo.

dfarrell07:
[Deploy (ovn)](https://github.com/submariner-io/shipyard/runs/7920825117?check_suite_focus=true#logs)

```
[cluster2] + create_ovn_kube_manifests
[cluster2] + pushd /go/src/github.com/submariner-io/shipyard/ovn-kubernetes/contrib/../dist/images
[cluster2] ./ovn-kubernetes/contrib/kind.sh: line 558: pushd: /go/src/github.com/submariner-io/shipyard/ovn-kubernetes/contrib/../dist/images: No such file or directory
[cluster2] Failed to install kind with OVN
```

dfarrell07 (approved):

scripts/shared/clusters.sh
skitt: Strictly speaking, `git remote add` fails if the remote already exists ;-).

skitt:
@astoycos FYI

scripts/shared/clusters.sh
astoycos: Ah good to know :)

PR #924, opened 2022-08-20
Use image overrides when deploying 

### LINKED ISSUE #915: Use image overrides when deploying
Right now Shipyard uses a simplistic approach, setting `repository` and `version` when deploying to `localhost:5000` and `local` respectively.
This necessitates that images be pre-loaded to the kind setup, even when testing projects that dont touch them (e.g. `lighthouse`).

Instead, we should use the image overrides feature to specify specific images to override, and let the rest be taken from the default repository.

As we have a `PRELOAD_IMAGES` variable that denotes which images we're preloading, using it would be classic for populating the overrides.

This also allows consuming projects to still preload a specific image if necessary (i.e. for testing changes locally), e.g. on `subctl` one could use `make e2e PRELOAD_IMAGES=submariner-operator`.

We should also add a test for deploying that the preloaded images are actually used (perhaps find if a pod uses a preloaded image, and if none found error out).

mkolesnik:
Use image overrides, based on `$PRELOAD_IMAGES`, for determining which
exact images to override.

This will allow us to test just the images that we want to test, and use
the default images elsewhere.

A dev can still specify specific images, e.g.
`make deploy PRELOAD_IMAGES='submariner-gateway submariner-operator'`

Resolves: #915 


skitt (approved):

tpantelis (approved):

dfarrell07:
This seems to have broken the System Tests workflow in the Operator repo.

To prevent such issues in the future, maybe we should add those tests to the Consuming Projects tests here (maybe under the `e2e-projects` label).

[Deployment (globalnet)](https://github.com/submariner-io/submariner-operator/runs/8115464830?check_suite_focus=true#logs)

```
[14:21:34.055] [dir=submariner-operator; cl=cluster1; fn=validate_equals]$ jq -r -M .spec.repository /tmp/submariner.cluster1.json
[cluster1] Expected 'localhost:5000' but got 'quay.io/submariner'
```

mkolesnik:
> This seems to have broken the System Tests workflow in the Operator repo.
> 
> To prevent such issues in the future, maybe we should add those tests to the Consuming Projects tests here (maybe under the `e2e-projects` label).
> 
> [Deployment (globalnet)](https://github.com/submariner-io/submariner-operator/runs/8115464830?check_suite_focus=true#logs)
> 
> ```
> [14:21:34.055] [dir=submariner-operator; cl=cluster1; fn=validate_equals]$ jq -r -M .spec.repository /tmp/submariner.cluster1.json
> [cluster1] Expected 'localhost:5000' but got 'quay.io/submariner'
> ```

Yes that might be wise, anyhow I'll fix the tests

mkolesnik:
Fix for operator system tests - https://github.com/submariner-io/submariner-operator/pull/2222

PR #944, opened 2022-09-13
Parametrize gateway topology deployment

MaxBab:
Add "gateways: 2" parameter to ".shipyard*" file in order to
parametrize the amount of nodes that should be labeld as
gateway nodes.
In case of the label missing, both on the cluster or global level, the
GW node label will be applied automatically on single node.
 
Example of configuration:
```
submariner: true
nodes: control-plane worker
clusters:
  cluster1:
  cluster2:
    nodes: control-plane worker worker
    gateways: 2
```

In the above example, the first cluster will get 1 gw node applied and
second cluster will have 2 gw nodes applied.


.shipyard.e2e.yml
```diff
@@ -5,3 +5,4 @@ nodes: control-plane worker
 clusters:
   cluster1:
   cluster2:
+    nodes: control-plane worker worker
```
mkolesnik: This should go in the project that needs it though, shipyard itself has no real need to test with that many modes

scripts/shared/cloud-prepare.sh
```diff
@@ -29,7 +29,58 @@ function check_gateway_exists() {
 
 function prepare_kind() {
     read -r -a nodes <<< "${cluster_nodes[$cluster]}"
-    kubectl label node "${cluster}-${nodes[-1]}" "${GATEWAY_LABEL}" --overwrite
+
+    # In order to test all gateway fail over test scenarios,
+    # one cluster will label 1 GW node and second 2 GW nodes.
+    worker_count=$(grep -o worker <<< "${nodes[*]}" | wc -l)
```
mkolesnik: I think it'll be more in line with Shipyard if you added support for labeling GW nodes in the settings, this way you can specify in the settings explicitly which nodes to label if you want (and if you dont, the default can remain the last node on the list).

scripts/shared/cloud-prepare.sh
```diff
@@ -29,7 +29,58 @@ function check_gateway_exists() {
...
+        kubectl label node "${cluster}-${nodes[-1]}" "${GATEWAY_LABEL}" --overwrite
+    fi
+
+    verify_gw_topology
+}
+
+# Verify that created clusters contains required gateway labeling
+# First cluster will contain 1 GW node
+# Second cluster will contain 2 GW nodes in active/passive mode
+function verify_gw_topology() {
```
mkolesnik: There's no reason to do it here, but you could add a dedicated test for it in the tests.

mkolesnik (changes requested):

.shipyard.e2e.yml
MaxBab: ack. Will remove.

scripts/shared/cloud-prepare.sh
MaxBab: @tpantelis explicitly asked me to add this verification to ensure the requested topology deployed.

scripts/shared/cloud-prepare.sh
MaxBab: By "adding support for labeling GW node in the setting" you mean to add a parameterized argument here?
https://github.com/submariner-io/shipyard/blob/devel/scripts/shared/lib/utils#L207

mkolesnik: Yes, specifically in this loop https://github.com/submariner-io/shipyard/blob/devel/scripts/shared/lib/utils#L227

scripts/shared/cloud-prepare.sh
mkolesnik: This seems more like a "system test" to me, which should be done in testing (same as we do for other targets - https://github.com/submariner-io/shipyard/tree/devel/test/scripts).

.shipyard.e2e.yml
MaxBab: Done

scripts/shared/cloud-prepare.sh
MaxBab: Implemented

scripts/shared/cloud-prepare.sh
MaxBab: Moved near to "verify_gw_status" function

scripts/shared/cloud-prepare.sh
```diff
@@ -28,8 +28,33 @@ function check_gateway_exists() {
 }
 
 function prepare_kind() {
-    read -r -a nodes <<< "${cluster_nodes[$cluster]}"
-    kubectl label node "${cluster}-${nodes[-1]}" "${GATEWAY_LABEL}" --overwrite
+    read -r -a nodes <<< "${label_gw_nodes[$cluster]}"
+
+    echo "yoyoyoyo"
```
sridhargaddam: yoyoyyo :)

MaxBab: Oh, removed. :)
Thanks.

scripts/shared/cloud-prepare.sh
```diff
@@ -28,8 +28,30 @@ function check_gateway_exists() {
 }
 
 function prepare_kind() {
-    read -r -a nodes <<< "${cluster_nodes[$cluster]}"
-    kubectl label node "${cluster}-${nodes[-1]}" "${GATEWAY_LABEL}" --overwrite
+    read -r -a nodes <<< "${label_gw_nodes[$cluster]}"
+
+    worker_count=$(grep -o worker <<< "${nodes[*]}" | wc -l || true)
+    master_count=$(grep -o control-plane <<< "${nodes[*]}" | wc -l || true)
+
```
tpantelis, line 32: I think we should default to `label-gateway: worker`, ie if `worker_count==0 && master_count==0 ` then `worker_count=1` to match the previous behavior and to avoid changing YAML in every project.

tpantelis (changes requested):

scripts/shared/cloud-prepare.sh
MaxBab: In the Slack discussion @skitt mentioned that he prefers to be able to define default parameter.
https://coreos.slack.com/archives/C0134E73VH6/p1663674120818639

So, according to your comment, I can suggest that we would leave the `label-gateway` parameter for the `.shipyard.e2e*` global definition, but it will not be mandatory:
```
---
submariner: true
nodes: control-plane worker
label-gateway: worker
clusters:
  cluster1:
  cluster2:
```

So if that parameter is not defined, ex.:
1)
```
---
submariner: true
nodes: control-plane worker
clusters:
  cluster1:
  cluster2:
```
or
2)
```
---
submariner: true
nodes: control-plane worker
clusters:
  cluster1:
  cluster2:
    nodes: control-plane worker worker
    label-gateway: worker worker
```

then, a single worker node as a gateway will be applied on all clusters in first example and on first cluster in second example.

@tpantelis 
What do you think?

MaxBab: Ok, let me clarigy a bit about my above comment.

If the `'label-gateway: worker` parameter is not defined on global or per cluster level, a single worker gateway node will be applied.
But if label parameter applied on global level, it will be used by all the clusters.
And if label parameter applied on specific cluster, this cluster will use the defined config and rest of the clusters will apply single worker node.

Is it acceptable?

tpantelis: That sounds reasonable so precedence is:

* Use `label-gateway` parameter if defined at the cluster level or
* Use `label-gateway` parameter if defined at the global level or
* No `label-gateway` defined so label one worker node as before

MaxBab: Exactly.
Will work on it.

MaxBab: Done

mkolesnik: Just to be clear, current default is labeling the last node (doesnt matter if its control-plane or worker).

scripts/shared/cloud-prepare.sh
```diff
@@ -27,18 +27,55 @@ function check_gateway_exists() {
...
         pub_ip=$(kubectl get nodes "$node" -o jsonpath="{.status.addresses[0].address}")
         kubectl annotate node "$node" gateway.submariner.io/public-ip=ipv4:"$pub_ip"
     fi
 }
 
+function prepare_kind() {
+    read -r -a nodes <<< "${label_gw_nodes[$cluster]}"
+
+    worker_count=$(grep -o worker <<< "${nodes[*]}" | wc -l || true)
+    master_count=$(grep -o control-plane <<< "${nodes[*]}" | wc -l || true)
```
mkolesnik: Do we care if the labeled node is control plane or worker?

If not, it would be simpler to have something like `gateways: 2` and let shipyard choose which 2 nodes to label.

scripts/shared/cloud-prepare.sh
```diff
@@ -27,18 +27,55 @@ function check_gateway_exists() {
     [[ $(kubectl get nodes -l "${GATEWAY_LABEL}" --no-headers | wc -l) -gt 0 ]]
 }
 
-function prepare_kind() {
-    read -r -a nodes <<< "${cluster_nodes[$cluster]}"
-    local node=${cluster}-${nodes[-1]}
-    kubectl label node "$node" "$GATEWAY_LABEL" --overwrite
```
mkolesnik, line 33: Why remove the labeling from here?
Now this call is duplicated 3 times..

scripts/shared/cloud-prepare.sh
```diff
@@ -27,18 +27,55 @@ function check_gateway_exists() {
...
+function check_and_apply_air_gap_env() {
```
mkolesnik: A better name would be `prepare_kind_node`, and you can keep the labeling logic inside

scripts/shared/cloud-prepare.sh
```diff
@@ -27,18 +27,55 @@ function check_gateway_exists() {
...
+        echo "Label $worker_count worker nodes as gateway"
+        worker_nodes=$(kubectl get nodes --no-headers=true \
+            -o custom-columns=NAME:.metadata.name | grep worker | head -n "$worker_count")
+
+        for node in ${worker_nodes}; do
+            kubectl label node "${node}" "${GATEWAY_LABEL}" --overwrite
+
+            check_and_apply_air_gap_env "$node"
+        done
+    else
```
mkolesnik: This will just always label the last node in case you didnt specify any worker nodes to label.

scripts/shared/lib/deploy_funcs
```diff
@@ -73,6 +73,37 @@ function connectivity_tests() {
     with_context "$target_cluster" remove_resource "${RESOURCES_DIR}/nginx-demo.yaml"
 }
 
+function verify_gw_topology() {
```
mkolesnik: This is a good thing to test, but as suggested previously please move to a distinct test that would test the different scenarios

.shipyard.e2e.ovn.yml
```diff
@@ -2,6 +2,7 @@
 cni: ovn
 submariner: true
 nodes: control-plane worker
+label-gateway: worker
```
mkolesnik: There no need for this then..

.shipyard.e2e.yml
```diff
@@ -1,6 +1,7 @@
 ---
 submariner: true
 nodes: control-plane worker
+label-gateway: worker
```
mkolesnik: Also no need for this

mkolesnik (changes requested):

scripts/shared/cloud-prepare.sh
MaxBab: @mkolesnik 
I would say we do care.
Normally, the gw node should be always a worker node.
And just in some outstanding cases place it on control-plane.

scripts/shared/cloud-prepare.sh
MaxBab: @mkolesnik 
I removed the labeling because I need to loop over the selected nodes.

scripts/shared/cloud-prepare.sh
MaxBab: @mkolesnik 
I created a separate function for the air gup config to not duplicate the same lines.

scripts/shared/cloud-prepare.sh
MaxBab: @mkolesnik 
Yes, that's the point.
It's a default config that the flow will fall back to if "label-gateway" parameter will not be specified.

.shipyard.e2e.ovn.yml
MaxBab: @mkolesnik 
I know it's not needed.
I placed it here to serve as an example.
Otherwise, people will not be aware of that option.

scripts/shared/lib/deploy_funcs
MaxBab: @mkolesnik 
Sorry, I'm not sure I understand the meaning.
Where exactly verification should be placed and how it will be executed?

.shipyard.e2e.ovn.yml
mkolesnik: There are other options not in the files, best way to raise awareness is document on the website: https://submariner.io/development/shipyard/settings/

scripts/shared/cloud-prepare.sh
mkolesnik: Yes you duplicate the labeling call, which also needs to happen per node, hence my comment on the "before" code - https://github.com/submariner-io/shipyard/pull/944/files#r1002693431

scripts/shared/cloud-prepare.sh
mkolesnik: You could just assemble a list of nodes here and label all of them in a loop..

Also labeling still needs to happen per node just like air gap handling

scripts/shared/cloud-prepare.sh
mkolesnik: Can you show an example in docs or code where that distinction actually matters?

AFAICT we don't care where the GW node is, the only thing we care about is that it had the correct infra support (ports can be opened, public IP, etc).

Going with a count instead of "worker/control-plane" distinction would be easier to use and maintain.

scripts/shared/cloud-prepare.sh
mkolesnik: You're missing the point of my original comment, in case you specify `label-gateway: control-plane` it would still label the last node - and in case you have any worker nodes, you will likely end up with 2 gateways.

scripts/shared/lib/deploy_funcs
mkolesnik: The tests for scripts are here: https://github.com/submariner-io/shipyard/tree/devel/test/scripts

You would need to add a file to test deploy

scripts/shared/cloud-prepare.sh
MaxBab: Changed to set the parameter as `gateways: x`.

scripts/shared/cloud-prepare.sh
MaxBab: Fixed.

scripts/shared/cloud-prepare.sh
MaxBab: Fixed.

.shipyard.e2e.ovn.yml
MaxBab: Removed.
Will add it within the documentation.

.shipyard.e2e.yml
MaxBab: Removed.

scripts/shared/cloud-prepare.sh
MaxBab: Fixed.

scripts/shared/lib/deploy_funcs
MaxBab: @mkolesnik 

As per the link you shared, I see that the only place where these scripts are called, are here:
https://github.com/submariner-io/shipyard/blob/devel/.github/workflows/scripts.yml

But I still don't have a full understanding of how exactly it should be placed and how it going to be called.
Appreciate more details.

scripts/shared/lib/utils
```diff
@@ -217,7 +217,7 @@ function load_settings() {
     local DEBUG_PRINT=false
     expect_env SETTINGS "Deployment settings file"
     declare -ga clusters
-    declare -gA cluster_cni cluster_nodes cluster_subm
+    declare -gA cluster_cni cluster_nodes cluster_subm gateways
```
mkolesnik:
```suggestion
    declare -gA cluster_cni cluster_nodes cluster_subm cluster_gateways
```

For consistency

scripts/shared/cloud-prepare.sh
```diff
@@ -28,15 +28,21 @@ function check_gateway_exists() {
...
-    if [[ "$AIR_GAPPED" = true ]]; then
-        local pub_ip
-        pub_ip=$(kubectl get nodes "$node" -o jsonpath="{.status.addresses[0].address}")
-        kubectl annotate node "$node" gateway.submariner.io/public-ip=ipv4:"$pub_ip"
-    fi
+    local gw_count
+    local gw_nodes
+    gw_count="${gateways[$cluster]:-1}"
+    gw_nodes=$(kubectl get nodes --no-headers=true \
+        -o custom-columns=NAME:.metadata.name | tail -n "$gw_count")
```
mkolesnik:
```suggestion
    local gw_count=${cluster_gateways[$cluster]:-1}
    readarray -d' ' nodes < <(kubectl get nodes -o jsonpath='{.items[*].metadata.name}')
```

scripts/shared/cloud-prepare.sh
```diff
@@ -28,15 +28,21 @@ function check_gateway_exists() {
...
+
+    for gw_node in ${gw_nodes}; do
+        kubectl label node "$gw_node" "$GATEWAY_LABEL" --overwrite
+
+        if [[ "$AIR_GAPPED" = true ]]; then
+            local pub_ip
+            pub_ip=$(kubectl get nodes "$gw_node" -o jsonpath="{.status.addresses[0].address}")
+            kubectl annotate node "$gw_node" gateway.submariner.io/public-ip=ipv4:"$pub_ip"
```
mkolesnik:
```suggestion
            pub_ip=$(kubectl get nodes "$node" -o jsonpath="{.status.addresses[0].address}")
            kubectl annotate node "$node" gateway.submariner.io/public-ip=ipv4:"$pub_ip"
```

scripts/shared/cloud-prepare.sh
```diff
@@ -28,15 +28,21 @@ function check_gateway_exists() {
...
-        kubectl annotate node "$node" gateway.submariner.io/public-ip=ipv4:"$pub_ip"
-    fi
+    local gw_count
+    local gw_nodes
+    gw_count="${gateways[$cluster]:-1}"
+    gw_nodes=$(kubectl get nodes --no-headers=true \
+        -o custom-columns=NAME:.metadata.name | tail -n "$gw_count")
+
+    for gw_node in ${gw_nodes}; do
+        kubectl label node "$gw_node" "$GATEWAY_LABEL" --overwrite
```
mkolesnik:
```suggestion
    for node in "${nodes[@]:0:$gw_count}"; do
        kubectl label node "$node" "$GATEWAY_LABEL" --overwrite
```

scripts/shared/lib/deploy_funcs
mkolesnik: You'll need to create a new test script to test deploy, and call it from that GHA that you mentioned.

scripts/shared/lib/utils
```diff
@@ -238,6 +238,7 @@ function load_settings() {
         cluster_nodes["${cluster}"]=$(_yq ".clusters.${cluster}.nodes // .nodes")
         cluster_subm["${cluster}"]=$(_yq ".clusters.${cluster}.submariner")
         cluster_subm["${cluster}"]=$(_yq ".submariner")
+        gateways["${cluster}"]=$(_yq ".clusters.${cluster}.gateways // .gateways")
```
mkolesnik:
```suggestion
        cluster_gateways["${cluster}"]=$(_yq ".clusters.${cluster}.gateways // .gateways")
```

scripts/shared/lib/utils
```diff
@@ -247,6 +248,7 @@ Cluster settings::
   cni - $(typeset -p cluster_cni | cut -f 2- -d=)
   nodes per cluster - $(typeset -p cluster_nodes | cut -f 2- -d=)
   install submariner - $(typeset -p cluster_subm | cut -f 2- -d=)
+  label gateway nodes - $(typeset -p gateways | cut -f 2- -d=)
```
mkolesnik:
```suggestion
  label gateway nodes - $(typeset -p cluster_gateways | cut -f 2- -d=)
```

mkolesnik (changes requested):
Thanks it's much simpler now :), a few suggestions

scripts/shared/lib/deploy_funcs
MaxBab: @mkolesnik 

I looked closely into the github actions and to me it looks like no sense to place it there.
Let me explain my point of view.

I have created the "verify_gw_topology" to ensure that the desired state of the gateway nodes provided in `.shipyard.e2e.yml` or `.shipyard.e2e.ovn.yml` is actually applied.
The importance here is to make sure that if first cluster need to get one gw node and second cluster two gw nodes, that would be the state.

I placed this function near the "verify_gw_status" check function in scripts/shared/deploy.sh as it looked to me right.

If I place the check function within the github actions, there is no real verification of the desired gw state as I can't really set a different configuration per the cluster over there.
I think the live check during the deployment will add a bug value of env validation.

But if you think it should not be there, I can remove it completely.
My main point is that I don't think github action could really check and verify the that kind of flow.

What do you think?

scripts/shared/lib/utils
MaxBab: Done.

scripts/shared/cloud-prepare.sh
MaxBab: @mkolesnik 
I did not applied this suggestion due to the comment on the loop suggestion.

scripts/shared/cloud-prepare.sh
MaxBab: @mkolesnik 
I tried to apply this suggestion, but apparently it look like this way of looping, with control of the range in bash is working only for integers and not strings.

scripts/shared/lib/utils
MaxBab: Done.

scripts/shared/lib/utils
MaxBab: Done.

scripts/shared/cloud-prepare.sh
MaxBab: Done.
Applied.

scripts/shared/cloud-prepare.sh
MaxBab: Done

scripts/shared/cloud-prepare.sh
MaxBab: I figured out why it didn't work for me.

Done.
Very nice solution.

tpantelis (approved):

scripts/shared/lib/deploy_funcs
mkolesnik: You're missing the point entirely..

`verify_gw_status` makes sure theres an established connection before proceeding.
This is because if we don;t wait, there could be a race between the tests starting to run, and connections establishing, and sometimes the tests will entirely fail before the connection established.

You verification is a functional test, testing that when you request more than one gateway you get it.
This is clearly a test, and not any impediment to the testing itself (from shipyard perspective).

Hence, I suggest to put it as a test which tests this functionality, rather than here which doesn't make much sense.

In other words, say I have a function `sum(int, int) int` then I wouldn't put a test each time I call it in the code to test `sum(1,1) == 2`, right?

MaxBab: @mkolesnik 

Ok, so how could I set a different gateways count between two clusters in GHA in order to make a proper validation?
The configuration is done in `.shipyard*` files.

mkolesnik: You can create a test which loads a specific settings file. The real part you need to test is `cloud-prepare`  so you run `make clusters` to create them and then `make cloud-prepare` and see that each cluster has the requested number of nodes.

In fact you don't even need a concrete settings file, you can store file contents in your test script in variables, and write them out to a temp file (you can use `mktemp` for creating a temp file).
However if its easier just create the concrete file inside the test directory.

You don't even need multiple clusters. It's enough to create a single cluster with say 2 nodes and after each test unlabel all gateway nodes.

MaxBab: Done

.github/workflows/verify.yml
```diff
@@ -0,0 +1,20 @@
+---
```
mkolesnik: Why put it in a new file, and not as part of `scripts.yml`?

test/scripts/validation/scenario_config1
```diff
@@ -0,0 +1,6 @@
+---
```
mkolesnik: File name should end with `.yml` so it also gets linted.

Also why not call it according to the contents? I.E `no_gateways_specified.yml`

test/scripts/validation/scenario_config1
```diff
@@ -0,0 +1,6 @@
...
+submariner: true
+nodes: control-plane worker
+clusters:
+  cluster1:
+    nodes: control-plane worker worker
```
mkolesnik: Is it intentional that you specify `nodes` here (as well as globally)?

Only this will take effect

test/scripts/validation/scenario_config2
```diff
@@ -0,0 +1,7 @@
+---
```
mkolesnik: Same comments for this file (respectively)

test/scripts/validation/test.sh
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env bash
+
+set -e
+
+source ${SCRIPTS_DIR}/lib/debug_functions
+source ${SCRIPTS_DIR}/lib/utils
+# source ${SCRIPTS_DIR}/cloud-prepare.sh
```
mkolesnik:
```suggestion
```

test/scripts/validation/test.sh
```diff
@@ -0,0 +1,53 @@
...
+
+function verify_gw_topology() {
+    local gateways_count
+    gateways="${cluster_gateways[$cluster]:-1}"
```
mkolesnik:
```suggestion
    local expected_gateways=${cluster_gateways[$cluster]:-1}
```

test/scripts/validation/test.sh
```diff
@@ -0,0 +1,53 @@
...
+
+    echo "Verify gateway nodes"
+    gateways_count=$(kubectl get nodes -l=submariner.io/gateway=true \
+        --no-headers=true -o custom-columns=NAME:.metadata.name | wc -l)
+
+    if [[ "$gateways" -ne "$gateways_count" ]]; then
+        echo "Expect $gateways gateways nodes but detected $gateways_count"
```
mkolesnik:
```suggestion
        echo "Expected ${expected_gateways} gateways nodes but detected ${gateways_count}"
```

test/scripts/validation/test.sh
```diff
@@ -0,0 +1,53 @@
...
+
+source ${SCRIPTS_DIR}/lib/debug_functions
+source ${SCRIPTS_DIR}/lib/utils
+# source ${SCRIPTS_DIR}/cloud-prepare.sh
+
+function verify_gw_topology() {
+    local gateways_count
+    gateways="${cluster_gateways[$cluster]:-1}"
+
+    echo "Verify gateway nodes"
```
mkolesnik:
```suggestion
    echo "Verifying ${expected_gateways} gateway nodes"
```

test/scripts/validation/test.sh
```diff
@@ -0,0 +1,53 @@
...
+    gateways_count=$(kubectl get nodes -l=submariner.io/gateway=true \
+        --no-headers=true -o custom-columns=NAME:.metadata.name | wc -l)
+
+    if [[ "$gateways" -ne "$gateways_count" ]]; then
+        echo "Expect $gateways gateways nodes but detected $gateways_count"
+        return 1
+    else
+        echo "Found expected number of gateways - $gateways"
+    fi
```
mkolesnik:
```suggestion
    fi

    echo "Found expected number of gateways - $gateways"
```

test/scripts/validation/test.sh
```diff
@@ -0,0 +1,53 @@
...
+source ${SCRIPTS_DIR}/lib/utils
+# source ${SCRIPTS_DIR}/cloud-prepare.sh
+
+function verify_gw_topology() {
+    local gateways_count
+    gateways="${cluster_gateways[$cluster]:-1}"
+
+    echo "Verify gateway nodes"
+    gateways_count=$(kubectl get nodes -l=submariner.io/gateway=true \
+        --no-headers=true -o custom-columns=NAME:.metadata.name | wc -l)
```
mkolesnik:
```suggestion
    gateways_count=$(kubectl get nodes -o yaml | yq '.items | length')
```

test/scripts/validation/test.sh
```diff
@@ -0,0 +1,53 @@
...
+        --no-headers=true -o custom-columns=NAME:.metadata.name)
+    
+    for node in "${gw_nodes[@]}"; do
+        kubectl label --overwrite nodes "$node" submariner.io/gateway-
+    done
+}
+
+echo "Prepare cluster"
+make clusters SETTINGS=test/scripts/validation/scenario_config1
+
```
mkolesnik:
```suggestion
declare_kubeconfig

```

No need for these test to do each time

test/scripts/validation/test.sh
```diff
@@ -0,0 +1,53 @@
...
+for scenario in scenario_config1 scenario_config2; do
+    export SETTINGS="test/scripts/validation/$scenario"
+    echo "Set gateway nodes according to $scenario scenario"
+    make cloud-prepare
+
+    declare_kubeconfig
```
mkolesnik:
```suggestion
```

test/scripts/validation/test.sh
```diff
@@ -0,0 +1,53 @@
...
+    load_settings
+    clusters=($(kind get clusters))
```
mkolesnik:
```suggestion
```

`cllusters` is initialized from settings, any reason to get it from kind?

It could have problems in an environment where someone has kind clusters unrelated to submariner

test/scripts/validation/test.sh
```diff
@@ -0,0 +1,53 @@
...
+    run_parallel "${clusters[*]}" verify_gw_topology
+    run_parallel "${clusters[*]}" remove_gw_labels
```
mkolesnik:
```suggestion
    run_all_clusters verify_gw_topology
    run_all_clusters remove_gw_labels
```

test/scripts/validation/test.sh
```diff
@@ -0,0 +1,53 @@
...
+done
+
+echo "Perform cleanup"
+make cleanup SETTINGS=test/scripts/validation/scenario_config1
```
mkolesnik: This should be inside a trap function, to make sure test always cleans up even if it fails, otherwise youll have rogue clusters left behind on your environment

test/scripts/validation/scenario_config2
```diff
@@ -0,0 +1,7 @@
...
+submariner: true
+nodes: control-plane worker
+clusters:
+  cluster1:
```
mkolesnik:
```suggestion
  clusters-test:
```

Perhaps something like this, in case anyone has any kind clusters already deployed with the name `cluster1` on their environment.

test/scripts/validation/test.sh
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env bash
```
mkolesnik: The directory this in should probably be named `clusters` since youre testing cluster deployments

test/scripts/validation/test.sh
```diff
@@ -0,0 +1,53 @@
...
+        echo "Found expected number of gateways - $gateways"
+    fi
+}
+
+function remove_gw_labels() {
+    local gw_nodes
+
+    echo "Reset gateway nodes (unlabel)"
+    readarray -t gw_nodes < <(kubectl get nodes -l=submariner.io/gateway=true \
+        --no-headers=true -o custom-columns=NAME:.metadata.name)
```
mkolesnik:
```suggestion
    readarray -d' ' nodes < <(kubectl get nodes -o jsonpath="{.items[*].metadata.name}")
```

scripts/shared/cloud-prepare.sh
```diff
@@ -28,15 +28,19 @@ function check_gateway_exists() {
...
-    kubectl label node "$node" "$GATEWAY_LABEL" --overwrite
-
-    if [[ "$AIR_GAPPED" = true ]]; then
-        local pub_ip
-        pub_ip=$(kubectl get nodes "$node" -o jsonpath="{.status.addresses[0].address}")
-        kubectl annotate node "$node" gateway.submariner.io/public-ip=ipv4:"$pub_ip"
-    fi
+    local gw_count="${cluster_gateways[$cluster]:-1}"
+
+    readarray -t nodes < <(kubectl get nodes --no-headers -o custom-columns=NAME:".metadata.name" | tac)
```
mkolesnik:
```suggestion
    readarray -d' ' nodes < <(kubectl get nodes -o jsonpath="{.items[*].metadata.name}")
```

Or if you insist on reversing the order (even though it doesn't matter):
```suggestion
    readarray -t nodes < <(kubectl get nodes -o yaml | yq '.items[].metadata.name' | sort -r)
```

scripts/shared/cloud-prepare.sh
```diff
@@ -28,15 +28,19 @@ function check_gateway_exists() {
...
+
+    for node in "${nodes[@]:0:$gw_count}"; do
+        kubectl label node "$node" "$GATEWAY_LABEL" --overwrite
+
+        if [[ "$AIR_GAPPED" = true ]]; then
```
mkolesnik:
```suggestion
        [[ "$AIR_GAPPED" = true ]] || continue
```

test/scripts/validation/test.sh
```diff
@@ -0,0 +1,53 @@
...
+    
+    for node in "${gw_nodes[@]}"; do
+        kubectl label --overwrite nodes "$node" submariner.io/gateway-
+    done
+}
+
+echo "Prepare cluster"
+make clusters SETTINGS=test/scripts/validation/scenario_config1
```
mkolesnik:
```suggestion
make clusters SETTINGS="$(dirname $0)/scenario_config1"
```

test/scripts/validation/test.sh
```diff
@@ -0,0 +1,53 @@
...
+
+for scenario in scenario_config1 scenario_config2; do
+    export SETTINGS="test/scripts/validation/$scenario"
```
mkolesnik:
```suggestion
    export SETTINGS=$(dirname $0)/${scenario}
```

mkolesnik (changes requested):

test/scripts/validation/test.sh
MaxBab: Mike, you are suggested to call it "validation".

.github/workflows/verify.yml
MaxBab: Changed.

test/scripts/validation/scenario_config1
MaxBab: Changed.

test/scripts/validation/scenario_config1
MaxBab: Changed.

test/scripts/validation/test.sh
MaxBab: Removed

test/scripts/validation/test.sh
MaxBab: Changed.

test/scripts/validation/test.sh
MaxBab: Changed.

test/scripts/validation/test.sh
MaxBab: Done.

test/scripts/validation/test.sh
MaxBab: Done

test/scripts/validation/test.sh
MaxBab: I need here to filter by the gateway label.
Modified accordingly.

test/scripts/validation/test.sh
MaxBab: Done

test/scripts/validation/test.sh
MaxBab: Removed.

test/scripts/validation/test.sh
MaxBab: Updated

test/scripts/validation/test.sh
MaxBab: Moved to the workflow steps.

test/scripts/validation/scenario_config2
MaxBab: Updated.

test/scripts/validation/test.sh
MaxBab: Updated.

scripts/shared/cloud-prepare.sh
MaxBab: Updated.

scripts/shared/cloud-prepare.sh
MaxBab: Updated.

test/scripts/validation/test.sh
MaxBab: Updated

test/scripts/validation/test.sh
MaxBab: Updated

test/scripts/clusters/no_gw_defined.yml
```diff
@@ -0,0 +1,5 @@
+---
+submariner: true
```
mkolesnik:
```suggestion
```

Not needed for testing cloud prepare :)

test/scripts/clusters/two_gw_defined.yml
```diff
@@ -0,0 +1,6 @@
+---
+submariner: true
```
mkolesnik:
```suggestion
```

Not needed for testing cloud prepare :)

test/scripts/clusters/test.sh
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env bash
```
mkolesnik: Actually directory should probably be named `cloud-prepare` as you're focus is on that part of the process

.github/workflows/scripts.yml
```diff
@@ -112,3 +112,16 @@ jobs:
...
+
+  verify_gateways:
+    name: Verify gateways
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out the repository
+        uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8
+
+      - name: Deploy and validate gateways custom topology
+        run: make script-test SCRIPT_TEST_ARGS="test/scripts/clusters/test.sh"
```
mkolesnik:
```suggestion
      - name: Deploy clusters and test cloud-prepare
        run: make script-test SCRIPT_TEST_ARGS="test/scripts/cloud-prepare/test.sh"
```

test/scripts/clusters/test.sh
```diff
@@ -0,0 +1,48 @@
...
+
+    echo "Found expected number of gateways - ${expected_gateways}"
+}
+
+function remove_gw_labels() {
+    local gw_nodes
+
+    echo "Reset gateway nodes (unlabel)"
+    readarray -t gw_nodes < <(kubectl get nodes -l=submariner.io/gateway=true \
+        --no-headers=true -o custom-columns=NAME:.metadata.name)
```
mkolesnik:
```suggestion
    readarray -t gw_nodes < <(kubectl get nodes -l=submariner.io/gateway=true -o jsonpath="{.items[*].metadata.name}")
```

test/scripts/validation/test.sh
mkolesnik: sorry for driving you crazy with the names

test/scripts/clusters/test.sh
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env bash
+
+set -e
```
mkolesnik:
```suggestion
set -e -o pipefail
```

Otherwise it won't actually stop the test in case it fails, and will exit successfully eventually.

test/scripts/clusters/test.sh
```diff
@@ -0,0 +1,48 @@
...
+
+source ${SCRIPTS_DIR}/lib/debug_functions
+source ${SCRIPTS_DIR}/lib/utils
+
+function verify_gw_topology() {
+    local gateways_count
+    local expected_gateways="${cluster_gateways[$cluster]:-1}"
```
mkolesnik: I'm not sure if it's a good idea to rely on this being read from file, if the logic for reading is broken this test wont catch it.

It's better to pass this to the function from outside, since we know how many gateways we actually expect.

mkolesnik (changes requested):

test/scripts/clusters/no_gw_defined.yml
MaxBab: Actually it is required.
Look here - https://github.com/submariner-io/shipyard/blob/devel/scripts/shared/cloud-prepare.sh#L13

test/scripts/clusters/two_gw_defined.yml
MaxBab: Actually it is required.
Look here - https://github.com/submariner-io/shipyard/blob/devel/scripts/shared/cloud-prepare.sh#L13

test/scripts/clusters/test.sh
MaxBab: Changed.

.github/workflows/scripts.yml
MaxBab: Updated.

test/scripts/clusters/test.sh
MaxBab: Updated.

test/scripts/clusters/test.sh
MaxBab: Added

test/scripts/clusters/test.sh
MaxBab: Modified.

test/scripts/cloud-prepare/test.sh
```diff
@@ -0,0 +1,52 @@
...
+}
+
+echo "Prepare cluster"
+make clusters SETTINGS="$(dirname $0)/no_gw_defined.yml"
+declare_kubeconfig
+
+for scenario in no_gw_defined.yml two_gw_defined.yml; do
+    export SETTINGS="$(dirname $0)/${scenario}"
+
+    test_gateways=$(yq e '.gateways' $SETTINGS)
```
mkolesnik: This will only work if `gateways` is defined on the global level not on cluster level.
Also its still prone to reading from settings file, which is more fragile.

I propose to just call the test with hard coded values (1 and 2) instead of relying on reading from settings.
You can wrap L46-L51 in a function `test_gateways` or something like that

mkolesnik (changes requested):

test/scripts/cloud-prepare/test.sh
MaxBab: @mkolesnik 
But all the purpose of this test is to ensure that the configuration could be properly read from the config file.
That's why I think it's preferred to use the previous way of reading from file:

`local gw_count="${cluster_gateways[$cluster]:-1}"`

It's the most convenient as uses the actual deployment flow.

mkolesnik: When testing programatically it's better not to reuse the same underlying mechanism if possible, to avoid using a broken mechanic.

Say if I put now `gateways: 0` in the two gw file, the test will still pass, but the printout won't really help us understand the test is broken.

Since you know how many gateways you expect in each scenario, no reason not to provide it as a constant value.

MaxBab: @mkolesnik 

I'm not sure I'm following your suggestion.
You said that I can wrap L46-L51 in a function test_gateways or something like that.

Now, looking at the code.
The loop here (line 40), applies and test two scenarios.
The "make cloud-prepare" (line 47) applies the settings from the scenario files.
I don't touch it, since it's the actual test of the functionality.

So, you would like to hardcode the input values for the "verify_gw_topology" test in line 50.
But since it's running as part of the loop, I can't really change the hardcore values.

test/scripts/cloud-prepare/test.sh
```diff
@@ -0,0 +1,52 @@
...
+    [[ "${test_gateways}" == null ]] && test_gateways=1
+
+    echo "Set gateway nodes according to ${scenario} scenario"
+    make cloud-prepare
+
+    load_settings
+    run_all_clusters verify_gw_topology "${test_gateways}"
+    run_all_clusters remove_gw_labels
```
mkolesnik: Something like this

```suggestion
function test_gateways() {
    local scenario=$1
    local expected_gateways=$2

    echo "Set gateway nodes according to ${scenario} scenario"
    make cloud-prepare SETTINGS="$(dirname $0)/${scenario}"

    run_all_clusters verify_gw_topology "$expected_gateways"
    run_all_clusters remove_gw_labels
}

...

test_gateways no_gw_defined.yml 1
test_gateways two_gw_defined.yml 2
```

MaxBab: @mkolesnik 
Applied the suggestion.

Just added "load_settings" function to be able to connect to the clusters while checking the gateways.
And exporting the SETTINGS as environment variable.
Otherwise, the flow fails with the following error:
`ERROR: Expected env variable 'SETTINGS' can't be empty (Deployment settings file)`

mkolesnik (approved):

MaxBab:
@tpantelis 

Appreciate your review.
The patch has a few changes since your last approval.

skitt (approved):

dfarrell07:
@MaxBab Do you want to handle this backport (assuming you do actually want it on the release branches)?

MaxBab:
@dfarrell07 

Yes, please.
Which release it could be backported to?

dfarrell07:
@MaxBab `release-0.14` I think is where you introduced the relevant work?

dfarrell07:
@MaxBab Here are the relevant docs: https://submariner.io/development/backports/

dfarrell07:
Maybe also good, once there's a backport, to list it here: https://github.com/submariner-io/releases/issues/602

dfarrell07:
@MaxBab Ping, still planning on doing this backport?

MaxBab:
@dfarrell07 

Sorry Daniel.
Missed that out.
A lot of work.

I've created the backport and added it to the list.

Thank you.

PR #945, opened 2022-09-13
Modify gateway cleanup flow

MaxBab:
The gateway cleanup step is done after execution of gateways redundancy tests.
The cleanup is required for "kind" environment only, as for any other real environment, "MachineSet" is responsible for bringing the gateway node to its expected state.

- Due to the refactoring of the gateways redunadncy tests, modify the gateway cleanup flow to be executed only on "kind" environment. The gateway cleanup will look for the any node with "submariner.io/gateway: false" label and set the label to true.

- Modify the "findNodesByGatewayLabel" function to return only the nodes with existing gateway label. Skip the node with the missing labels.

Refactoring details:


test/e2e/framework/gateways.go
```diff
@@ -184,30 +183,18 @@ func (f *Framework) DeleteGateway(cluster ClusterIndex, name string) {
 	}, NoopCheckResult)
 }
 
-// GatewayCleanup ensures that only the active gateway node is flagged as gateway node which could not be after a failed
-//  test which left the system on an unexpected state.
-func (f *Framework) GatewayCleanup() {
+// KindGatewayCleanup will be executed only on kind environment.
+// It will restore the gateway nodes to its initial state.
+// Other environments do not need any gw cleanup as MachineSet is responsible to keeping the gw nodes in active states.
+func (f *Framework) KindGatewayCleanup() {
```
tpantelis: Minor thing but I think we can keep the function name as is since it internally checks if the environment is "kind". Having `Kind` in the function name implies the caller should ensure the environment is "kind" but that's not the case.

test/e2e/framework/nodes.go
```diff
@@ -28,8 +28,8 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 )
 
-// FindNodesByGatewayLabel finds the nodes in a given cluster by matching 'submariner.io/gateway' value. A missing label
-// is treated as false. Note the control plane node labeled as master is ignored.
+// FindNodesByGatewayLabel finds the nodes in a given cluster by matching 'submariner.io/gateway' value.
+// Nodes with the missing label will be ignored. Note the control plane node labeled as master is ignored.
```
tpantelis, line 32: Won't there be case(s) where the caller wants to treat a missing label as false? We could pass in a `*bool` where, if nil, treat a missing label as false.

tpantelis (changes requested):

test/e2e/framework/gateways.go
MaxBab: ack, makes sense.
Will change the name back.

test/e2e/framework/nodes.go
MaxBab: @tpantelis 
I though about adding additional label to the function.
But it looks to me like it's going to overload the function and make a bit complex to understand:
`findNodesByGatewayLabel(cluster, true)`
vs
`findNodesByGatewayLabel(cluster, true, false)`

In additional, the "findNodesByGatewayLabel" literally says that it finds a gateway node, while currently, if I'm searching for a gateway node with label "false", it returns to me all non GW worker.

The last thing, is that no test is actually looking for a gateway node with label "false".

What do you think?

tpantelis: In actuality, a non-GW is one w/o the label, ie a user wouldn't specifically label a node with "false". Is there a specific case where you want to differentiate between "false" and missing? 

[This caller](https://github.com/submariner-io/submariner/blob/ff8d073bdb04cbaedbaf4aecf60ff2b21c73249a/test/e2e/redundancy/gateway_failover.go#L143) and [this](https://github.com/submariner-io/submariner/blob/ff8d073bdb04cbaedbaf4aecf60ff2b21c73249a/test/e2e/redundancy/route_agent_restart.go#L50) are probably expect _missing label == false_.  But maybe you'll be changing those anyway.

MaxBab: Yes, the use case where I need to find a node with "false" label and not with missing, is the `GatewayCleanup` function defined within this patch.
It is used to locate and restore the gateway that set to "false" during GatewayFailover test on "kind" environment.
As we discussed, in "kind" I can't crash the node, because of will crash the machine that is running the "kind", so I'm adding a second GW node and during the test it simulates the failover (as it's working today).
So, after the test I need to identify that node and bring it back as "true" in gateway label.

Yes, I'm changing both the use cases you mentioned in links.

tpantelis: ok sounds good.

tpantelis (approved):

skitt (approved):

PR #946, opened 2022-09-13
Add DoFailover function to failover GW node of k8s/ocp cluster

MaxBab:
- The DoFailover function will perform a fail over of the gateway node depends on the environment. If the platform is "kind", the failover will happen by setting the gateway label to false. In case of any other environment, the function will crash the node.

  The use of crashing the node on "kind" env is not possible as it will crash the host as well.

- Add a function to detect the ProviderID value.

Details of the change:

Depends On https://github.com/submariner-io/shipyard/pull/963

### DIFF
--- a/test/e2e/framework/framework.go
+++ b/test/e2e/framework/framework.go
@@ -23,6 +23,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"strings"
 	"time"
 
 	. "github.com/onsi/gomega"
@@ -568,4 +569,11 @@ func NestedString(obj map[string]interface{}, fields ...string) string {
 	Expect(err).To(Succeed())
 
 	return str
+}
+
+func DetectProvider(cluster ClusterIndex, nodeName string) string {
+	node, err := KubeClients[cluster].CoreV1().Nodes().Get(context.TODO(), nodeName, metav1.GetOptions{})
+	Expect(err).NotTo(HaveOccurred())
+
+	return strings.Split(node.Spec.ProviderID, ":")[0]
 }
--- a/test/e2e/framework/gateways.go
+++ b/test/e2e/framework/gateways.go
@@ -198,3 +198,33 @@ func (f *Framework) GatewayCleanup() {
 		}
 	}
 }
+
+// Perform a gateway failover.
+// The failover for the real environment will crash the gateway node.
+// The failover for the kind environment will set the submariner.io/gateway label to "false" on the gw node.
+func (f *Framework) DoFailover(cluster ClusterIndex, gwNode, gwPod string) {
+	provider := DetectProvider(cluster, gwNode)
+
+	if provider == "kind" {
+		f.SaveGatewayNode(cluster, gwNode)
+		f.SetGatewayLabelOnNode(cluster, gwNode, false)
+	} else {
+		cmd := []string{"sh", "-c", "echo 1 > /proc/sys/kernel/sysrq && echo b > /proc/sysrq-trigger"}
+
+		_, _, err := f.ExecWithOptions(&ExecOptions{
+			Command:       cmd,
+			Namespace:     TestContext.SubmarinerNamespace,
+			PodName:       gwPod,
+			ContainerName: SubmarinerGateway,
+			CaptureStdout: false,
+			CaptureStderr: true,
+		}, cluster)
+		if err != nil {
+			if strings.Contains(err.Error(), "unable to upgrade connection: container not found") {
+				By(fmt.Sprintf("Successfully crashed gateway node %q", gwNode))
+			} else {
+				Expect(err).NotTo(HaveOccurred())
+			}
+		}
+	}
+}


test/e2e/framework/gateways.go
```diff
@@ -211,3 +211,28 @@ func gatewayNames(gateways []unstructured.Unstructured) []string {
...
+			Command:       cmd,
+			Namespace:     TestContext.SubmarinerNamespace,
+			PodName:       gwPod,
+			ContainerName: SubmarinerGateway,
+			CaptureStdout: false,
+			CaptureStderr: true,
+		}, cluster)
+		if err != nil {
+			Errorf("Gateway failover error %q", err)
+		}
```
tpantelis: It seems we should fail the test?

```suggestion
                 Expect(err).NotTo(HaveOccurred(), "Gateway failover error")
```

test/e2e/framework/framework.go
```diff
@@ -562,3 +563,9 @@ func NestedString(obj map[string]interface{}, fields ...string) string {
 
 	return str
 }
+
+func DetectProvider(cluster ClusterIndex, nodeName string) string {
+	node, _ := KubeClients[cluster].CoreV1().Nodes().Get(context.TODO(), nodeName, metav1.GetOptions{})
```
tpantelis: We really shouldn't ignore errors.

```suggestion
	node, err := KubeClients[cluster].CoreV1().Nodes().Get(context.TODO(), nodeName, metav1.GetOptions{})
	Expect(err).NotTo(HaveOccurred())
```

tpantelis (changes requested):

test/e2e/framework/framework.go
MaxBab: You're right.
Will fix it.

test/e2e/framework/gateways.go
MaxBab: @tpantelis 
Actually, there is a catch and a reason I'm not fails the test.
I'm sending the crash command to the pod and it crashes the node.
The function does not aware of which command I'm actually sending, so it tries to ensure that the command had been send properly.
But the problem is that the pod and node crashed, to it gives back an error that the pod has not been found.
So, it's like an expected error.

Maybe I can add a comment about it?
Or maybe you have some idea of how could I handle it?

tpantelis: yeah at least a comment. So I assume we may or may not get an error returned depending on timing or is one always returned? It would be ideal to identify the specific error that is expected and ignore that. Is that feasible?

MaxBab: I'mm add a comment.

Regarding the returned error, it always returned.
And yes, it's the same.
So you're saying that I can add a condition that if and error exists and does not contain specific error, fail the test.

Ok, will do.

MaxBab: Added a condition to check specific error.
Any other error will fail the test.

test/e2e/framework/framework.go
MaxBab: Added

test/e2e/framework/gateways.go
```diff
@@ -198,3 +198,32 @@ func (f *Framework) GatewayCleanup() {
...
+			Command:       cmd,
+			Namespace:     TestContext.SubmarinerNamespace,
+			PodName:       gwPod,
+			ContainerName: SubmarinerGateway,
+			CaptureStdout: false,
+			CaptureStderr: true,
+		}, cluster)
+		if err != nil {
+			if strings.HasPrefix(err.Error(), "unable to upgrade connection: container not found") {
+				fmt.Sprintln("Failover succeeded")
```
tpantelis:
```suggestion
                                 By(fmt.Sprintf("Successfully crashed gateway node %q", gwNode))         
```

test/e2e/framework/gateways.go
```diff
@@ -198,3 +198,32 @@ func (f *Framework) GatewayCleanup() {
...
+		_, _, err := f.ExecWithOptions(&ExecOptions{
+			Command:       cmd,
+			Namespace:     TestContext.SubmarinerNamespace,
+			PodName:       gwPod,
+			ContainerName: SubmarinerGateway,
+			CaptureStdout: false,
+			CaptureStderr: true,
+		}, cluster)
+		if err != nil {
+			if strings.HasPrefix(err.Error(), "unable to upgrade connection: container not found") {
```
tpantelis: Relying on a specific error message is a bit fragile. I'm wondering if we can check a specific  error type using the K8s error API functions, eg `apierrors.IsNotFound(err)`. Can you print out the full error with `%#v` formatting so we can see its type and other info?

tpantelis (changes requested):

test/e2e/framework/gateways.go
MaxBab: ack

test/e2e/framework/gateways.go
MaxBab: Will check that.

MaxBab: @tpantelis 
This is the output of the error type:
`&errors.errorString{s:"unable to upgrade connection: container not found (\"submariner-gateway\")"}`

Could you guide me on how to move on, please?

tpantelis: OK so it's not a specific error type we can key off of. So we'll have to check for a specific string as you're doing although I think we should use `strings.Contains` instead.

test/e2e/framework/gateways.go
MaxBab: Done

test/e2e/framework/gateways.go
MaxBab: Done

tpantelis (approved):

test/e2e/framework/gateways.go
```diff
@@ -198,3 +198,33 @@ func (f *Framework) GatewayCleanup() {
...
+		_, _, err := f.ExecWithOptions(&ExecOptions{
+			Command:       cmd,
+			Namespace:     TestContext.SubmarinerNamespace,
+			PodName:       gwPod,
+			ContainerName: SubmarinerGateway,
+			CaptureStdout: false,
+			CaptureStderr: true,
+		}, cluster)
+		if err != nil {
+			if strings.Contains(err.Error(), "unable to upgrade connection: container not found") {
```
sridhargaddam, line 223: Not sure if this message is correct. Is this a typo?

test/e2e/framework/gateways.go
```diff
@@ -198,3 +198,33 @@ func (f *Framework) GatewayCleanup() {
 		}
 	}
 }
+
+// Perform a gateway failover.
+// The failover for the real environment will crash the gateway node.
+// THe failover for the kind environment will set the "false" label on the gw node.
```
sridhargaddam: nit: 
```suggestion
// The failover for the kind environment will set the submariner.io/gateway label to "false" on the gw node.
```

test/e2e/framework/gateways.go
MaxBab: @sridhargaddam 
This message is correct.

I'm looking for a specific error message to identify it node crash succeeded.
Any other error message means that crash did not succeeded.

test/e2e/framework/gateways.go
MaxBab: ack
Will change.

test/e2e/framework/gateways.go
sridhargaddam: Okay, but what does "unable to upgrade connection" mean in this context? The purpose of this API is to trigger failover. So something like "unable to trigger failover" might be more apt.

test/e2e/framework/gateways.go
```diff
@@ -198,3 +198,33 @@ func (f *Framework) GatewayCleanup() {
...
+func (f *Framework) DoFailover(cluster ClusterIndex, gwNode, gwPod string) {
+	provider := DetectProvider(cluster, gwNode)
+
+	if provider == "kind" {
+		f.SaveGatewayNode(cluster, gwNode)
+		f.SetGatewayLabelOnNode(cluster, gwNode, false)
+	} else {
+		cmd := []string{"sh", "-c", "echo 1 > /proc/sys/kernel/sysrq && echo b > /proc/sysrq-trigger"}
```
sridhargaddam, line 212: After running this command, is the node getting restarted?

test/e2e/framework/gateways.go
MaxBab: @sridhargaddam 
What happens here, is that I'm sending the crash command to the pod which runs on the required node by using "ExecWithOptions" function.

The function sends the command and expect to receive ack from the pod.
But the node already crashed.
So, it tries to restore the connection.

This error message I get as response from the api.
It's not something I control.

test/e2e/framework/gateways.go
MaxBab: Yes, the node is crashed and restarted.
After boots up, it continue to work as expected.

test/e2e/framework/gateways.go
sridhargaddam: Oh ok. So you are comparing the error message. I felt that you were returning the error. Got it.

test/e2e/framework/gateways.go
MaxBab: Fixed

sridhargaddam (approved):

PR #947, opened 2022-09-14
Find multiple gateways

MaxBab:
As part of the gateway failover test refactoring, the test will look for a cluster that contains two gateway nodes.
Add a function that will search for two gateway nodes over available clusters and return selected cluster.

Refactoring details:

Depends On https://github.com/submariner-io/shipyard/pull/964/files

### DIFF
--- a/test/e2e/framework/nodes.go
+++ b/test/e2e/framework/nodes.go
@@ -31,6 +31,11 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 )
 
+const (
+	gatewayStatusLabel  = "gateway.submariner.io/status"
+	gatewayStatusActive = "active"
+)
+
 // FindGatewayNodes finds nodes in a given cluster by matching 'submariner.io/gateway' value.
 func FindGatewayNodes(cluster ClusterIndex) []v1.Node {
 	nodes, err := KubeClients[cluster].CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{
@@ -51,6 +56,19 @@ func FindNonGatewayNodes(cluster ClusterIndex) []v1.Node {
 	Expect(err).NotTo(HaveOccurred())
 
 	return nodes.Items
+}
+
+// FindClusterWithMultipleGateways finds the cluster with multiple GW nodes.
+// Returns cluster index.
+func (f *Framework) FindClusterWithMultipleGateways() int {
+	for idx := range TestContext.ClusterIDs {
+		gatewayNodes := FindGatewayNodes(ClusterIndex(idx))
+		if len(gatewayNodes) >= 2 {
+			return idx
+		}
+	}
+
+	return -1
 }
 
 // SetGatewayLabelOnNode sets the 'submariner.io/gateway' value for a node to the specified value.
--- a/test/e2e/framework/util.go
+++ b/test/e2e/framework/util.go
@@ -63,3 +63,15 @@ func NewRequirement(key string, op selection.Operator, vals []string) labels.Req
 
 	return *r
 }
+
+// FindOtherClusterIndex looks within the environment for another cluster
+// besides the one provided and returns its index or -1 if none other is found.
+func FindOtherClusterIndex(mainCluster int) int {
+	for idx := range TestContext.ClusterIDs {
+		if idx != mainCluster {
+			return idx
+		}
+	}
+
+	return -1
+}
--- a/test/e2e/framework/pods.go
+++ b/test/e2e/framework/pods.go
@@ -21,6 +21,7 @@ package framework
 import (
 	"context"
 	"fmt"
+	"time"
 
 	v1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -61,6 +62,42 @@ func (f *Framework) AwaitPodsByAppLabel(cluster ClusterIndex, appName, namespace
 // for the pod to materialize.
 func (f *Framework) AwaitSubmarinerGatewayPod(cluster ClusterIndex) *v1.Pod {
 	return &f.AwaitPodsByAppLabel(cluster, SubmarinerGateway, TestContext.SubmarinerNamespace, 1).Items[0]
+}
+
+// AwaitActiveGatewayPod looks for active gateway pod.
+// Returns pod object or nil.
+func (f *Framework) AwaitActiveGatewayPod(cluster ClusterIndex, checkPod func(*v1.Pod) bool) *v1.Pod {
+	for retries := 1; retries <= 30; retries++ {
+		var activePod, retPod *v1.Pod
+		gwPods := f.AwaitPodsByAppLabel(cluster, SubmarinerGateway, TestContext.SubmarinerNamespace, -1)
+
+		for i := range gwPods.Items {
+			pod := &gwPods.Items[i]
+			if pod.Labels[gatewayStatusLabel] == gatewayStatusActive {
+				if activePod != nil {
+					Errorf("Found 2 active gateway pods: %q and %q", activePod.Name, pod.Name)
+
+					retPod = nil
+
+					break
+				}
+
+				activePod = pod
+
+				if checkPod == nil || checkPod(pod) {
+					retPod = pod
+				}
+			}
+		}
+
+		if retPod != nil {
+			return retPod
+		}
+
+		time.Sleep(5 * time.Second)
+	}
+
+	return nil
 }
 
 // DeletePod deletes the pod for the given name and namespace.


test/e2e/framework/nodes.go
```diff
@@ -59,6 +59,32 @@ func findNodesByGatewayLabel(cluster int, isGateway bool) []*v1.Node {
 	return retNodes
 }
 
+func (f *Framework) FindMultipleGatewaysCluster(clusters []string) ([]int, bool) {
```
tpantelis: So this function finds one cluster with multiple GWs and returns it in an array as the first element. I assume this is so the caller can easily identify some other cluster if needed. However this wasn't intuitive to me at first. Perhaps we can simplify it to find and return just the cluster index with multiple GWs (or -1 if not found). If a caller needs to identify some other cluster then we can provide a helper to do so.

```suggestion
func (f *Framework) FindClusterWithMultipleGateways(clusters []string) int {
```

tpantelis: We really should pass the cluster indices to search otherwise the indices of the local `clusters` array may not align with the global indices `findNodesByGatewayLabel` expects.

That said, why not search all the globally defined clusters or will there be a use case where the caller will want to search a subset?

tpantelis (changes requested):

test/e2e/framework/nodes.go
MaxBab: @tpantelis 

Ok, I'm trying to follow.
You're saying that it's best to return an index of the cluster or if not clusters found, return "-1".

This is how I'm calling the function:
`clustersIndex, gatewaysExist := f.FindMultipleGatewaysCluster(framework.TestContext.ClusterIDs)`

Then I'm taking the first cluster from the list - `clusterIndex[0]`.
And since I'm in any case need only the first cluster, I can return only the first found cluster index.

Am I got it right?

tpantelis: Yes - returning an array with the remaining clusters as well is extraneous.

MaxBab: Great, will make the changes.
Thanks!

MaxBab: Hi @tpantelis 

Will appreciate another advice, please.
We agreed that the `FindMultipleGatewaysCluster` function will return an index of the single cluster with multiple GW nodes.

But I need identify the second cluster as well.
Meaning, if the cluster with 2 GW nodes has an index 1, I need to define the second cluster with index 0.
For the second cluster, it doesn't matter for me, what the amount of the GW nodes it has.

So, I thought about two possible solutions.
1) Create a function `findGatewayFailoverClusters`. That function will call for the `FindMultipleGatewaysCluster` in turn and when get the index of the cluster with multiple GW nodes, will loop through the "framework.TestContext.ClusterIDs" to find a second cluster which does not match the returned cluster index. Then it will return two variables. One for primaryCluster and one secondaryCluster.
2) Perform the loop within the body of the test to define the second cluster.

What do you think is preferred?
Or maybe any other way?

Thank you.

tpantelis: > Will appreciate another advice, please. We agreed that the `FindMultipleGatewaysCluster` function will return an index of the single cluster with multiple GW nodes.

Actually I notice in https://github.com/submariner-io/submariner/pull/2014 that you also need the GW node instances so rather than querying for those again we might as well return them here as well.

> 
> But I need identify the second cluster as well. Meaning, if the cluster with 2 GW nodes has an index 1, I need to define the second cluster with index 0. For the second cluster, it doesn't matter for me, what the amount of the GW nodes it has.
> 
> So, I thought about two possible solutions.
> 
> 1. Create a function `findGatewayFailoverClusters`. That function will call for the `FindMultipleGatewaysCluster` in turn and when get the index of the cluster with multiple GW nodes, will loop through the "framework.TestContext.ClusterIDs" to find a second cluster which does not match the returned cluster index. Then it will return two variables. One for primaryCluster and one secondaryCluster.
> 2. Perform the loop within the body of the test to define the second cluster.
> 
> What do you think is preferred? Or maybe any other way?
> 

I suggest option 2. You could add a helper function called `FindOtherClusterIndex` (put it in _util.go_).

MaxBab: Cool, thanks!

MaxBab: @tpantelis 

A question regarding the helper function, please.

If I'm adding a function to util.go within shipyard repo, it does not available from the GatewayFailover test in submariner repo.
The framework module already imported.
What am I missing?

Thank you.

tpantelis: Is the new function exported from the package, ie capitalized?

MaxBab: Yes.
It looks like the following:

`
func FindOtherClusterIndex(initialCluster int) {
	...func body...
}
`

skitt: If you want to use a new function provided by Shipyard in Submariner, you need to update the dependency on Shipyard (`go get github.com/submariner-io/shipyard@devel` once the PR is merged). To test locally, you need to add a local replace: `replace github.com/submariner-io/shipyard => ../shipyard` in Submariner’s `go.mod` (but make sure you don’t commit that).

MaxBab: @skitt 

Yes, I know.
I'm already using that way for local testing.
The strange thing is that any new function within the `framework.go`, `gateways.go` or `nodes.go` under shipyard, gets available in the submariner repo (with the "replace").
But the function under `util.go` is not recognized.

Of course, I;m doing a `go mod vendor`after any change in shipyard repo.

skitt: After your `go mod vendor`, is `vendor/github.com/submariner-io/shipyard/test/e2e/framework/util.go` in your Submariner repo updated to include your changes in Shipyard?

MaxBab: Yes, I can see it.

But I still go the following error.

```
# github.com/submariner-io/submariner/test/e2e/redundancy
test/e2e/redundancy/gateway_failover.go:148:23: f.FindOtherClusterIndex undefined (type *"github.com/submariner-io/submariner/test/e2e/framework".Framework has no field or method FindOtherClusterIndex)
note: module requires Go 1.18
FAIL	github.com/submariner-io/submariner/test/e2e [build failed]
FAIL
```

MaxBab: I also thought that the problem is that I used go 1.17.
I upgraded to go 1.18.6, and the error still the same.

skitt: Ah, you’re calling the function on a framework, but it’s not declared as a `Framework` function. If `FindOtherClusterIndex` needs information from `Framework`, you need to declare it as

    func (f *Framework) FindOtherClusterIndex(…) …

If it doesn’t, you need to call it as `framework.FindOtherClusterIndex(…)`, not `f.FindOtherClusterIndex(…)`.

MaxBab: Right!
I forgot that I already seen similar issue.

@skitt Thanks a lot for showing me the mistake.

test/e2e/framework/util.go
```diff
@@ -55,3 +55,15 @@ func ExpectNoErrorWithOffset(offset int, err error, explain ...interface{}) {
 
 	ExpectWithOffset(1+offset, err).NotTo(HaveOccurred(), explain...)
 }
+
+// FindOtherClusterIndex looks within the environment for the
+// other cluster than provided to the function.
```
tpantelis:
```suggestion
// FindOtherClusterIndex looks within the environment for another cluster
// besides the one provided and returns its index or -1 if none other is found.
```

test/e2e/framework/nodes.go
```diff
@@ -59,6 +65,31 @@ func findNodesByGatewayLabel(cluster int, isGateway bool) []*v1.Node {
 	return retNodes
 }
 
+// FindMultipleGatewaysCluster finds the cluster with multiple GW nodes.
+// Returns cluster index and active GW Pod and Node names.
+func (f *Framework) FindMultipleGatewaysCluster(clusters []string) (int, map[string]string) {
```
tpantelis:
```suggestion
func (f *Framework) FindClusterWithMultipleGateways(clusters []string) (int, map[string]string) {
```

tpantelis: Also, we really should pass the cluster indices to search otherwise the indices of the local clusters array may not align with the global indices `findNodesByGatewayLabel` expects.

That said, why not search all the globally defined clusters instead of passing in the clusters to search?

test/e2e/framework/nodes.go
```diff
@@ -59,6 +65,31 @@ func findNodesByGatewayLabel(cluster int, isGateway bool) []*v1.Node {
...
+			gwPods := f.AwaitPodsByAppLabel(ClusterIndex(cluster), SubmarinerGateway, TestContext.SubmarinerNamespace, -1)
+
+			for i := range gwPods.Items {
+				if gwPods.Items[i].Labels[gatewayStatusLabel] == gatewayStatusActive {
+					activeGW["Pod"] = gwPods.Items[i].Name
+					activeGW["Node"] = gwPods.Items[i].Labels[gatewayNodeLabel]
+
+					return cluster, activeGW
+				}
+			}
```
tpantelis: It's cleaner to return the pod and node directly rather than embedded in a map with well-known keys. However, perhaps we shouldn't overload this function to return the active GW as well. Instead I suggest we add an `AwaitActiveGatewayPod` function (in _pods.go_):

```Go
func (f *Framework) AwaitActiveGatewayPod(cluster ClusterIndex, checkPod func(*v1.Pod) bool) *v1.Pod {
    for retries := 1; retries <= 30; retries++ {
        var activePod, retPod *v1.Pod

        gwPods := f.AwaitPodsByAppLabel(ClusterIndex(cluster), SubmarinerGateway, TestContext.SubmarinerNamespace, -1)
	for i := range gwPods.Items {
	    pod := &gwPods.Items[i]
	    if pod.Labels[gatewayStatusLabel] == gatewayStatusActive {
	        if activePod != nil {
		    Errorf("Found 2 active gateway pods: %q and %q", activePod.Name, pod.Name)
		    retPod = nil
		    break
		}

		activePod = pod

		if checkPod == nil || checkPod(pod) {
		    retPod = pod
		}
	    }
        }

        if retPod != nil {
            return retPod
        }

        time.Sleep(2 * time.Second)
    }

    return nil
}
```

Then https://github.com/submariner-io/submariner/pull/2014 can call this to get the initial active GW:

```Go
primaryCluster := f.FindMultipleGatewaysCluster(framework.TestContext.ClusterIDs)
...
initialGWPod := f.AwaitActiveGatewayPod(primaryCluster, nil)
Expect(initialGWPod).ToNot(BeNil(), "Did not find an active gateway pod")
```

And to get the new GW pod after failover:

```Go
newGWPod := f.AwaitActiveGatewayPod(primaryCluster, func(pod *v1.Pod) bool {
    return pod.Spec.NodeName != initialGWPod.Spec.NodeName
})
```

Here we just return in cluster index.

tpantelis (changes requested):

test/e2e/framework/nodes.go
MaxBab: @tpantelis 

In the current function I'm also returning the GW node name.
Is it ok to return the gw node name in `AwaitActiveGatewayPod` function or better to create a dedicated function that will return the node name of gw node?

tpantelis: The node name can be extracted from the `Pod`, ie `pod.Spec.NodeName`.

MaxBab: Got it.
Thanks.

test/e2e/framework/nodes.go
MaxBab: @tpantelis 

I just would like to make sure I understand you correctly.
You're saying that instead of passing the `framework.TestContext.ClusterIDs` to the function, better to loop over the `framework.TestContext.ClusterIDs` within the cluster.
And the `framework.TestContext.ClusterIDs` is the globally defined clusters.

Am I get it right?

tpantelis: yes just loop over the global `framework.TestContext.ClusterIDs`.

test/e2e/framework/util.go
MaxBab: Done

test/e2e/framework/nodes.go
MaxBab: Done.

test/e2e/framework/nodes.go
MaxBab: Thank you for the idea and example.
That's a great code re use.

tpantelis (approved):

test/e2e/framework/nodes.go
```diff
@@ -59,6 +64,19 @@ func findNodesByGatewayLabel(cluster int, isGateway bool) []*v1.Node {
 	return retNodes
 }
 
+// FindClusterWithMultipleGateways finds the cluster with multiple GW nodes.
+// Returns cluster index and active GW Pod and Node names.
+func (f *Framework) FindClusterWithMultipleGateways() int {
+	for idx := range TestContext.ClusterIDs {
+		gatewayNodes := findNodesByLabel(idx, GatewayLabel, true)
```
tpantelis:
```suggestion
		gatewayNodes := FindGatewayNodes(idx)
```

tpantelis (changes requested):

test/e2e/framework/nodes.go
MaxBab: Done

test/e2e/framework/nodes.go
```diff
@@ -59,6 +64,19 @@ func findNodesByGatewayLabel(cluster int, isGateway bool) []*v1.Node {
 	return retNodes
 }
 
+// FindClusterWithMultipleGateways finds the cluster with multiple GW nodes.
+// Returns cluster index and active GW Pod and Node names.
```
sridhargaddam:
```suggestion
// Returns cluster index
```

sridhargaddam (commented):
One small observation, otherwise looks good.

test/e2e/framework/nodes.go
MaxBab: Done

sridhargaddam (approved):

tpantelis (approved):

PR #948, opened 2022-09-14
Find Non Gateway RouteAgent pod nodes

MaxBab:
Due to the gateway failover refactoring, "kind" environment will have all worker nodes serve as gateway nodes.
This is done to save the resources of the environment running "kind".

Because of that, the test will not be able to find a non gateway worker and will fail.
Add a function to look for the non gateway route agent node in case a non gateway node is not found.

This change is required only for the "kind" env.

Details of the refactoring:

### DIFF
--- a/test/e2e/framework/nodes.go
+++ b/test/e2e/framework/nodes.go
@@ -68,3 +68,37 @@ func (f *Framework) SetGatewayLabelOnNode(cluster ClusterIndex, nodeName string,
 			return err
 		})
 }
+
+// FindAnyNonGatewayRouteAgentPodNodes looks for the route agent pod on any cluster node, including control plane
+// The function will be used in upstream ci environment, lack of resources where two GW nodes are used.
+func (f *Framework) FindAnyNonGatewayRouteAgentPodNodes(cluster ClusterIndex) []*v1.Node {
+	gwNodesSet := make(map[string]bool)
+	gwNodes := f.FindNodesByGatewayLabel(cluster, true)
+
+	for _, node := range gwNodes {
+		gwNodesSet[node.Name] = true
+	}
+
+	var selectedNode string
+	routeAgentPods := f.AwaitPodsByAppLabel(cluster, RouteAgent, TestContext.SubmarinerNamespace, -1)
+
+	for i := range routeAgentPods.Items {
+		if !gwNodesSet[routeAgentPods.Items[i].Spec.NodeName] {
+			selectedNode = routeAgentPods.Items[i].Spec.NodeName
+		}
+	}
+
+	searchNodes := AwaitUntil("list nodes", func() (interface{}, error) {
+		// Ignore the control plane node labeled as master as it doesn't allow scheduling of pods
+		return KubeClients[cluster].CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{
+			LabelSelector: "kubernetes.io/hostname=" + selectedNode,
+		})
+	}, NoopCheckResult).(*v1.NodeList)
+
+	nodes := []*v1.Node{}
+	for i := range searchNodes.Items {
+		nodes = append(nodes, &searchNodes.Items[i])
+	}
+
+	return nodes
+}


go.mod
```diff
@@ -7,6 +7,7 @@ require (
 	github.com/onsi/gomega v1.20.2
 	github.com/pkg/errors v0.9.1
 	github.com/psampaz/go-mod-outdated v0.8.0
+	github.com/submariner-io/admiral v0.13.0
```
tpantelis: **shipyard** cannot depend on **admiral** - that would create a circular dependency.

test/e2e/framework/nodes.go
```diff
@@ -68,3 +69,37 @@ func (f *Framework) SetGatewayLabelOnNode(cluster ClusterIndex, nodeName string,
...
+			LabelSelector: "kubernetes.io/hostname=" + selectedNode,
+		})
+	}, NoopCheckResult).(*v1.NodeList)
+
+	nodes := []*v1.Node{}
+	for i := range searchNodes.Items {
+		nodes = append(nodes, &searchNodes.Items[i])
+	}
+
+	return nodes
```
tpantelis, line 103: You can just return `searchNodes` here to simplify, ie instead of returning `[]*v1.Node`.

```suggestion
	return AwaitUntil("list nodes", func() (interface{}, error) {
		// Ignore the control plane node labeled as master as it doesn't allow scheduling of pods
		return KubeClients[cluster].CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{
			LabelSelector: "kubernetes.io/hostname=" + selectedNode,
		})
	}, NoopCheckResult).(*v1.NodeList)
```

test/e2e/framework/nodes.go
```diff
@@ -68,3 +69,37 @@ func (f *Framework) SetGatewayLabelOnNode(cluster ClusterIndex, nodeName string,
 			return err
 		})
 }
+
+// FindAnyNonGatewayRouteAgentPodNodes looks for the route agent pod on any cluster node, including control plane
+// The function will be used in upstream ci environment, lack of resources where two GW nodes are used.
+func (f *Framework) FindAnyNonGatewayRouteAgentPodNodes(cluster ClusterIndex) []*v1.Node {
+	gwNodesSet := stringset.New()
```
tpantelis: You can simply use a map here instead of a set. 
 
```suggestion
	gwNodesSet := make(map[string]bool)
```

tpantelis (changes requested):

go.mod
MaxBab: Got it.
Will remove.

test/e2e/framework/nodes.go
MaxBab: ack

test/e2e/framework/nodes.go
MaxBab: You're right.
Thanks for the idea.

test/e2e/framework/nodes.go
MaxBab: Hi @tpantelis 

Regarding your suggestion to return `*v1.NodeList` instead of `[]*v1.Node`.

The reason I'm returning `[]*v1.Node` data structure is because in within the RouteAgentRestart test, I'm doing the following flow.
```
	nodes := f.FindNodesByGatewayLabel(framework.ClusterA, onGateway)
	// This will happen only in "kind" environment because of the flow of
	// testing gateway failover test
	if len(nodes) == 0 && !onGateway {
		By(fmt.Sprintf("No Non Gateway worker nodes found in %q. Looking for non worker nodes", clusterAName))
		nodes = f.FindAnyNonGatewayRouteAgentPodNodes(framework.ClusterA)
	}
```

And since I'm rewriting the `nodes` variable and `f.FindNodesByGatewayLabel` function returns `[]*v1.Node` data structure, I need to return the same structure.

What do you think, please?

tpantelis: OK - that would make it consistent.

go.mod
MaxBab: Done

test/e2e/framework/nodes.go
MaxBab: Replaced.

MaxBab:
Closing as all the necessary changes will be handled by the following patches:
https://github.com/submariner-io/shipyard/pull/964
https://github.com/submariner-io/shipyard/pull/998

PR #952, opened 2022-09-19
Simplify scripts GHA matrix

mkolesnik:
Instead of defining a new key for each extra toggle we want to use, define a single one and use it for the various combinations.


skitt (approved):

tpantelis (approved):

dfarrell07 (commented):
Nice idea, thanks!

PR #953, opened 2022-09-19
Add load balancer deployment support

mkolesnik:
This makes it easy to deploy a test environment with the load balancer, by calling e.g. `make deploy using=load-balancer`.

For kind based deployments, `metallb` is deployed and used.

Depends on #952 

### DIFF
--- a/release-notes/20220919-load-balancer.md
+++ b/release-notes/20220919-load-balancer.md
@@ -1,0 +1,5 @@
+<!-- markdownlint-disable MD041 -->
+Support was added in the Shipyard project to easily deploy Submariner with a LoadBalancer type Service in front.
+To use, simply specify the target (e.g. `deploy`) with `USING=load-balancer` or `LOAD_BALANCER=true`.
+For kind-based deployments, [MetalLB](https://metallb.universe.tf/) is deployed to provide the capability.
+The MetalLB version can be specified using `METALLB_VERSION=x.y.z`.
--- a/scripts/shared/lib/clusters_kind
+++ b/scripts/shared/lib/clusters_kind
@@ -111,10 +111,15 @@ function provider_create_cluster() {
     "${kind}" create cluster ${image_flag:+"$image_flag"} --name="${cluster}" --config="${RESOURCES_DIR}/${cluster}-config.yaml"
     kind_fixup_config
 
-    ( deploy_cni; ) &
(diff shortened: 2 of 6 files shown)


release-notes/20220919-load-balancer.md
```diff
@@ -0,0 +1,5 @@
+<!-- markdownlint-disable MD041 -->
+Load balancer support was added to easily deploy with Submariner with the load balancer in front.
+To use, simply use the target (e.g. `deploy`) with `USING=load-balancer` or `LOAD_BALANCER=true`.
+For kind based deployments, [metallb](https://metallb.universe.tf/) is deployed to provide the capability.
+The version can be specified using `METALLB_VERSION=x.y.z`.
```
nyechiel:
```suggestion
Support was added in the Shipyard project to easily deploy Submariner with a LoadBalancer type Service in front.
To use, simply specify the target (e.g. `deploy`) with `USING=load-balancer` or `LOAD_BALANCER=true`.
For kind-based deployments, [MetalLB](https://metallb.universe.tf/) is deployed to provide the capability.
The MetalLB version can be specified using `METALLB_VERSION=x.y.z`.
```

scripts/shared/lib/clusters_kind
```diff
@@ -203,6 +208,32 @@ function deploy_ovn_cni(){
     echo "OVN CNI deployed."
 }
 
+function deploy_load_balancer() {
+    local kind_ip start_ip end_ip
```
yboaron: If LB is deployed on 2 clusters , IIUC both clusters will use the same IPAM range x1.x2.x3.100 - x1.x2.x3.250

We should have separate range for each cluster

mkolesnik: Yes, thats why cluster CIDR is used in the IP range, though now that I think of it probably better to use the GN CIDR..

mkolesnik: I updated so that GN CIDR is used when GN enabled, but generally your concern has already been addressed.

skitt (approved):

dfarrell07 (approved):

nyechiel (approved):

nyechiel:
@mkolesnik I will let you merge this; can you also update the website docs?

PR #959, opened 2022-09-23
Increase duration while checking gw connections

sridhargaddam:
Sometimes it was seen that public-ip resolution using api.ipify.org was timing out (default interval is 30s). Because of this local endpoint is created with some delay and the side-effect of this issue is that connections take slightly longer time to get established. Currently, in our scripts we wait for 3 mins after installing Submariner on the clusters and if connections are not established within this duration the CI job is marked as failed.

However, looking at the post-mortem scripts it was seen that connections were all successfully established after a while. So, this PR increases the wait interval from 3 mins to 4 mins.

### DIFF
--- a/scripts/shared/lib/deploy_funcs
+++ b/scripts/shared/lib/deploy_funcs
@@ -81,7 +81,7 @@ function connectivity_tests() {
 }
 
 function verify_gw_status() {
-    sleep_duration=6
+    sleep_duration=8
     # helm doesn't use the operator yet, and connection status is based on the operator object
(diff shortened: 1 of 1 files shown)


sridhargaddam:
This PR will fix the flaky issues seen with some of the CI jobs. If we continue to see similar failures with api.ipify.org going forward, we can either remove it from the list of resolvers or move it to the end of the resolver chain.

tpantelis (approved):

nyechiel (approved):

PR #961, opened 2022-09-29
Add moreutils to our base image

skitt:
This provides sponge, which is useful for system tests where we want to see *and* check the output, e.g.

    subctl ... | tee /dev/stderr | sponge | grep -q 'expected text'

This copies the output to standard error, accumulates all the output sent to standard output (so that subctl finishes running before the pipeline terminates), and filters the output through grep. The output copied to standard error shows up unchanged in the logs, and the outcome of the whole pipeline is determined by the presence or absence of 'expected text' in the output. Without sponge, if the output is matched, the pipeline terminates immediately, which results in truncated output from subctl.


nyechiel (approved):

dfarrell07 (approved):
Interesting, hadn't seen that tool, very cool!

tpantelis (approved):

PR #964, opened 2022-10-03
Introduce "FindGatewayNodes" and "FindNonGatewayNodes" functions

MaxBab:
In order to make the function meaning cleaner, refactor the "findNodesByGatewayLabel" function.

- Create a "findNodesByLabel" function that is able to search for the nodes within the cluster by defined lablel and return the nodes with the label or without by defining a boolean.

- Create a "FindGatewayNodes" and "FindNonGatewayNodes" functions that will search for the relevant nodes - Gateway and NonGateway.

The patch is a part of the following effort:

### DIFF
--- a/test/e2e/framework/framework.go
+++ b/test/e2e/framework/framework.go
@@ -290,7 +290,7 @@ func InitNumClusterNodes() error {
 
 func fetchClusterIDs() {
 	for i := range KubeClients {
-		gatewayNodes := findNodesByGatewayLabel(i, true)
+		gatewayNodes := FindGatewayNodes(ClusterIndex(i))
 		if len(gatewayNodes) == 0 {
 			continue
 		}
--- a/test/e2e/framework/nodes.go
+++ b/test/e2e/framework/nodes.go
@@ -23,40 +23,34 @@ import (
 	"strconv"
 	"strings"
 
+	. "github.com/onsi/gomega"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/selection"
 	"k8s.io/apimachinery/pkg/types"
 )
 
-// FindNodesByGatewayLabel finds the nodes in a given cluster by matching 'submariner.io/gateway' value.
-// Nodes with the missing label will be ignored. Note the control plane node labeled as master is ignored.
-func (f *Framework) FindNodesByGatewayLabel(cluster ClusterIndex, isGateway bool) []*v1.Node {
-	return findNodesByGatewayLabel(int(cluster), isGateway)
+// FindGatewayNodes finds nodes in a given cluster by matching 'submariner.io/gateway' value.
+func FindGatewayNodes(cluster ClusterIndex) []v1.Node {
+	nodes, err := KubeClients[cluster].CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{
+		LabelSelector: labels.Set{GatewayLabel: "true"}.String(),
+	})
+	Expect(err).NotTo(HaveOccurred())
+
+	return nodes.Items
 }
 
-func findNodesByGatewayLabel(cluster int, isGateway bool) []*v1.Node {
-	nodes := AwaitUntil("list nodes", func() (interface{}, error) {
-		// Ignore the control plane node labeled as master as it doesn't allow scheduling of pods
-		return KubeClients[cluster].CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{
-			LabelSelector: "!node-role.kubernetes.io/master",
-		})
-	}, NoopCheckResult).(*v1.NodeList)
+// FindNonGatewayNodes finds nodes in a given cluster that doesn't match 'submariner.io/gateway' value.
+func FindNonGatewayNodes(cluster ClusterIndex) []v1.Node {
+	nodes, err := KubeClients[cluster].CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{
+		LabelSelector: labels.NewSelector().Add(
+			NewRequirement(GatewayLabel, selection.Exists, []string{}),
+			NewRequirement(GatewayLabel, selection.NotEquals, []string{"true"})).String(),
+	})
+	Expect(err).NotTo(HaveOccurred())
 
-	expLabelValue := strconv.FormatBool(isGateway)
-	retNodes := []*v1.Node{}
-
-	for i := range nodes.Items {
-		value, exists := nodes.Items[i].Labels[GatewayLabel]
-		if !exists {
-			continue
-		}
-
-		if value == expLabelValue {
-			retNodes = append(retNodes, &nodes.Items[i])
-		}
-	}
-
-	return retNodes
+	return nodes.Items
 }
 
 // SetGatewayLabelOnNode sets the 'submariner.io/gateway' value for a node to the specified value.
--- a/test/e2e/framework/util.go
+++ b/test/e2e/framework/util.go
@@ -22,8 +22,9 @@ import (
 	"strings"
 
 	. "github.com/onsi/gomega"
-	// We need GCP authentication.
-	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/selection"
+	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp" // We need GCP authentication.
 	restclient "k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
 )
@@ -54,4 +55,11 @@ func ExpectNoErrorWithOffset(offset int, err error, explain ...interface{}) {
 	}
 
 	ExpectWithOffset(1+offset, err).NotTo(HaveOccurred(), explain...)
+}
+
+func NewRequirement(key string, op selection.Operator, vals []string) labels.Requirement {
+	r, err := labels.NewRequirement(key, op, vals)
+	Expect(err).To(Succeed())
+
+	return *r
 }


test/e2e/framework/framework.go
```diff
@@ -285,7 +285,7 @@ func InitNumClusterNodes() error {
 
 func fetchClusterIDs() {
 	for i := range KubeClients {
-		gatewayNodes := findNodesByGatewayLabel(i, true)
+		gatewayNodes := findNodesByLabel(i, GatewayLabel, true)
```
tpantelis:
```suggestion
		gatewayNodes := FindGatewayNodes(i)
```

test/e2e/framework/nodes.go
```diff
@@ -28,30 +28,31 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 )
 
-// FindNodesByGatewayLabel finds the nodes in a given cluster by matching 'submariner.io/gateway' value.
-// Nodes with the missing label will be ignored. Note the control plane node labeled as master is ignored.
-func (f *Framework) FindNodesByGatewayLabel(cluster ClusterIndex, isGateway bool) []*v1.Node {
-	return findNodesByGatewayLabel(int(cluster), isGateway)
+// FindGatewayNodes finds nodes in a given cluster by matching 'submariner.io/gateway' value.
+func (f *Framework) FindGatewayNodes(cluster ClusterIndex) []*v1.Node {
+	return findNodesByLabel(int(cluster), GatewayLabel, true)
```
tpantelis: We really don't need the `findNodesByLabel` function, we can simply use a `LabelSelector`:

```suggestion
	return return KubeClients[cluster].CoreV1().Nodes().List(context.TODO(), 
	    metav1.ListOptions{
	               LabelSelector: labels.Set{GatewayLabel: "true"}.String(),
	           })
```

test/e2e/framework/nodes.go
```diff
@@ -28,30 +28,31 @@ import (
...
 }
 
-func findNodesByGatewayLabel(cluster int, isGateway bool) []*v1.Node {
+// FindGatewayNodes finds nodes in a given cluster that doesn't match 'submariner.io/gateway' value.
+func (f *Framework) FindNonGatewayNodes(cluster ClusterIndex) []*v1.Node {
+	return findNodesByLabel(int(cluster), GatewayLabel, false)
```
tpantelis: Similarly, using a `LabelSelector`:

```suggestion
	return KubeClients[cluster].CoreV1().Nodes().List(context.TODO(), 
	    metav1.ListOptions{
	               LabelSelector: labels.NewSelector().Add(
				newRequirement(GatewayLabel, selection.Exists, []string{}),
				newRequirement(GatewayLabel, selection.NotEquals, []string{"true"})).String(),
	           })
```

New helper function:

```Go
func newRequirement(key string, op selection.Operator, vals []string) labels.Requirement {
	r, err := labels.NewRequirement(key, op, vals)
	Expect(err).To(Succeed())

	return *r
}
```

tpantelis (changes requested):

test/e2e/framework/nodes.go
MaxBab: @tpantelis 

I have a question regarding your suggestion.
If we are using the `LabelSelector` instead of the `findNodesByLabel` function, it means that we are changing the return structure.

It will change from `[]*v1.Node` to `*v.NodeList`.
So in each place and test with the reference to the function will need to change.
For example:
https://github.com/submariner-io/submariner/blob/devel/test/e2e/redundancy/gateway_failover.go#L68-L70
https://github.com/submariner-io/submariner/blob/devel/test/e2e/redundancy/gateway_failover.go#L76

Will change from `gatewayNodes[0].Name` to `gatewayNodes.Items[0].Name`.

Is it acceptable for you?

tpantelis: You can return NodeList.Items.

MaxBab: Didn't think about it.
Thank you for the idea.

test/e2e/framework/framework.go
MaxBab: @tpantelis 

I faced another issue.
I have applied the below suggestion:
```
func (f *Framework) FindGatewayNodes(cluster ClusterIndex) []v1.Node {
		LabelSelector: labels.Set{GatewayLabel: "true"}.String(),
	})

	return nodes.Items
}
```


The problem is that the `fetchClusterIDs` function that tries to call the `FindGatewayNodes` is yet initialized with the test framework.
So I can't really call it like `f.FindGatewayNodes(i)` or `framework.FindGatewayNodes(i)`.
Currently, it uses a regular function without the framework definition.

What do you think?

skitt: You’re not actually using `f` in the function, so you can declare it as

```
func FindGatewayNodes(cluster ClusterIndex) []v1.Node {
```

Does that fix things?

MaxBab: @skitt 

The thing is that this function is widely used in the `submariner` repo tests, so I need to use the framework to defined it within the test.

tpantelis: > 
> The thing is that this function is widely used in the `submariner` repo tests

`FindGatewayNodes`? That function is new with this PR.

MaxBab: You're right.

The `FindGatewayNodes` and `FindNonGatewayNodes` functions replaces the following functions:
`func findNodesByGatewayLabel(cluster int, isGateway bool) []*v1.Node`
and
`func (f *Framework) FindNodesByGatewayLabel(cluster ClusterIndex, isGateway bool)`
which calls it.

In https://github.com/submariner-io/shipyard/blob/devel/test/e2e/framework/framework.go#L291, the call of the old function needs to be replaced with the new one.
`func (f *Framework) FindGatewayNodes(cluster ClusterIndex) []v1.Node`

But since it's a framework based function, I can't call it.

I can create a base function and call it with the new ones as it's done currently.

tpantelis: `FindGatewayNodes` can be stand alone as @skitt mentioned:

`func FindGatewayNodes(cluster ClusterIndex) []v1.Node`

MaxBab: @tpantelis @skitt 

Please, let me try to clarify the state of the issue.
Current patch proposes refactoring of the `findNodesByGatewayLabel` function.

Current implementation consists of two functions.
The first one is a base function:
`func findNodesByGatewayLabel(cluster int, isGateway bool) []*v1.Node` [link](https://github.com/submariner-io/shipyard/blob/devel/test/e2e/framework/nodes.go#L37)
This function is standalone and used within the `fetchClusterIDs` function [link](https://github.com/submariner-io/shipyard/blob/devel/test/e2e/framework/framework.go#L291).

The second, is a framework based function:
`func (f *Framework) FindNodesByGatewayLabel(cluster ClusterIndex, isGateway bool) []*v1.Node` [link](https://github.com/submariner-io/shipyard/blob/devel/test/e2e/framework/nodes.go#L33)
This function uses the base function described above with some input parameters.
This function used within the shipyard repo in the following place [link] (https://github.com/submariner-io/shipyard/blob/devel/test/e2e/framework/gateways.go#L191)

Beside the shipyard repo, the `FindNodesByGatewayLabel` function is used within the tests of `submariner` repository and called and test framework initialized function.
* https://github.com/submariner-io/submariner/blob/devel/test/e2e/redundancy/gateway_failover.go#L68
* https://github.com/submariner-io/submariner/blob/devel/test/e2e/redundancy/gateway_failover.go#L143
* https://github.com/submariner-io/submariner/blob/devel/test/e2e/redundancy/gateway_failover.go#L68
* https://github.com/submariner-io/submariner/blob/devel/test/e2e/redundancy/route_agent_restart.go#L50
* https://github.com/submariner-io/submariner/blob/devel/test/e2e/cluster/add_remove_cluster.go#L39
* https://github.com/submariner-io/submariner/blob/devel/test/e2e/cluster/add_remove_cluster.go#L61


With new changes, those calls needs to be replaced with the new functions from this patch, like:
```
func (f *Framework) FindGatewayNodes(cluster ClusterIndex) []v1.Node {
		LabelSelector: labels.Set{GatewayLabel: "true"}.String(),
	})

	return nodes.Items
}
```

This function does not rely on any standalone function.
I can replace it for all the tests with no issues as test framework based function.
But I can't use it within the shipyard repo for `fetchClusterIDs` function, because the test framework is yet initialized at that place.
As I mentioned above, in current implementation it uses the standalone function, which the test framework function relies on.


I can convert this function into a standalone, but then it will require me to call this function from the tests in the following way:
Instead of `f.FindGatewayNodes(i)` use `framework.FindGatewayNodes(i)`.
I guess it should work.
But from my point of view it will create some sort of mess in the functions call flow.

Will appreciate you thoughts.
Thanks.

skitt: `f.FindGatewayNodes(i)` and `framework.FindGatewayNodes(i)` are both just function calls. The difference is that `f.FindGatewayNodes(i)` calls the function associated with the type of `f` (and `f` can even be `nil`, as long as it’s typed), and the function can use `f` (but doesn’t, in this case); whereas `framework.FindGatewayNodes(i)` calls the function in the package aliased as `framework`. There’s no difference in terms of “function call flow”.

MaxBab: Thank you.
I got the point.

Will check and verify that I'm able to call the function in the `framework.func_name` way.

MaxBab: Set it as a standalone functions.
Working.
Thank you for the help.

test/e2e/framework/nodes.go
MaxBab: Suggested changes implemented.
Helper function added.

test/e2e/framework/nodes.go
MaxBab: Implemented.

test/e2e/framework/util.go
```diff
@@ -22,6 +22,9 @@ import (
 	"strings"
 
 	. "github.com/onsi/gomega"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/selection"
+
```
MaxBab: @tpantelis @skitt @sridhargaddam 

Hi Guys,
Would appreciate your advice here.

I have a linting error for these imports.

I get the following error from "gci":
`test/e2e/framework/util.go:27: File is not `gci`-ed with --skip-generated -s standard,default (gci)`

It actually complains on the empty line 27.
But if I'm removing the empty line, it throws a different error:
```
test/e2e/framework/util.go:26: File is not `goimports`-ed (goimports)
	"k8s.io/apimachinery/pkg/selection"
```
And if I'm running the "goimports", it adds the empty line back.

Could you advice, please?

tpantelis: I've run into this before - the linters clash. I fixed it by putting the comment on the same line as the import.

tpantelis (approved):

test/e2e/framework/util.go
MaxBab: Thanks!!

dfarrell07 (approved):

Issue #858, opened 2022-07-06
Stop using Weave


skitt:
Weave is largely unsupported upstream, and it’s not clear we actually _need_ it specifically; we initially added Weave so that we could work on network policy support.

We should look at the current CNI space, in particular kind-friendly CNIs, and switch to whatever seems most appropriate.

mkolesnik:
How about just using the default kindnet for now?

IDK if we have any deficiencies with it, but if Weave is unsupported then using the default would be our best bet until we hit something it doesn't provide.

sridhargaddam:
When deployed with default (aka kindnet), subctl CNI detection reports the CNI as "generic" but all the test-cases are passing fine. We can enhance the CNI detection code to detect and report the CNI in a future PR.

PR #956, opened 2022-09-20
Use the default kind CNI, drop support for Weave

skitt:
Weave is no longer actively maintained, and our tests work with the default kind CNI. Until we need something more featureful, switch to the default kind CNI, and remove support for Weave altogether.

See https://github.com/submariner-io/shipyard/issues/858.

Signed-off-by: Stephen Kitt <skitt@redhat.com>

Depends on https://github.com/submariner-io/lighthouse/pull/880
Depends on https://github.com/submariner-io/releases/pull/495
Depends on https://github.com/submariner-io/subctl/pull/270
Depends on https://github.com/submariner-io/submariner/pull/2029
Depends on https://github.com/submariner-io/submariner-charts/pull/268
Depends on https://github.com/submariner-io/submariner-operator/pull/2246
Fixes #858

PR #967, opened 2022-10-04
Fix flaky e2e failures with kindnet cni

### LINKED ISSUE #966: Flaky e2e failures with kindnet CNI
While transitioning to kindnet CNI in our KIND deployments, it was seen that HostNetworking e2e tests were occasionally failing. On debugging this issue, it was found that `kindnet` CNI does not create a dedicated CNI Interface on the nodes. Submariner relies on the CNI Interface to support some of the use-cases like Health-check between the GW nodes and HostNetwork to remote cluster support.

However, it was seen that as soon as we schedule a pod on the node, `kindnet` CNI creates a `veth-xxx` interface on the Host which can be used as a CNI Interface. So, in order to continue validating the above Submariner use-cases, we can schedule a dummypod as a DaemonSet in the clusters as a workaround (and this can be removed when we switch to a different CNI).

sridhargaddam:
While transitioning to kindnet CNI in our KIND deployments, it was seen that HostNetworking e2e tests were occasionally failing. On debugging this issue, it was found that kindnet CNI does not create a dedicated CNI Interface on the nodes. Submariner relies on the CNI Interface to support some of the use-cases like Health-check between the GW nodes and HostNetwork to remote cluster support. However, it was seen that as soon as we schedule a pod on the node, kindnet CNI creates a veth-xxx interface on the Host which can be used as a CNI Interface. So, in order to continue validating the above Submariner use-cases, this PR schedules a dummypod as a DaemonSet in the clusters as a workaround (and this can be removed when we switch to a different CNI).

Fixes: https://github.com/submariner-io/shipyard/issues/966
Related to: https://github.com/submariner-io/shipyard/issues/858

### DIFF
--- a/scripts/shared/deploy.sh
+++ b/scripts/shared/deploy.sh
@@ -31,6 +31,8 @@ readonly MARKETPLACE_NAMESPACE="olm"
 IPSEC_PSK="$(dd if=/dev/urandom count=64 bs=8 | LC_CTYPE=C tr -dc 'a-zA-Z0-9' | fold -w 64 | head -n 1)"
 # shellcheck disable=SC2034
 readonly IPSEC_PSK
+# we use this namespace to deploy a dummypod as a daemonSet as a workaround for kindnet clusters.
+readonly KINDNET_WORKAROUND_NS="subm-kindnet-workaround"
 
 ### Common functions ###
 
@@ -65,6 +67,11 @@ function create_catalog_source() {
   fi
 
   echo "[INFO](${cluster}) Catalog source ${cs} created"
+}
+
+function schedule_dummy_pod_on_all_clusters() {
+    run_subm_clusters create_namespace ${KINDNET_WORKAROUND_NS}
+    run_subm_clusters deploy_daemonset ${KINDNET_WORKAROUND_NS} "${RESOURCES_DIR}"/dummypod.yaml
 }
 
 # Create an OperatorGroup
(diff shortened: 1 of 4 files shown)


dfarrell07:
```
In scripts/shared/deploy.sh line 74:
    run_subm_clusters deploy_daemonset ${KINDNET_WORKAROUND_NS} ${RESOURCES_DIR}/dummypod.yaml
                                                                ^--------------^ SC2086 (info): Double quote to prevent globbing and word splitting.
```
[Shell](https://github.com/submariner-io/shipyard/actions/runs/3182668168/jobs/5188913551#logs)

aswinsuryan (approved):

tpantelis (approved):

scripts/shared/deploy.sh
```diff
@@ -67,6 +69,11 @@ function create_catalog_source() {
   echo "[INFO](${cluster}) Catalog source ${cs} created"
 }
 
+function schedule_dummy_pod_on_all_clusters() {
+    run_subm_clusters create_namespace ${KINDNET_WORKAROUND_NS}
+    run_subm_clusters deploy_daemonset ${KINDNET_WORKAROUND_NS} "${RESOURCES_DIR}"/dummypod.yaml
```
tpantelis, line 74: Should we delete the daemonset/namespace?

sridhargaddam: The daemonSet pod runs the command "sleep infinity". They will be deleted when the KIND clusters are deleted.

dfarrell07:
Do we expect all the deploy jobs to still fail?

sridhargaddam:
> Do we expect all the deploy jobs to still fail?

The deploy jobs are failing because shipyard is still running with weave CNI and its fully broken now :-( 
We need the following fix - https://github.com/submariner-io/shipyard/pull/956/files
But it needs rebase and since it does not have the fix from this PR it would fail too. So, let me include a small change to switch from weavenet to kindnet in this PR itself, so that we can validate the full functionality as part of this PR.

tpantelis:
> The deploy jobs are failing because shipyard is still running with weave CNI and its fully broken now :-( We need the following fix - https://github.com/submariner-io/shipyard/pull/956/files But it needs rebase and since it does not have the fix from this PR it would fail too. So, let me include a small change to switch from weavenet to kindnet in this PR itself, so that we can validate the full functionality as part of this PR.

I think we should just combine the 2 PRs, ie include the commit from #956 here.

dfarrell07 (approved):

PR #977, opened 2022-10-14
Deploy dummypod ds even on control-nodes

### LINKED ISSUE #966: Flaky e2e failures with kindnet CNI

sridhargaddam:
In order to support HostNetwork use-cases with the kindnet CNI, as part of one of the previous PRs[1], we were deploying a dummyPod as a daemonSet. The DS pods are scheduled on all the worker nodes of the cluster except for the control-node. Normally, the control-nodes host some pods like coredns/provisioner etc on the nodes. In such scenarios, scheduling dummyPod on the control-nodes was not necessary. However, in some KIND deployments, it was seen that all the coredns, provisioner pods were scheduled on the worker nodes and not on the control node. Because of this, the route-agent was unable to discover the CNI interface on the control-node and it was causing failures in e2e hostnetwork use-case.

This PR modifies the dummyPod daemonSet yaml to include the necessary tolerations so that they will be scheduled even on the control-nodes as well, which should fix the e2e errors.

[1] https://github.com/submariner-io/shipyard/pull/967/files

Fixes: https://github.com/submariner-io/shipyard/issues/966

### DIFF
--- a/scripts/shared/resources/dummypod.yaml
+++ b/scripts/shared/resources/dummypod.yaml
@@ -12,6 +12,13 @@ spec:
       labels:
         app: dummypod
     spec:
+      tolerations:
+        - key: node-role.kubernetes.io/control-plane
+          operator: Exists
+          effect: NoSchedule
+        - key: node-role.kubernetes.io/master
+          operator: Exists
+          effect: NoSchedule
       containers:
         - name: dummypod
           image: "quay.io/submariner/nettest:devel"


sridhargaddam:
Looking into the linter issue

tpantelis (approved):

vthapar (approved):

dfarrell07 (approved):

PR #981, opened 2022-10-17
Stop vendoring

### LINKED ISSUE #804: Go compilation shouldn’t rely on vendoring
Since we don’t ship the contents of `vendor`, and since Go 1.14+ with modules can build without local vendoring, we should remove all the vendoring parts of our build infrastructure.

See https://go.dev/ref/mod#vendoring for details.

skitt:
go mod vendor is no longer necessary; see
https://go.dev/ref/mod#vendoring for details. Dependencies will be fulfilled using the module cache.

Fixes: #804


Makefile.inc
```diff
@@ -274,7 +274,7 @@ post-mortem:
 	$(SCRIPTS_DIR)/post_mortem.sh
 
 # [unit] executes the Go unit tests of the project
-unit: $(VENDOR_MODULES)
+unit:
```
mkolesnik, line 282: You should also remove it from `e2e` in `Makefile`

mkolesnik (changes requested):

mkolesnik (approved):

dfarrell07 (approved):

tpantelis (approved):

PR #995, opened 2022-10-20
Remove all security permissions for netshoot pod

dfarrell07:
This container doesn't need need any security permissions, remove all.

This will hopefully resolve issues with KEP-2579: Pod Security Admission Control, particularly new features in OpenShift 4.12.


dfarrell07:
> Error from server (BadRequest): error when creating "STDIN": Deployment in version "v1" cannot be handled as a Deployment: strict decoding error: unknown field "spec.securityContext"

I'm following an example from the GDoc, but clearly doing something wrong...

dfarrell07:
It would be nice to get this meged before rc0 so it will be tested during the test days.

skitt (approved):

tpantelis (approved):

PR #998, opened 2022-10-20
Search for NonGatewayNodes with GatewayLabel absent

MaxBab:
When searching for NonGatewayNodes, the function should search for the nodes with missing "submariner.io/gateway" label instead of the label with "false" value.

Filter out the nodes with "node-role.kubernetes.io/master" label as it doesn't allow to schedule the pods on it.

### DIFF
--- a/test/e2e/framework/nodes.go
+++ b/test/e2e/framework/nodes.go
@@ -50,7 +50,8 @@ func FindGatewayNodes(cluster ClusterIndex) []v1.Node {
 func FindNonGatewayNodes(cluster ClusterIndex) []v1.Node {
 	nodes, err := KubeClients[cluster].CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{
 		LabelSelector: labels.NewSelector().Add(
-			NewRequirement(GatewayLabel, selection.Exists, []string{}),
+			// Ignore the control plane node labeled as master as it doesn't allow scheduling of pods
+			NewRequirement("node-role.kubernetes.io/master", selection.DoesNotExist, []string{}),
 			NewRequirement(GatewayLabel, selection.NotEquals, []string{"true"})).String(),
 	})
 	Expect(err).NotTo(HaveOccurred())


MaxBab:
@tpantelis 

Please, take a look.
I needed to modify the `FindNonGatewayNodes` function to meet the requirements.

test/e2e/framework/nodes.go
```diff
@@ -50,8 +50,9 @@ func FindGatewayNodes(cluster ClusterIndex) []v1.Node {
 func FindNonGatewayNodes(cluster ClusterIndex) []v1.Node {
 	nodes, err := KubeClients[cluster].CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{
 		LabelSelector: labels.NewSelector().Add(
-			NewRequirement(GatewayLabel, selection.Exists, []string{}),
-			NewRequirement(GatewayLabel, selection.NotEquals, []string{"true"})).String(),
+			// Ignore the control plane node labeled as master as it doesn't allow scheduling of pods
+			NewRequirement("node-role.kubernetes.io/master", selection.DoesNotExist, []string{}),
+			NewRequirement(GatewayLabel, selection.DoesNotExist, []string{})).String(),
```
tpantelis: Since our semantics elsewhere are such that a node is a GW if the gateway label exists **and** is equal to "true", I think we should be consistent here as well. The following handles non-existent label as well (in my testing):

```suggestion
			NewRequirement(GatewayLabel, selection.NotEquals, []string{"true"})).String(),
```

tpantelis (changes requested):

test/e2e/framework/nodes.go
MaxBab: Done

tpantelis (approved):

MaxBab:
Hi @dfarrell07 ,

Could you take a look at the patch, please?

dfarrell07 (approved):

PR #1014, opened 2022-11-07
Add OCP on AWS with SubM deployment test

dfarrell07:
Include a `make deploy using=aws-ocp` job in the scripts tests.

### DIFF
--- a/.github/workflows/scripts.yml
+++ b/.github/workflows/scripts.yml
@@ -57,6 +57,7 @@ jobs:
             lighthouse: lighthouse
           - extra-toggles: air-gap
             ovn: ovn
+          - extra-toggles: aws-ocp
           - extra-toggles: ovn
           - deploytool: operator
             extra-toggles: lighthouse


dfarrell07:
```
[10:53:58.855] [dir=shipyard; fn=ensure_openshift_install]$ curl -Ls https://mirror.openshift.com/pub/openshift-v4/clients/ocp//openshift-install-linux.tar.gz
[10:53:58.858] [dir=shipyard; fn=ensure_openshift_install]$ tar -xzf - -C /root/.local/bin openshift-install
gzip: stdin: not in gzip format
```

CC @mkolesnik

Is this failing in CI because of missing secrets? Would it be possible to add CI if we store them in GH secrets for the repo/org?

mkolesnik:
> ```
> [10:53:58.855] [dir=shipyard; fn=ensure_openshift_install]$ curl -Ls https://mirror.openshift.com/pub/openshift-v4/clients/ocp//openshift-install-linux.tar.gz
> [10:53:58.858] [dir=shipyard; fn=ensure_openshift_install]$ tar -xzf - -C /root/.local/bin openshift-install
> gzip: stdin: not in gzip format
> ```
> 
> CC @mkolesnik
> 
> Is this failing in CI because of missing secrets? Would it be possible to add CI if we store them in GH secrets for the repo/org?

No it's actually due to #1013 but keep in mind OCP 4.12 isn't very stable right now, so we might want to test this with K8s version = 1.24 for the time being.

Regardless, we would need to store and pass credentials somehow, currently we auto-mount the aws credentials file (`~/.aws/credentials`). Maybe we can store it as a GH secret and dump it into a credentials file on the job itself?

@skitt any ideas re this?

Also it won't work currently as the OCP installer is interactive, unless you have some way of driving it.
What we could do is prepare a template beforehand and copy it for this job (so that it doesn't interfere with manual invocations) and that should negate the need to interactively fill values.

One more thing to keep in mind is that we'd need to cleanup in the end (which we currently don't do), and we'll have to always attempt it even if the step failed, so `gh-actions/e2e/action.yaml` would need to be updated to handle cleanup (could be a toggle so that we don;t waste time on cleanup by default) *and* the cleanup needs to be done `using=aws-ocp` as well.

skitt:
https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/about-security-hardening-with-openid-connect should allow us to give a GHA access to AWS without storing credentials even in GH secrets.

Pre-populated templates work very well for non-interactive `openshift install`, that’s what I use all the time.

dfarrell07:
Another angle to this discussion is the cost. We likely should run this periodically/infrequently, like weekly. And also have the option to trigger via labels if needed. It would also be cool if we could quantify the exact cost of a run, so we know and can budget on a monthly/yearly basis correctly.

dfarrell07:
Opened a dedicated issue to track.

PR #1041, opened 2022-11-17
Generate 12 characters short hashes

mkolesnik:
The hashes are used for versioning, amongst other things, the images. Sometimes the hash can be longer than the default 7, but when releasing the calculated hash is 7 characters long.

To be consistent, without risk of collisions, always generate a 12 character long hash.


skitt (approved):

maayanf24 (approved):

mkolesnik:
Backported as #1042

PR #1062, opened 2022-12-06
Do not skip "master" nodes for non-gw nodes search

MaxBab:
The "FindNonGatewayNodes" function search for the nodes of the cluster that do not hold the gateway label.
- Filter the nodes by missing the gateway label instead of comparing it to "false"

- A change between k8s version 1.23 and 1.25 is the removal of "master" label from the control-plane nodes. In order to not rely on this label, which existance changes between the versions, do not filter the nodes by it.

### DIFF
--- a/test/e2e/framework/nodes.go
+++ b/test/e2e/framework/nodes.go
@@ -50,8 +50,6 @@ func FindGatewayNodes(cluster ClusterIndex) []v1.Node {
 func FindNonGatewayNodes(cluster ClusterIndex) []v1.Node {
 	nodes, err := KubeClients[cluster].CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{
 		LabelSelector: labels.NewSelector().Add(
-			// Ignore the control plane node labeled as master as it doesn't allow scheduling of pods
-			NewRequirement("node-role.kubernetes.io/master", selection.DoesNotExist, []string{}),
 			NewRequirement(GatewayLabel, selection.NotEquals, []string{"true"})).String(),
 	})
 	Expect(err).NotTo(HaveOccurred())


test/e2e/framework/nodes.go
```diff
@@ -50,9 +50,7 @@ func FindGatewayNodes(cluster ClusterIndex) []v1.Node {
 func FindNonGatewayNodes(cluster ClusterIndex) []v1.Node {
 	nodes, err := KubeClients[cluster].CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{
 		LabelSelector: labels.NewSelector().Add(
-			// Ignore the control plane node labeled as master as it doesn't allow scheduling of pods
-			NewRequirement("node-role.kubernetes.io/master", selection.DoesNotExist, []string{}),
-			NewRequirement(GatewayLabel, selection.NotEquals, []string{"true"})).String(),
+			NewRequirement(GatewayLabel, selection.DoesNotExist, []string{})).String(),
```
tpantelis: I think we should keep the GW label check as is (as long as it works). The GW criteria is the label set to “true” so conversely the label missing or set to “false” is more complete/correct.

tpantelis (changes requested):

test/e2e/framework/nodes.go
MaxBab: Done.
I have verified the fix on "kind" environment with k8s version - 1.23 and 1.25.
Also, I have verified it on ocp based environment.

tpantelis: We want just `NotEquals` and not also `DoesNotExist`.

tpantelis (changes requested):

dfarrell07 (approved):

tpantelis (approved):

PR #1082, opened 2022-12-20
Add `TEST_ARGS` variable and fail fast E2E on CI

### LINKED ISSUE #45: E2E checks should abort on the first failure
Currently, E2E tests run the full test suite, even when tests are failing. The failures often show up as time-outs, and a failing test suite thus takes a very long time; but in most cases, we’re only interested in whether the tests pass or fail, not the details of the failure (after the first one). This is particularly visible with flaky tests and causes significant delays in the release process.

We should add an option to abort on the first failure, and enable it when run from GHAs.

This would involve:

- [ ] Providing some way of controlling whether `make e2e` runs all tests or aborts on the first failure (at the Ginkgo level); this should default to run all tests, to preserve existing behaviour for local invocations.
- [ ] Setting up all the E2E GHAs to run `make e2e` with the setting to abort on first failure; this might involve changes in all projects.

This shouldn’t affect the GHA test matrices at all — we still want to run E2E in all desired configurations, the only change is that within each run, the first failure should abort _that_ run.

mkolesnik:
Add the variable to allow passing any arbitrary arguments to the E2E tests, and set to fail fast by default on the CI.

This will save time on CI in case of test failures, and could be disabled when necessary.
On local invocations, the default behavior will remain running all tests, but the flag can be passed explicitly if desired.

Resolves https://github.com/submariner-io/enhancements/issues/45


mkolesnik:
I guess it just proved itself :)

https://github.com/submariner-io/shipyard/actions/runs/3739671285/jobs/6347156836

skitt (approved):

sridhargaddam (approved):

mkolesnik:
This was backported to 0.,14 as #1085

Backporting to 0.13 has too many conflicts and will not be performed

PR #1086, opened 2022-12-22
Switch to ginkgo v2

skitt:
The GINKGO_PANIC replacement values come from
https://github.com/kubernetes/kubernetes/commit/05c0f4a298a82a167b85323ac189bf6a528f288a

See https://onsi.github.io/ginkgo/MIGRATING_TO_V2#migration-strategy-5
for the ginkgo configuration migration strategy.

The shell scripts handle both v1 and v2 to avoid breaking dependent
projects.

### DIFF
--- a/scripts/shared/e2e.sh
+++ b/scripts/shared/e2e.sh
@@ -27,12 +27,21 @@ function generate_kubecontexts() {
 function test_with_e2e_tests {
     cd "${DAPPER_SOURCE}/${TESTDIR}"
 
-    # shellcheck disable=SC2086 # TEST_ARGS is split on purpose
-    ${GO:-go} test -v -timeout 30m -args -test.timeout 15m \
-        -submariner-namespace $SUBM_NS "${clusters[@]/#/-dp-context=}" \
-        -ginkgo.v -ginkgo.randomizeAllSpecs -ginkgo.trace \
-        -ginkgo.reportPassed -ginkgo.reportFile "${DAPPER_OUTPUT}/e2e-junit.xml" \
-        $TEST_ARGS 2>&1 | tee "${DAPPER_OUTPUT}/e2e-tests.log"
+    if go mod graph | grep 'submariner.*ginkgo/v2' >/dev/null; then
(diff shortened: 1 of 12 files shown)


scripts/shared/e2e.sh
```diff
@@ -36,8 +36,8 @@ function test_with_e2e_tests {
 
     ${GO:-go} test -v -timeout 30m -args -test.timeout 15m \
         -submariner-namespace $SUBM_NS "${clusters[@]/#/-dp-context=}" ${gn:+"$gn"} \
-        -ginkgo.v -ginkgo.randomizeAllSpecs -ginkgo.trace \
-        -ginkgo.reportPassed -ginkgo.reportFile "${DAPPER_OUTPUT}/e2e-junit.xml" \
+        --ginkgo.v --ginkgo.randomize-all --ginkgo.trace \
```
sridhargaddam: Do we need any changes in the consuming projects like submariner as well?

skitt: Yes, and I initially thought everything would be OK because we’d only get this after bumping the Shipyard dependency, but since `e2e.sh` changes all the other Submariner projects will break.

skitt: I’ve updated the shell scripts to handle both versions of ginkgo and avoid breaking dependent projects.

scripts/shared/e2e.sh
```diff
@@ -7,8 +7,13 @@ source "${SCRIPTS_DIR}/lib/utils"
 
 # shellcheck disable=SC2206 # Split on purpose
 ginkgo_args=(${TEST_ARGS})
-[[ -n "${FOCUS}" ]] && ginkgo_args+=("-ginkgo.focus=${FOCUS}")
-[[ -n "${SKIP}" ]] && ginkgo_args+=("-ginkgo.skip=${SKIP}")
+if grep -q ginkgo/v2 go.mod; then
```
skitt: This needs to support looking for `go.mod` in parent directories.

tpantelis (approved):

dfarrell07 (approved):

PR #1097, opened 2023-01-05
Group container run output for GHA

mkolesnik:
Usually we don't really care to see the whole info of the Shipyard container invocation on the log in GHA.

This will group it in a collapsible group for easier log reading on the CI.


scripts/shared/entry
```diff
@@ -15,4 +15,5 @@ else
 fi
 
 shift
+[[ -z "$GITHUB_ACTOR" ]] || echo "::endgroup::"
```
skitt:
```suggestion
# This ends the group started in Makefile.dapper
[[ -z "$GITHUB_ACTOR" ]] || echo "::endgroup::"
```

skitt (changes requested):
Let’s add a comment to point to where the group starts.

skitt (approved):

mkolesnik:
I changed the wording to make more sense when looking in the GHA log 
image

mkolesnik:
Seems that relying on various `GITHUB_` variables to be set is problematic, added the `CI` to be passed down so that we can definitively identify a CI run.

dfarrell07 (approved):

PR #1107, opened 2023-01-15
Support local Go modules

### LINKED ISSUE #644: Dapper-based builds should support local replaces
It’s often convenient, when working on changes across repositories, to use local `replaces` in `go.mod`. Unfortunately this fails when building with Dapper since we only copy the current project into the build container. Such builds should support local replaces by also copying other projects as needed.

There is a workaround: run `go mod vendor` before the build.

mkolesnik:
Support using local Go modules inside the Shipyard container by specifying one or more (colon separated) paths with `LOCAL_MODULES`.

The corresponding Submariner modules will be used from the local file system instead of being pulled from GitHub.

Resolves #644 


skitt:
Given that this assumes that the local Go modules are already available alongside the current module, are there any scenarios where this would be used _without_ existing `replace` statements? If not, we could use the latter directly, instead of a new `LOCAL_MODULES` variable.

PR #1112, opened 2023-01-18
Support local Go modules

### LINKED ISSUE #644: Dapper-based builds should support local replaces

mkolesnik:
Support using local Go modules inside the Shipyard container by specifying one or more (colon separated) paths with `LOCAL_MODULES`.

The corresponding Submariner modules will be used from the local file system instead of being pulled from GitHub.

Resolves #644 


skitt:
Given that this assumes that the local Go modules are already available alongside the current module, are there any scenarios where this would be used _without_ existing `replace` statements? If not, we could use the latter directly, instead of a new `LOCAL_MODULES` variable.

    go list -m -f "{{.Path}} {{.Replace}}" all | grep -v "<nil>"

will list the replacements in the current module. See https://go.dev/ref/mod#go-list-m for the list of available fields.

A simpler change which I think would improve this would be to use `..` instead of `replaces` — that avoids a few other related changes (such as the extra `-prune`s on `find`).

mkolesnik:
> Given that this assumes that the local Go modules are already available alongside the current module, are there any scenarios where this would be used _without_ existing `replace` statements? If not, we could use the latter directly, instead of a new `LOCAL_MODULES` variable.

This doesn't assume anything about the locations of the replaces on the host machines, you can use whatever location you'd like.
In fact this doesn't care about any `replace` statements you might have already.

> 
> ```
> go list -m -f "{{.Path}} {{.Replace}}" all | grep -v "<nil>"
> ```
> 
> will list the replacements in the current module. See https://go.dev/ref/mod#go-list-m for the list of available fields.

This captures any replacement, which isn't very relevant to us.

e.g.
```bash
[root@7eae250e587f submariner-operator]# go list -m -f "{{.Path}} {{.Replace}}" all | grep -v "<nil>"
go: can't compute 'all' using the vendor directory
	(Use -mod=mod or -mod=readonly to bypass.)
[root@7eae250e587f submariner-operator]# go list -mod=mod -m -f "{{.Path}} {{.Replace}}" all | grep -v "<nil>"
sigs.k8s.io/controller-runtime sigs.k8s.io/controller-runtime v0.13.1-0.20221213002533-ca4b4def4fb3
```

> 
> A simpler change which I think would improve this would be to use `..` instead of `replaces` — that avoids a few other related changes (such as the extra `-prune`s on `find`).

Yes unfortunately I tried that first [1] and it doesn't work for the image builds in consuming projects, hence I put an explicit comment stating why it's necessary to mount inside the `replaces` directory:
```
# We mount inside a `replaces` sub-directory to make it easily accessible to any further container invocations (i.e. when building images).
```

Since `docker build` doesn't support mounting volumes during the build, and you can't copy anything outside the working directory, this is the only way I could get it to work.

[1] e.g. https://github.com/mkolesnik/shipyard/actions/runs/3936783583/jobs/6733621424#step:7:733

skitt:
> > Given that this assumes that the local Go modules are already available alongside the current module, are there any scenarios where this would be used _without_ existing `replace` statements? If not, we could use the latter directly, instead of a new `LOCAL_MODULES` variable.
> 
> This doesn't assume anything about the locations of the replaces on the host machines, you can use whatever location you'd like. In fact this doesn't care about any `replace` statements you might have already.

OK, let me rephrase: given that this assumes that the local Go modules are already available locally, not handled as a generic Go dependency (downloaded as necessary, copied from the local Go module cache, etc.), and it seems likely that developers working on changes that involve changes across multiple local Go modules will use `replace` statements in the relevant `go.mod` files, wouldn’t it be better to use the latter instead of a new `LOCAL_MODULES` variable?

Basically what I would like to have is the standard Go development model using `replace` be transparently supported by `make foo` even when that ends up using Dapper.

> > ```
> > go list -m -f "{{.Path}} {{.Replace}}" all | grep -v "<nil>"
> > ```
> >   
> > will list the replacements in the current module. See https://go.dev/ref/mod#go-list-m for the list of available fields.
> 
> This captures any replacement, which isn't very relevant to us.

Yes, it wasn’t supposed to be a fully-worked-out solution, just show how to retrieve `replace` information without manually parsing `go.mod`.

skitt:
> > A simpler change which I think would improve this would be to use `..` instead of `replaces` — that avoids a few other related changes (such as the extra `-prune`s on `find`).
> 
> Yes unfortunately I tried that first [1] and it doesn't work for the image builds in consuming projects, hence I put an explicit comment stating why it's necessary to mount inside the `replaces` directory:
> 
> ```
> # We mount inside a `replaces` sub-directory to make it easily accessible to any further container invocations (i.e. when building images).
> ```

I interpreted the “easily” as meaning that there was a harder approach that didn’t require this.

> Since `docker build` doesn't support mounting volumes during the build, and you can't copy anything outside the working directory, this is the only way I could get it to work.

Indeed, and in the early Axon patches to address this, the container build directory was changed to the parent to avoid this limitation.

mkolesnik:
> I interpreted the “easily” as meaning that there was a harder approach that didn’t require this.

Yes we could have a much harder approach but it's better to avoid.

> Indeed, and in the early Axon patches to address this, the container build directory was changed to the parent to avoid this limitation.

I wouldn't feel comfortable with doing this on a wide scale, the proposed solution is much cleaner (e.g. the replaces get mounted ro, and you only mount exactly what you want).

mkolesnik:
> OK, let me rephrase: given that this assumes that the local Go modules are already available locally, not handled as a generic Go dependency (downloaded as necessary, copied from the local Go module cache, etc.), and it seems likely that developers working on changes that involve changes across multiple local Go modules will use `replace` statements in the relevant `go.mod` files, wouldn’t it be better to use the latter instead of a new `LOCAL_MODULES` variable?

OK I see your point now, but it still conflicts with the limitations for building images.
As explained, mounting directories during the build is not possible, and copying them in is only possible from inside the working directory and not accessing any parents.

> Basically what I would like to have is the standard Go development model using `replace` be transparently supported by `make foo` even when that ends up using Dapper.

I can think of **a** solution, not sure it's optimal as it involves a bit of copying, but should work overall:

Outside `.dapper` invocation:
1. Gather replaces from inside the go mod (as you requested).
2. For any local replaces found in the `go.mod`:
  2.1. git clone them into the corresponding `vendor` directory.

Inside `.dapper` invocation:
1. Gather replaces from inside the go mod.
2. For any local replaces found in the `go.mod`:
  2.1. git clone them from the corresponding `vendor` directory into the requested directory.

WDYT?

skitt:
It seems to me there are two different scenarios here, which explains the misunderstandings above.

The first is CI. This is immediately relevant in the “consuming” jobs run from Shipyard, and the goal is to run consuming builds not only with the new Shipyard Dapper base image, but also using the new Shipyard Go module (with the `replace` you tested in https://github.com/mkolesnik/shipyard/pull/1). Further down the line, local Go module support in CI would allow us to test PRs with dependencies on other PRs with the contents of those PRs, which would be very useful.

The second is local development. For example, if I’m working on the operator but need to make related changes to Admiral, I would like to be able to work in Admiral and have a local `replace` in the operator, pulling in my modified Admiral. I want to have `make unit` or `make e2e` test the same code as `make unit LOCAL_BUILD=1`, and likewise for all the other targets. In particular, this means being able to test uncommitted changes in both repositories.

In the latter case, using the existing `replace`, typically `../admiral` for Admiral, just works with your approach, with the appropriate path instead of `replaces/foo`. The current PR, as is, breaks the local development workflow: `replace` statements in `go.mod` are overwritten by those created by `scripts/shared/entry`.

For CI, the use-case is simpler, and externally-visible changes don’t matter, so the “consuming” job itself could set things up using this approach. That is to say, `entry` would follow whatever the _existing_ `replace` statements say, without using a new `replaces` directory; the “consuming” job itself, when it sets up local modules for its own purposes, would use a `replaces` directory inside the consuming project, and set up `replace` statements in `go.mod` accordingly.

mkolesnik:
Thanks @skitt I eventually understood what you meant :grin: 

I was indeed thinking mostly of CI but I understood what you wanted and updated the PR.

Please re-review

.dapper
```diff
@@ -116,6 +122,11 @@ pass_credentials "${HOME}/.aws/credentials"
 # Pass through ~/.gcp/osServiceAccount.json so that the container can operate on GCP (if needed).
 pass_credentials "${HOME}/.gcp/osServiceAccount.json"
 
+# Attempt to mount any local replaces specified in `go.mod`.
+while read -a replace; do
+    mount_volume "${replace[3]}" "${DAPPER_SOURCE}/${replace[3]}"
+done < <(sed -n -e '/\breplace\b/p' go.mod)
```
skitt: This will only work for individual `replace` lines, and will catch non-local replacements.

```suggestion
while read replace; do
    mount_volume "${replace}" "${DAPPER_SOURCE}/${replace}"
done < <(go list -mod=mod -m -f "{{.Replace}}" all | grep -E '^\.\.?/')
```

skitt (changes requested):

.dapper
mkolesnik: The problem is, we cant guarantee that go is available (and the correct version) outside the container (this is outside the dapper).

As for blocks, we can look for `=>` instead to cover both cases (obviously there will be some handling for different directives).

As for non-local replacements its not a problem since mount_volume ignores ones that dont exist, but for clarity we can test it here as well.

skitt: Ah yes, indeed, so

```suggestion
# We can't rely on having the right version (or any version) of go, so parse go.mod manually.
while read -a replace; do
    mount_volume "${replace[1]}" "${DAPPER_SOURCE}/${replace[1]}"
done < <(grep -o -E '=>[ ]*\.\.?/[^ ]+' go.mod)
```

skitt (approved):

tpantelis (approved):

PR #1114, opened 2023-01-23
Use local `Shipyard` when testing consuming E2E

### LINKED ISSUE #756: Consuming project check should check updated library as well
Currently the "E2E consuming projects" check rebuilds the image, but the Shipyard go library isn't shipped in the image but rather through go.mod.
We should update the check so that it uses the library with the changes from the PR.

One way this is achievable is by copying the go files into the base image and using local replace in `go.mod` to point to the code inside the image. Or we could allow mounting external volumes through our "dapper" and use local replace that way.

mkolesnik:
Consuming projects running E2E tests rely on code supplied by Shipyard. Use the local Shipyard when testing them, to make sure the latest code is tested, and catch any breakages.

Resolves #756 

Depends on #1112 

### DIFF
--- a/.github/workflows/consuming.yml
+++ b/.github/workflows/consuming.yml
@@ -41,36 +41,41 @@ jobs:
           - project: submariner-operator
             k8s_version: '1.24'
     steps:
-      - name: Check out the Shipyard repository
-        uses: actions/checkout@3df4ab11eba7bda6032a0b82a6bb43b11571feac
-
-      - name: Build the latest Shipyard image
-        run: make images
-
       - name: Check out the ${{ matrix.project }} repository
         uses: actions/checkout@3df4ab11eba7bda6032a0b82a6bb43b11571feac
         with:
           # This is replaced to stable branch by auto release process
           ref: devel
           repository: submariner-io/${{ matrix.project }}
-          path: ${{ matrix.project }}
+
+      # Check out Shipyard as a sub directory of the project, so that `go replace` can work.
+      # As it's all run inside an ephemeral container, this way guarantees the Shipyard bits will be readily available for `go replace`.
+      - name: Check out the Shipyard repository
+        uses: actions/checkout@3df4ab11eba7bda6032a0b82a6bb43b11571feac
+        with:
+          path: shipyard
+
+      - name: Build the latest Shipyard images
+        run: make -C shipyard images
 
       - name: Copy Shipyard resources
-        run: cp -n Dockerfile.* Makefile.dapper .dapper ${{ matrix.project }}/
+        run: cp -n shipyard/{Dockerfile.*,Makefile.dapper,.dapper} .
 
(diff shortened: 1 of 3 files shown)


.github/workflows/consuming.yml
```diff
@@ -41,36 +41,39 @@ jobs:
...
       - name: Copy Shipyard resources
-        run: cp -n Dockerfile.* Makefile.dapper .dapper ${{ matrix.project }}/
+        run: cp -n shipyard/{Dockerfile.*,Makefile.dapper,.dapper} .
 
       - name: Make sure ${{ matrix.project }} is using the built Shipyard image
-        run: sed -i 's/shipyard-dapper-base:*.*/shipyard-dapper-base:dev/' ${{ matrix.project }}/Dockerfile.dapper
+        run: sed -i 's/shipyard-dapper-base:*.*/shipyard-dapper-base:dev/' Dockerfile.dapper
+
+      - name: Make sure ${{ matrix.project }} is using the Shipyard go module
+        run: make go-replace REPLACE=github.com/submariner-io/shipyard=./shipyard
```
skitt, line 68: This is the significant change — most of the rest of this PR is the result of inverting the extraction. Is there any particular reason to do so? The commit message doesn’t explain that part of the change.

mkolesnik: Not sure what you mean?

What's missing from the commit message? It already specifies a reason:
```
Consuming projects running E2E tests rely on code supplied by Shipyard.
Use the local Shipyard when testing them, to make sure the latest code
is tested, and catch any breakages.
```

skitt: The commit message explains these two lines. The PR also inverts the extraction: previously we checked out Shipyard, then checked out the project undergoing testing into a subdirectory; but the PR changes that so that the project being tested is checked out, and then Shipyard is checked out into a subdirectory of that. The commit message doesn’t explain that change. Is it actually necessary?

nyechiel:
@mkolesnik can you take another look at this one please?

nyechiel:
@mkolesnik ping?

nyechiel:
@mkolesnik ping

nyechiel:
@mkolesnik can you take a look?

.github/workflows/consuming.yml
mkolesnik: Right, it needed to be reversed for the Shipyard directory to be available inside the container that's being created to run the tests, otherwise id thye're reversed or even on the same level it's not available, so the `go replace` wont work.

I'll add this as a comment here and in the commit

skitt (approved):
I checked the other projects and the removal of the `working-directory` parameter on the shared e2e action won’t break anything.

tpantelis (approved):

mkolesnik:
Resolved the conflict (checkout version was bumped)

PR #1149, opened 2023-02-27
Update K8s in E2E, add 1.25

### LINKED ISSUE #1148: Update the default version of K8s used in tests
We’re currently defaulting to 1.23 for tests, but that’s reaching EOL. We should update to a newer version (perhaps whatever version is deployed by kind by default).

maayanf24:
Update default K8s version in E2E to 1.25, since 1.23 is EOL.

Fixes: #1148


skitt (approved):
For the record, 1.25 is chosen because it’s the default version used in kind (currently).

(It would be useful to add this to the commit message.)

tpantelis (approved):

PR #1178, opened 2023-03-13
Remove local-registry flag

aswinsuryan:
Removing the -lr(local-registry) flag from the cluster create command. This would still use local build, but does not push it local registry


aswinsuryan:
Not required

PR #1247, opened 2023-05-10
Fix issues flagged by Shellcheck

skitt:
See individual commits for details.


tpantelis (approved):

sridhargaddam (approved):

skitt:
It seems we need to backport this to allow linting on the 0.15 branch to pass.

PR #1265, opened 2023-06-12
Check deployed images against newly-built images

### LINKED ISSUE #479: Check that the images we deploy are those we built
To avoid testing old images, instead of those we’re supposed to build, we should:

* store the hashes of the images when we build them (_e.g._ in the existing `package/.image...` trackers)
* once our pods are up and running, check that the container hashes match the expected hashes

This can be made generic: we only need to check the `package` contents.

skitt:
See individual commits for details.

This still needs fixes in Helm deployments and an opt-out for upgrade tests.

Fixes: #479.


scripts/shared/deploy.sh
```diff
@@ -130,6 +130,25 @@ fi
...
+        if [ "$deployed" != "$expected" ]; then
+            printf "Image %s is deployed with %s, expected %s\n" "$image" "$deployed" "$expected"
+            image_mismatch=true
+        else
+            printf "Successfully checked image %s, deployed with %s\n" "$image" "$deployed"
+        fi
+    done
+done
+if [ "$image_mismatch" = true ]; then
+    kubectl get pods -A -o json
```
skitt, line 148: This is too verbose and will be simplified before the patch is submitted for final review.

dfarrell07: Does this still need tweaking?

dfarrell07 (approved):

scripts/shared/deploy.sh
dfarrell07: Yes I see it in the logs now, very verbose https://github.com/submariner-io/shipyard/actions/runs/5389288771/jobs/9783083207?pr=1265#step:6:3131

dfarrell07:
Looks like all the Helm jobs are failing

scripts/shared/deploy.sh
```diff
@@ -130,6 +130,25 @@ fi
...
+    for deployed in $(kubectl get pods -A -o json | jq -r '.items[].status.containerStatuses[].imageID' | grep "$image"); do
+        if [ "$deployed" != "$expected" ]; then
+            printf "Image %s is deployed with %s, expected %s\n" "$image" "$deployed" "$expected"
+            image_mismatch=true
+        else
+            printf "Successfully checked image %s, deployed with %s\n" "$image" "$deployed"
+        fi
+    done
+done
+if [ "$image_mismatch" = true ]; then
```
mkolesnik, line 147: This should be run when opted-in, since you might want to deploy with various images when deploying locally or when we're doing upgrade testing

mkolesnik (changes requested):

tpantelis:
Is this still relevant?

skitt:
Yes; it needs improved logging on errors, and a way of disabling the test (as Mike says, and for the Helm charts; there’s probably stuff to fix in the Helm chart builds but that can be done separately).

PR #1273, opened 2023-06-27
Support deploying nexodus

mkolesnik:
Deploy a nexodus service running on KIND and local nexd pods in the Submariner clusters, when creating clusters with `using=nexodus`.


Makefile.dapper
```diff
@@ -12,6 +12,12 @@ export MAKEFLAGS
 export PROJECT
 export USING
 
+# We need the host's actual IP to connect to the nexodus service, otherwise it wont work
+ifneq (,$(filter nexodus,$(USING)))
+HOST_IP = $(shell ip route get 1 | grep -o "src .*" | cut -f2 -d' ')
+export HOST_IP
+endif
```
skitt: Might as well provide this in all cases:

```suggestion
# Determine the host's actual IP (required to connect to the nexodus service)
HOST_IP = $(shell ip route get 1 | grep -o "src .*" | cut -f2 -d' ')
export HOST_IP
```

Makefile.dapper
```diff
@@ -12,6 +12,12 @@ export MAKEFLAGS
 export PROJECT
 export USING
 
+# We need the host's actual IP to connect to the nexodus service, otherwise it wont work
+ifneq (,$(filter nexodus,$(USING)))
+HOST_IP = $(shell ip route get 1 | grep -o "src .*" | cut -f2 -d' ')
```
skitt, line 16: `ip` can produce JSON output:

```suggestion
HOST_IP = $(shell ip --json route get 1 | jq -r '.[0].prefsrc)
```

scripts/shared/resources/nexd.yaml
```diff
@@ -0,0 +1,55 @@
...
+          command:
+            - nexd
+          args:
+            - --username
+            - admin
+            - --password
+            - floofykittens
+            - --insecure-skip-tls-verify
+            - router
+            - --child-prefix
```
sridhargaddam: Now that we understood the reason for the issue, probably you can un-comment the second child_prefix below.

scripts/shared/resources/nexd.yaml
```diff
@@ -0,0 +1,55 @@
+---
+apiVersion: apps/v1
+kind: DaemonSet
```
sridhargaddam: As we are spawning this as a separate Pod and not from the active submariner-gateway pod, we cannot support HA use-cases. Can you add a note somewhere in this PR suggesting that multiple nodes should not be labelled as GW nodes for the use-case to work properly.

scripts/shared/lib/clusters_kind
```diff
@@ -264,11 +267,32 @@ function download_ovnk() {
...
+    curl -JL -o /usr/local/bin/mkcert "https://dl.filippo.io/mkcert/latest?for=linux/amd64"
+    chmod +x /usr/local/bin/mkcert
+
+    echo "Cloning nexodus from source"
+    rm -rf nexodus
+    git clone --single-branch --branch main --depth 1 https://github.com/nexodus-io/nexodus.git
+    git config --global --add safe.directory "$DAPPER_SOURCE"/nexodus
+
+    git -C nexodus apply "$RESOURCES_DIR"/nexodus.patch
+    make -C nexodus run-on-kind
```
sridhargaddam: Since the defaultIPAMv4Cidr ("100.64.0.0/10") used internally by the nexodus service overlaps with the ServiceCIDRs of the Submariner KIND clusters, we have to find a solution for that. 

The nexodus team suggested the following steps.
```
# Create a custom CIDR org

./nexctl --host https://api.try.nexodus.127.0.0.1.nip.io \
    --username admin \
    --password floofykittens \
    organization create \
    --name=zone-a    \
    --cidr=10.195.0.0/24 \
    --cidr-v6 2001:db8:abcd:1234:c000::/66  \
    --description=hub-zone-a

# returns the UUID: 5ed75656-8e28-4900-a81d-4a0608029b06

# Now use that uuid to join that specific org
sudo nexd --insecure-skip-tls-verify --username admin --password floofykittens --organization-id=5ed75656-8e28-4900-a81d-4a0608029b06  router --child-prefix=10.131.0.0/16 https://try.nexodus.127.0.0.1.nip.io

```

Alternately, I think we can just modify (i.e., replace using scripts) the [defaultIPAMv4Cidr](https://github.com/nexodus-io/nexodus/blob/6302ac97535d3c16cdcbc73619342ac6c0aa049d/internal/handlers/organization.go#L23) used  by Nexodus service before running the `make run-on-kind` command. Your thoughts?

scripts/shared/resources/nexd.yaml
mkolesnik: This is just for the POC, I'm not sure we'll be deploying like this if we add proper support

scripts/shared/lib/clusters_kind
mkolesnik: Hmm maybe I'll just patch it out for now since I already patch some of their stuff anyhow, though I prefer they add proper support

Makefile.dapper
mkolesnik: Since this is in `Makefile.dapper` which runs outside the container, I wonder if we should rely on `jq` which forces everyone to have it installed.
Going with `grep | cut` should work for everyone as those are quite ubiquitous, unlike `jq`.

skitt: Fair enough, let’s leave it as-is.

scripts/shared/resources/nexd-kustomization.yaml
```diff
@@ -0,0 +1,37 @@
...
+        value: ${HOST_IP}
+      - op: replace
+        path: /spec/template/spec/containers/0/command
+        value:
+          - /bin/sh
+          - -c
+          - |
+            CAROOT=/etc/nexodus/.certs /bin/mkcert -install
+            /bin/nexd --username=\$USERNAME --password=\$PASSWORD --service-url=\$URL \\
+            router --disable-nat "${NEXODUS_CIDRS[@]}"
```
sridhargaddam, line 34: Do we really need the `--disable-nat`. Its use-case seems to be different.

scripts/shared/lib/clusters_kind
```diff
@@ -284,13 +284,27 @@ function deploy_nexodus() {
...
+    NEXODUS_CERT=$(kubectl --kubeconfig="$DAPPER_OUTPUT"/kubeconfigs/kind-config-nexodus-dev \
+                       get secret -n nexodus nexodus-ca-key-pair -o json | jq -r '.data."ca.crt"')
+    declare -g NEXODUS_CERT
 }
 
 function deploy_nexd() {
-    source "${SCRIPTS_DIR}"/lib/deploy_funcs
-    kubectl create ns nexodus
-    deploy_resource "$RESOURCES_DIR"/nexd.yaml nexodus
+    local NEXODUS_CIDRS=(--child-prefix "${cluster_CIDRs[${cluster}]}" --child-prefix "${service_CIDRs[${cluster}]}")
```
sridhargaddam, line 302: After applying this PR, I see that nexodus pod in both the clusters is configured with the same CIDRs so the connections are not working.

```
  containers:
  - command:
    - /bin/sh
    - -c                                                                                                                                                                                     
    - | 
      CAROOT=/etc/nexodus/.certs /bin/mkcert -install
      /bin/nexd --username=$USERNAME --password=$PASSWORD --service-url=$URL \
      router --disable-nat --child-prefix 10.131.0.0/16 --child-prefix 100.67.0.0/16

```

sridhargaddam: I think this is due to race condition as we are invoking the `deploy_nexd` in parallel but using the same `"$DAPPER_OUTPUT"/nexd/kustomization.yaml` rendered for both the clusters. If we use different output files the problem might be resolved.

```
[08:38:44.353] [dir=shipyard; cl=cluster1; fn=provider_create_cluster]$ delete_cluster_on_fail deploy_nexd
[08:38:44.366] [dir=shipyard; cl=cluster1; fn=delete_cluster_on_fail]$ wait 5664
[08:38:44.366] [dir=shipyard; cl=cluster1; fn=delete_cluster_on_fail]$ deploy_nexd
[08:38:44.377] [dir=shipyard; cl=cluster1; fn=deploy_nexd]$ mkdir -p /go/src/github.com/submariner-io/shipyard/output/nexd
[cluster2] daemon set "dummypod" successfully rolled out
[08:38:44.403] [dir=shipyard; cl=cluster1; fn=deploy_nexd]$ kubectl apply -k /go/src/github.com/submariner-io/shipyard/output/nexd/
[08:38:44.410] [dir=shipyard; cl=cluster1; fn=kubectl]$ command kubectl --context=cluster1 apply -k /go/src/github.com/submariner-io/shipyard/output/nexd/
[08:38:44.416] [dir=shipyard; cl=cluster2; fn=provider_create_cluster]$ delete_cluster_on_fail deploy_nexd
[08:38:44.422] [dir=shipyard; cl=cluster2; fn=delete_cluster_on_fail]$ wait 5691
[08:38:44.422] [dir=shipyard; cl=cluster2; fn=delete_cluster_on_fail]$ deploy_nexd
[08:38:44.427] [dir=shipyard; cl=cluster2; fn=deploy_nexd]$ mkdir -p /go/src/github.com/submariner-io/shipyard/output/nexd
[08:38:44.436] [dir=shipyard; cl=cluster2; fn=deploy_nexd]$ kubectl apply -k /go/src/github.com/submariner-io/shipyard/output/nexd/
[08:38:44.441] [dir=shipyard; cl=cluster2; fn=kubectl]$ command kubectl --context=cluster2 apply -k /go/src/github.com/submariner-io/shipyard/output/nexd/
```

mkolesnik: Good catch I will fix it

scripts/shared/resources/nexd-kustomization.yaml
mkolesnik: > By default, Nexodus network routers perform NAT, specifically, source NAT for devices coming from a Nexodus mesh with a destination of one of the devices not running the Nexodus agent. This enables connectivity to those devices without any configuration on the devices.
> 
> You have the option to disable NAT with --disable-nat which will cause the remote non-Nexodus devices to receive traffic from the Nexodus agent devices without any address translations. This mode requires routes to be added (or redistributed in your network IGP) for hosts in 192.168.1.0/24 to reach Nexodus nodes 100.100.0.0/16 via the Nexodus Network Router eth0 ip of 192.168.1.10.

mkolesnik:
Closing this, we can implement Nexodus support properly, this was just a quick POC

PR #1307, opened 2023-07-26
Bump to k8s.io 1.24.16 dependencies

skitt:
This is the last planned 1.24 release.


tpantelis (approved):

skitt:
It turns out that 0.14 tracks 1.25 (see the operator), so let’s handle this through dependabot.

PR #1395, opened 2023-09-19
Automated backport of #1338: Add OVN IC CI
#1373: Use a specific OVNK commit

skitt:
Backport of #1338 #1373 on release-0.15.

#1338: Add OVN IC CI
#1373: Use a specific OVNK commit

For details on the backport process, see the [backport requests](https://submariner.io/development/backports/) page.

PR #1410, opened 2023-10-02
Fix 1395

skitt:


dfarrell07 (approved):

.github/workflows/testing.yml
```diff
@@ -124,6 +124,7 @@ jobs:
             ovn: ovn
           - extra-toggles: dual-stack
           - extra-toggles: ovn
+          - extra-toggles: ovn-ic
```
sridhargaddam, line 127: We dont run OVN-IC in 0.15 right, so why this change?

scripts/shared/lib/clusters_kind
```diff
@@ -261,11 +257,47 @@ function download_ovnk() {
     )
 }
 
+function prepare_ovn_ic() {
```
sridhargaddam, line 260: If this is only to prepare the env' for OVN to build images from a specific commit, its better to rename it as prepare_for_ovn() or prepare_ovn()

sridhargaddam (changes requested):

.github/workflows/testing.yml
skitt: Good point, I’ll fix https://github.com/submariner-io/shipyard/pull/1395 to drop that.

PR #1431, opened 2023-10-19
Fix gateway failover test

### LINKED ISSUE #1127: Bug in gateway testing framework
Hi,

I have found a commit that introduces a bug.

Changes made in below commit breaks the test if node names in Openshift cluster has a dot "." character.
We are using node names which have a dot "." character and therefore the test fails to find a gateway.

What I understand is that the gateway names (in Gateway resource) created by operator are the same as the node name.

https://github.com/submariner-io/shipyard/commit/d3a497428fd5ddbfe131d86c56ac7beedf36c309

MaxBab:
The gateway failover test fails to locate gateway resource node in On-premise environments.
All environments created in a public cloud platforms sets the node in a Gateway resource without any suffix.
On-premise environment may set node name in a way, that suffix will be defined in Gateway resource.

Search for node in a Gateway resource in both ways - with and without a suffix.

Fixes: https://github.com/submariner-io/shipyard/issues/1127

### DIFF
--- a/test/e2e/framework/gateways.go
+++ b/test/e2e/framework/gateways.go
@@ -37,17 +37,26 @@ var gatewayGVR = &schema.GroupVersionResource{
 	Resource: "gateways",
 }
 
-func (f *Framework) AwaitGatewayWithStatus(cluster ClusterIndex, name, status string) *unstructured.Unstructured {
+func findGateway(cluster ClusterIndex, name string) (*unstructured.Unstructured, error) {
 	gwClient := gatewayClient(cluster)
-	// Set short name without a suffix as exists in the Gateway resource
-	name = strings.Split(name, ".")[0]
+	resGw, err := gwClient.Get(context.TODO(), name, metav1.GetOptions{})
+
+	if apierrors.IsNotFound(err) {
+		// Some environments sets a node in Gateway resource without a suffix
+		resGw, err = gwClient.Get(context.TODO(), strings.Split(name, ".")[0], metav1.GetOptions{})
+	}
+
+	if apierrors.IsNotFound(err) {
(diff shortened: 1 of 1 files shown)


test/e2e/framework/gateways.go
```diff
@@ -39,11 +39,14 @@ var gatewayGVR = &schema.GroupVersionResource{
...
-	// Set short name without a suffix as exists in the Gateway resource
-	name = strings.Split(name, ".")[0]
 	obj := AwaitUntil(fmt.Sprintf("await Gateway on %q with status %q", name, status),
 		func() (interface{}, error) {
 			resGw, err := gwClient.Get(context.TODO(), name, metav1.GetOptions{})
+			// Some environments sets a node in Gateway resource without a suffix
+			shortName := strings.Split(name, ".")[0]
+			if resGw == nil {
+				resGw, err = gwClient.Get(context.TODO(), shortName, metav1.GetOptions{})
+			}
```
tpantelis: It's better to check the return error.  Also to avoid duplication with `AwaitGatewayFullyConnected`, can you put this code in a function `findGateway(cluster ClusterIndex, name string) (*unstructured.Unstructured, error)` ?

```suggestion
			if apierrors.IsNotFound(err) {
			         // Some environments set the Gateway resource name without a suffix
				resGw, err = gwClient.Get(context.TODO(), strings.Split(name, ".")[0], metav1.GetOptions{})
			}
```

tpantelis (changes requested):

test/e2e/framework/gateways.go
MaxBab: Done.

tpantelis (approved):

skitt (approved):

PR #1459, opened 2023-11-06
Fix nil interface{} checks

tpantelis:
PR https://github.com/submariner-io/shipyard/pull/1447 changed places where we check for nil `interface{}` in the callback functions for `AwaitUntil` to cast first. However this can result in an invalid type conversion panic depending on how the value is returned from the operation function. In Go, an interface var contains both a type and value so there are 2 cases for a nil:

 1) The type is non-nil and the value is nil
 2) The type and value are nil

For 1, the `interface{}` var is not actually nil so checking `if result == nil` yields false. Eg,

```
  var s *string
  var i interface{} = s

  if i == nil { // yields false b/c i is typed
  }
```

In this case, we can cast to the expected type and then check for nil.

 ```
 s := i.(*string)
 if s == nil {  // yields true
 }
```

For 2, the `interface{}` var is untyped so nil check yields true:

 ```
 var i interface{} = nil
  if i == nil { // yields true
  }
```

So in cases where we directly return nil from an operation function, the nil check in the result function is valid.

The actual reason for #1447 was b/c of a nil pointer panic in `AwaitGatewaysWithStatus`. This function was previously changed to call `findGateway` and return its values. However `findGateway` returns `*unstructured.Unstructured` so now the operation function returned a typed `interface{}` var. So if `findGateway` returns a nil `*unstructured.Unstructured` value, the `result == nil` check in the result function yields false. This resulted in the panic when `result` was later casted and accessed.

So we basically need to undo #1447. For `AwaitGatewayWithStatus` and `AwaitGatewayFullyConnected`, check the error from `findGateway` and return a nil `interface{}` value.

### DIFF
--- a/test/e2e/framework/clusterglobalegressip.go
+++ b/test/e2e/framework/clusterglobalegressip.go
@@ -52,12 +52,11 @@ func AwaitAllocatedEgressIPs(client dynamic.ResourceInterface, name string) []st
 			return resGip, err
 		},
 		func(result interface{}) (bool, string, error) {
-			obj := result.(*unstructured.Unstructured)
-			if obj == nil {
+			if result == nil {
 				return false, fmt.Sprintf("Egress IP resource %q not found yet", name), nil
 			}
 
-			globalIPs := getGlobalIPs(obj)
+			globalIPs := getGlobalIPs(result.(*unstructured.Unstructured))
 			if len(globalIPs) == 0 {
 				return false, fmt.Sprintf("Egress IP resource %q exists but allocatedIPs not available yet", name), nil
 			}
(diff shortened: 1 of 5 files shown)


skitt (approved):

sridhargaddam (approved):

tpantelis:
The globalnet w/ helm job is failing:

`E1106 16:18:52.373476       1 leaderelection.go:327] error retrieving resource lock submariner-operator/submariner-globalnet-lock: leases.coordination.k8s.io "submariner-globalnet-lock" is forbidden: User "system:serviceaccount:submariner-operator:submariner-globalnet" cannot get resource "leases" in API group "coordination.k8s.io" in the namespace "submariner-operator"`

Need to update the helm chart to add `submariner-globalnet` permissions. This issue is orthogonal so shouldn't block this PR so force merging.

PR #1464, opened 2023-11-07
Remove helm jobs

tpantelis:
The helm jobs use the latest released charts version, which now is _0.16.1_, which doesn't seem right for the _devel_ branch. So remove the jobs and let submariner-charts test helm.


skitt (approved):

tpantelis:
I guess a reason to keep these jobs would be to test the actual script, since it lives in shipyard, to ensure there's no breakage.

PR #1466, opened 2023-11-08
Bump to Fedora 39

skitt:
This brings Go 1.21. Go 1.21 no longer ships .a libraries, so the corresponding clean-up is removed.

Python is upgraded to 3.12, which breaks j2cli; this switches to the packaged python3-jinja2-cli instead. The latter has progressed since j2cli was created, and should support all the j2cli features needed in Submariner (more precisely, for OVN).

The official GitHub client is available as a package, install that instead of the upstream tarball.


package/Dockerfile.shipyard-dapper-base
```diff
@@ -43,20 +43,18 @@ ENV HOST_ARCH=${DAPPER_HOST_ARCH} ARCH=${DAPPER_HOST_ARCH} PATH=/go/bin:/root/.l
 # We remove:
 # - DNF cache
 # - Any unnecessary packages and executables
-# - Precompiled packages in go (see https://github.com/golang/go/issues/47257)
 RUN dnf -y install --nodocs --setopt=install_weak_deps=False \
                    gcc git-core curl moby-engine make golang kubernetes-client \
-                   findutils moreutils upx jq gitlint procps-ng python3-setuptools \
-                   qemu-user-static python3-pip skopeo file unzip && \
-    pip install j2cli[yaml] --user && \
```
sridhargaddam, line 51: Please update Line25

sridhargaddam (changes requested):

sridhargaddam (approved):

dfarrell07 (approved):

skitt:
The switch to `jinja2` breaks OVN deployments; that can probably be worked around externally. The best solution would be to migrate OVNKubernetes to `jinja2` as well, but before we’d be able to use that we’d have to fix the deployment issues we run into when trying to use newer versions of OVNK...

tpantelis (approved):

PR #1497, opened 2023-12-06
Refactor Shipyard to be consumable by other projects

mkolesnik:
These commits refactor Shipyard and break it down into smaller more basic components which are Submariner agnostic.
This allows it to be used in projects outside the Submariner org, i.e. https://github.com/skupperproject/skupper/pull/1317


scripts/shared/deploy.sh
```diff
@@ -123,6 +142,7 @@ declare_kubeconfig
 
 load_library deploy DEPLOYTOOL
 deploytool_prereqs
+[[ "$PROVIDER" == kind ]] || run_all_clusters schedule_dummy_pod
```
skitt: Isn’t this the wrong way round? If `schedule_dummy_pod` is supposed to run with kind,

```
[[ "$PROVIDER" != kind ]] || run_all_clusters schedule_dummy_pod
```

skitt (changes requested):

skitt (approved):

tpantelis (approved):

PR #1498, opened 2023-12-06
Group dependabot updates in a single PR

tpantelis:
...per package-ecosystem.


skitt (approved):
I wonder if it wouldn’t be better to group Go dependency updates by dependency family (_e.g._ Kubernetes), but we can revisit this later if the grouped PRs cause issues.

mkolesnik (approved):

PR #1499, opened 2023-12-11
Fix linting image build failure due to PEP 668

mkolesnik:
Pip had a check enabled that fails the install due to packages being managed by the OS.
The added flag ignores this check.

Otherwise, the build fails with:
```
11.85 × This environment is externally managed
11.85 ╰─>
11.85     The system-wide python installation should be maintained using the
11.85     system package manager (apk) only.
...
11.85 note: If you believe this is a mistake, please contact your Python
  installation or OS distribution provider. You can override this, at the risk
  of breaking your Python installation or OS, by passing
  --break-system-packages.
11.85 hint: See PEP 668 for the detailed specification.
```


mkolesnik:
Error can be seen in latest linting jobs, e.g. https://github.com/submariner-io/shipyard/actions/runs/7129355323/job/19507164695?pr=1497#step:3:325

skitt (approved):

tpantelis (approved):

PR #1501, opened 2023-12-14
Remove helm from the consuming E2E GHA

tpantelis:


skitt (approved):

.github/workflows/consuming.yml
```diff
@@ -19,17 +19,10 @@ jobs:
       fail-fast: false
       matrix:
         project: ['admiral', 'lighthouse', 'subctl', 'submariner', 'submariner-operator']
-        deploytool: ['operator', 'helm']
+        deploytool: ['operator']
```
mkolesnik: Since we don't plan to override this in any permutation, it's better to inline this in step itself (L#65).

As a side note, operator is the default anyhow, so this can be dropped altogether.

mkolesnik (changes requested):

.github/workflows/consuming.yml
tpantelis: Done

mkolesnik (approved):

PR #1528, opened 2024-01-18
Update OVN Image and commit

aswinsuryan:
OVN now builds and pushes image with every commit. So changed to use OVN Image than buildng locally.

Added pip to shipyard-dapper-base as it is required for
kind.sh to set up the Jinja2 tools

### DIFF
--- a/package/Dockerfile.shipyard-dapper-base
+++ b/package/Dockerfile.shipyard-dapper-base
@@ -44,7 +44,7 @@ ENV HOST_ARCH=${DAPPER_HOST_ARCH} ARCH=${DAPPER_HOST_ARCH} PATH=/go/bin:/root/.l
 # - Any unnecessary packages and executables
 RUN dnf -y install --nodocs --setopt=install_weak_deps=False \
                    gcc git-core curl moby-engine make golang kubernetes-client \
-                   findutils moreutils upx jq gitlint procps-ng \
+                   findutils moreutils upx jq gitlint procps-ng pip\
                    qemu-user-static python3-jinja2-cli skopeo file unzip gh && \
     rpm -e --nodeps containerd && \
     rpm -qa "selinux*" | xargs -r rpm -e --nodeps && \
--- a/scripts/shared/lib/clusters_kind
+++ b/scripts/shared/lib/clusters_kind
@@ -243,20 +243,16 @@ function provider_failed() {
 function prepare_ovn() {
     export OVN_IMAGE="localhost:5000/ovn-daemonset-f:latest"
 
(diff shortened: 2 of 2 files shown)


tpantelis (approved):

scripts/shared/lib/clusters_kind
```diff
@@ -243,20 +243,16 @@ function provider_failed() {
...
-    echo "Building ovn-kubernetes with interconnect (OVN-IC) from source"
-    echo "This will become unnecessary if OVN CI image publishing is fixed"
-    echo "https://github.com/ovn-org/ovn-kubernetes/actions/workflows/docker.yml"
     rm -rf ovn-kubernetes
     git clone https://github.com/ovn-org/ovn-kubernetes
     pushd ovn-kubernetes || exit
-    git checkout d462cf088c44ff501e5fcf746e9163193afa2dc4
 
-    make -C go-controller
+    # When updating commit, Update the OVN_SRC_IMAGE to the corressponding commit
```
skitt, line 250: Do we still need to clone the ovn-kubernetes repository?

package/Dockerfile.shipyard-dapper-base
```diff
@@ -44,7 +44,7 @@ ENV HOST_ARCH=${DAPPER_HOST_ARCH} ARCH=${DAPPER_HOST_ARCH} PATH=/go/bin:/root/.l
 # - Any unnecessary packages and executables
 RUN dnf -y install --nodocs --setopt=install_weak_deps=False \
                    gcc git-core curl moby-engine make golang kubernetes-client \
-                   findutils moreutils upx jq gitlint procps-ng \
+                   findutils moreutils upx jq gitlint procps-ng pip\
```
skitt, line 47: What do we need pip for?

skitt (changes requested):

scripts/shared/lib/clusters_kind
skitt: Oh, for the `kind.sh` script.

package/Dockerfile.shipyard-dapper-base
skitt: I imagine this is required so that `kind.sh` can set up the new Jinja2 tools; please add the reason to the commit message.

aswinsuryan: Done.

scripts/shared/lib/clusters_kind
aswinsuryan: ya that is right.

skitt (approved):

PR #1538, opened 2024-01-25
Default to K8s 1.28 in e2e

skitt:
Other GHAs were changed to default to 1.28, but the custom GHA used throughout Submariner wasn't updated and still defaulted to 1.25. This fixes that.

### DIFF
--- a/gh-actions/e2e/action.yaml
+++ b/gh-actions/e2e/action.yaml
@@ -9,7 +9,7 @@ inputs:
   k8s_version:
     description: 'Version of Kubernetes to use for clusters'
     required: false
-    default: '1.25'
+    default: '1.28'
   using:
     description: 'Various options to pass via using="..."'
     required: false


gh-actions/e2e/action.yaml
```diff
@@ -9,7 +9,7 @@ inputs:
   k8s_version:
     description: 'Version of Kubernetes to use for clusters'
     required: false
-    default: '1.25'
+    default: '1.28'
```
tpantelis, line 12:
```suggestion
    default: '1.29'
```

tpantelis (changes requested):

tpantelis:
Shouldn't we default to 1.29?

skitt:
> Shouldn't we default to 1.29?

Not if we want to be able to run the e2e jobs — we can update this once we get a new version of kind.

tpantelis (approved):

skitt:
#1539 should fix the cleanup failures.

aswinsuryan (approved):

PR #1539, opened 2024-01-25
Prevent parallel cleanup of kind clusters

skitt:
kind uses a lock file to protect kubeconfig when it updates it after stopping a cluster, but it fails if the lock file already exists. Avoid this by cleaning up serially.


tpantelis:
```
In scripts/shared/lib/cleanup_kind line 10:
    PARALLEL=false
    ^------^ SC2034 (warning): PARALLEL appears unused. Verify use (or export if used externally).
```

tpantelis (approved):

aswinsuryan (approved):

PR #1622, opened 2024-04-18
Use cloud-provider-kind instead of MetalLB

skitt:
kind has a cloud-provider which can take care of LoadBalancer-type services; this works fine with LB-mode Submariner, so use that instead of MetalLB.

Currently the command is installed as cmd, which is why it is renamed during container construction. This is fixed for the next release of the cloud-provider.

cloud-provider-kind runs in the foreground, so this wraps it in a container run alongside the kind clusters, on the host network. The container is stopped on cleanup.

The dependency version will be tracked using dependabot.


dfarrell07 (approved):
Nice!

tpantelis (approved):

skitt:
Switching to draft because this leaves haproxy containers running (and shouldn’t).

yboaron:
I think we should consider keeping MetalLB as well since it is the default LB provider for baremetal. Do you think we can support both methods?

skitt:
> I think we should consider keeping MetalLB as well since it is the default LB provider for baremetal. Do you think we can support both methods?

We probably can, yes.

PR #1655, opened 2024-05-30
Revert to Go 1.21 and Kubernetes 1.29

skitt:
This only touches the main module; the tools can remain on Go 1.22 for our purposes.

This must only be merged on the release-0.18 branch once it has been created.

Depends on https://github.com/submariner-io/releases/issues/923


tpantelis:
We decided that we were going to do this on the 0.18 release branch once we do rc0 next week. That way we leave devel as is so we don't have to re-upgrade to Go 1.22 et al later.

skitt:
> We decided that we were going to do this on the 0.18 release branch once we do rc0 next week. That way we leave devel as is so we don't have to re-upgrade to Go 1.22 et al later.

I know, see the PR description.

PR #1727, opened 2024-09-22
Add support for Kind IPV6 clusters deployment

yboaron:
Run 'make clusters using=ipv6-stack' to deploy IPV6 Kind clusters.
This capability will be useful for verifying Submariner's IPv6 support.

### DIFF
--- a/Makefile.clusters
+++ b/Makefile.clusters
@@ -14,7 +14,7 @@ K8S_VERSION ?= 1.31
 METALLB_VERSION ?= 0.13.5
 OLM_VERSION ?= v0.18.3
 PROVIDER ?= kind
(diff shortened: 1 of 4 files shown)


tpantelis:
Should we add "ipv6_stack" to the Deployment GHA matrix?

yboaron:
> Should we add "ipv6_stack" to the Deployment GHA matrix?

Do you mean to update [this](https://github.com/submariner-io/shipyard/blob/5882e558eb5cb69769bb7b370243e9282f0f707e/.github/workflows/testing.yml#L123) ? 

Does it make sense to add it to GHA matrix before V6 is supported on Submariner side?

tpantelis:
> Does it make sense to add it to GHA matrix before V6 is supported on Submariner side?

yeah I guess not.

tpantelis (approved):

dfarrell07 (approved):

PR #1728, opened 2024-09-26
Bump to Fedora 41

skitt:
This provides Go 1.23.

Compared to Fedora 40 images, this reduces the shipyard-dapper-base image size from 775MiB to 723MiB. Only installing the QEMU binaries that are needed further reduces that to 664MiB.


skitt:
Fedora 41 is still in beta so this might not be advisable right now but the risk should be low.

tpantelis:
> Fedora 41 is still in beta so this might not be advisable right now but the risk should be low.

Do you feel confident to merge this for RC0 on Mon?

skitt:
> > Fedora 41 is still in beta so this might not be advisable right now but the risk should be low.
> 
> Do you feel confident to merge this for RC0 on Mon?

If we don’t need Go 1.23 for rc0, probably not.

tpantelis:
> > > Fedora 41 is still in beta so this might not be advisable right now but the risk should be low.
> > 
> > 
> > Do you feel confident to merge this for RC0 on Mon?
> 
> If we don’t need Go 1.23 for rc0, probably not.

We don't - let's hold off.

skitt:
Fedora 41 has been released.

tpantelis (approved):

dfarrell07 (approved):

PR #1730, opened 2024-09-30
Check unmerged commits in dependencies

skitt:
This adds a script which examines submariner-io dependencies with versions referencing untagged git commits, that is to say commits which haven't been released. It ensures that any such dependencies point to a commit in the relevant branch, to avoid merging PRs with pointers to unmerged commits. Since the project rebases on PR merge, the commit hashes change, and dependent PRs can end up being merged with a reference to an in-development commit; that can break local builds.

Since Shipyard doesn't have any submariner-io dependencies, this isn't integrated into CI yet; it will be used in other Submariner projects.

See https://github.com/submariner-io/submariner-operator/pull/3237 for a check involving this (and correctly failing).

### DIFF
--- a/scripts/shared/check-non-release-versions.sh
+++ b/scripts/shared/check-non-release-versions.sh
@@ -1,0 +1,26 @@
+#!/bin/bash
+
+tmpdir=$(mktemp -d)
+trap 'rm -rf $tmpdir' EXIT
+
+# List all submariner-io dependencies with a - in their version
+# We're looking for versions pointing to commits, of the form
+# vX.Y.Z-0.YYYYMMDDhhmmss-hash
+failed=0
+shopt -s lastpipe
+GOWORK=off go list -m -mod=mod -json all |
+    jq -r 'select(.Path | contains("/submariner-io/")) | select(.Main != true) | select(.Version | contains ("-")) | select(.Version | length > 14) | "\(.Path) \(.Version)"' |
+    while read -r project version; do
+        cd "$tmpdir" || exit 1
+        git clone "https://$project"
+        cd "${project##*/}" || exit 1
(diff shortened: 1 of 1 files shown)


tpantelis (approved):

scripts/shared/check-non-release-versions.sh
```diff
@@ -0,0 +1,22 @@
...
+
+tmpdir=$(mktemp -d)
+trap 'rm -rf $tmpdir' EXIT
+
+# List all submariner-io dependencies with a - in their version
+# We're looking for versions pointing to commits, of the form
+# vX.Y.Z-0.YYYYMMDDhhmmss-hash
+GOWORK=off go list -m -mod=mod -json all | \
+    jq -r 'select(.Path | contains("/submariner-io/")) | select(.Main != true) | select(.Version | contains ("-")) | select(.Version | length > 14) | "\(.Path) \(.Version)"' | \
+    while read -r project version; do
```
dfarrell07, line 13: Nit: Should there be indentation inside this loop?

Which highlights that [shellcheck doesn't do format checking](https://github.com/koalaman/shellcheck/issues/2463). They recommend [shfmt](https://github.com/mvdan/sh). I can look into it if other people think it would be helpful.

skitt: There should indeed, thanks! `shfmt` could be useful; our indentation in particular is inconsistent, running `shfmt -i 4` would fix that.

dfarrell07: Raised https://github.com/submariner-io/shipyard/issues/1737 to track.

dfarrell07 (approved):

PR #1751, opened 2024-11-18
Add support for Kind OVNK IPV6 deployment

yboaron:
Depends on: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4955

### DIFF
--- a/package/Dockerfile.shipyard-dapper-base
+++ b/package/Dockerfile.shipyard-dapper-base
@@ -4,7 +4,7 @@ FROM fedora:41
 # good balance between compression time and resulting image size.
 ARG UPX_LEVEL=-5
 ENV DAPPER_HOST_ARCH=amd64 SHIPYARD_DIR=/opt/shipyard SHELL=/bin/bash \
-    DAPPER_RUN_ARGS="--net=kind"
+    DAPPER_RUN_ARGS="--net=kind --cap-add=NET_ADMIN --privileged=true"
 ENV HOST_ARCH=${DAPPER_HOST_ARCH} ARCH=${DAPPER_HOST_ARCH} PATH=/go/bin:/root/.local/bin:/usr/local/go/bin:$PATH \
     GOLANG_ARCH_amd64=amd64 GOLANG_ARCH_arm=armv6l GOLANG_ARCH=GOLANG_ARCH_${DAPPER_HOST_ARCH} \
     GOPATH=/go GO111MODULE=on GOPROXY=https://proxy.golang.org \
--- a/scripts/shared/lib/clusters_kind
+++ b/scripts/shared/lib/clusters_kind
@@ -264,6 +264,10 @@ function prepare_ovn() {
 
     # When updating commit, Update the OVN_SRC_IMAGE to the corressponding commit
     git checkout 24b0ae73a996e409bfefad7b90cb42224e34be54
+    git fetch origin pull/4955/head:kind_ipv6_fix
+    git config user.email "fufu@example.com"
+    git config user.name "fufu"
+    git cherry-pick kind_ipv6_fix
     local OVN_SRC_IMAGE="ghcr.io/ovn-org/ovn-kubernetes/ovn-kube-u:master@sha256:ba102783d520f0474e5c7dd5f2a0a1dce0ec2bda6cd42ac547621892e57c25e2"
 
     docker pull "${OVN_SRC_IMAGE}"
--- a/scripts/shared/lib/clusters_kind
+++ b/scripts/shared/lib/clusters_kind
@@ -196,12 +196,24 @@ EOF
 
 function deploy_kind_ovn(){
     export K8S_VERSION
-    export NET_CIDR_IPV4="${cluster_CIDRs[${cluster}]}"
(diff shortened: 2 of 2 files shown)


yboaron:
@mkolesnik , QQ: I'm getting [1] error when trying to deploy Kind clusters with OVNK and IPV6 with this PR(with **make clusters using=ipv6-stack,ovn**), while running directly **./ovn-kubernetes/contrib/kind.sh ....**  things looks fine, any idea ?

[1] 
`[cluster2] ++ sysctl net.ipv6.conf.all.forwarding
[cluster1] ++ sysctl net.ipv6.conf.all.forwarding
[cluster2] ++ awk '{print $3}'
[cluster1] ++ awk '{print $3}'
[cluster2] + TMPVAR=0
[cluster2] + echo 'net.ipv6.conf.all.forwarding is equal to 0'
[cluster2] net.ipv6.conf.all.forwarding is equal to 0
[cluster2] + '[' 0 '!=' 1 ']'
[cluster1] + TMPVAR=0
[cluster2] + '[' true == true ']'
[cluster2] + sudo sysctl -w net.ipv6.conf.all.forwarding=1
[cluster1] + echo 'net.ipv6.conf.all.forwarding is equal to 0'
[cluster1] net.ipv6.conf.all.forwarding is equal to 0
[cluster1] + '[' 0 '!=' 1 ']'
[cluster1] + '[' true == true ']'
[cluster1] + sudo sysctl -w net.ipv6.conf.all.forwarding=1
[cluster2] sysctl: permission denied on key "net.ipv6.conf.all.forwarding"
[cluster1] sysctl: permission denied on key "net.ipv6.conf.all.forwarding"
[cluster2] Failed to run './ovn-kubernetes/contrib/kind.sh -ov localhost:5000/ovn-daemonset-f:latest -cn cluster2 -ric -n4 -i6 -sw -lr `

package/Dockerfile.shipyard-dapper-base
```diff
@@ -4,7 +4,7 @@ FROM fedora:41
 # good balance between compression time and resulting image size.
 ARG UPX_LEVEL=-5
 ENV DAPPER_HOST_ARCH=amd64 SHIPYARD_DIR=/opt/shipyard SHELL=/bin/bash \
-    DAPPER_RUN_ARGS="--net=kind"
+    DAPPER_RUN_ARGS="--net=kind --cap-add=NET_ADMIN --privileged=true"
```
yboaron, line 7: We only need the extra networking permissions to container for OVN/IPV6 deployment.

@mkolesnik @skitt , is it possible to set DAPPER_RUN_ARGS content based on CNI, IP family env vars ?

skitt: `DAPPER_RUN_ARGS` is a bit of a cheat — it’s not used as an environment variable as such inside the container, so nothing running in the container can set it meaningfully. The `.dapper` scripts extracts the value of the variable from the container image before running it.

However since we decide what `.dapper` does, anything can be done. Could you explain what the exacts criteria (as in, which values in which files) are and which containers need to run with privileges?

yboaron: For IPV6 OVN deployment (**make clusters using=ipv6-stack,ovn**) OVN-K script [1] sets net.ipv6.conf.all.forwarding and sudo sysctl -w net.ipv6.conf.all.disable_ipv6 .
so, container should run with net privileges.

[1] 
https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/contrib/kind.sh#L642

skitt: It only does so if the values don’t match what it expects. Unfortunately [GH doesn’t enable IPv6 forwarding](https://github.com/submariner-io/shipyard/actions/runs/12792227597/job/35662173778?pr=1813) by default, so changing that would be necessary.

However it would be possible to do this as an extra step before running the jobs, instead of making the containers themselves privileged…

skitt: See https://github.com/submariner-io/shipyard/actions/runs/12792280698/job/35662349880?pr=1813

yboaron: > It only does so if the values don’t match what it expects

Yep, without net privileges we can't even read kernel's ipv6 config (sysctl net.ipv6.conf.all.forwarding)

> However it would be possible to do this as an extra step before running the jobs, instead of making the containers themselves privileged…

Maybe  I'm missing something here, this idea will be useful only on GH,  will it be possible to deploy Kind+IPV6,OVN locally without making the containers privileged?

skitt: > Yep, without net privileges we can't even read kernel's ipv6 config (sysctl net.ipv6.conf.all.forwarding)

That’s surprising, it works for me and should work fine as long as `/proc/sys` is mounted in the container (`sysctl` literally only reads from that).

skitt: For local deployments, we can check the settings early and give an appropriate error, same as for other `sysctl` settings. In most cases I expect developer machines will be set up with IPv6 enabled and forwarded anyway.

skitt:
Would it be possible to add an IPv6 deployment in the deployment tests, to check that this does what is expected?

PR #1766, opened 2024-12-11
Switch to Fedora for nettest and use packaged netperf

skitt:
netperf currently fails to build with the latest Alpine images, even with _GNU_SOURCE fixes.

To allow building the nettest image again, this switches to Fedora, where netperf is available as a package (so we don't need to deal with building it). The image size is more than doubled, from 24.2MiB with Alpine to 59.8MiB with Fedora.

This ends up with nmap ncat rather than busybox ncat, so the nettest scripts are adjusted to suit, and a busybox nc emulation script is included for other users of nc.

### DIFF
--- a/package/Dockerfile.nettest
+++ b/package/Dockerfile.nettest
@@ -1,32 +1,20 @@
-FROM alpine
+ARG FEDORA_VERSION=41
 
-WORKDIR /app
+FROM --platform=${BUILDPLATFORM} fedora:${FEDORA_VERSION} AS base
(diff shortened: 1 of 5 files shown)


skitt:
This probably needs the `nc` compatibility script for `subctl`.

skitt:
Hmm, the “nginx demo” isn’t working…

tpantelis (approved):

vthapar (approved):

PR #1773, opened 2024-12-17
Listener pod ncat fix

vthapar:


test/e2e/framework/network_pods.go
```diff
@@ -278,7 +278,7 @@ func (np *NetworkPod) buildTCPCheckListenerPod() {
 						"for i in $(seq 1 $BUFS_NUM);" +
 							" do echo [dataplane] listener says $SEND_STRING;" +
 							" done" +
-							" | nc -w $CONN_TIMEOUT -l -v -p $LISTEN_PORT -s 0.0.0.0 >/dev/termination-log 2>&1",
+							" | nc -w ${CONN_TIMEOUT}s -l -v 0.0.0.0 $LISTEN_PORT >/dev/termination-log 2>&1",
```
skitt, line 281:
```suggestion
							" | ncat -w ${CONN_TIMEOUT}s -l -v 0.0.0.0 $LISTEN_PORT >/dev/termination-log 2>&1",
```

test/e2e/framework/network_pods.go
```diff
@@ -328,7 +328,7 @@ func (np *NetworkPod) buildTCPCheckConnectorPod() {
 						"for in in $(seq 1 $BUFS_NUM);" +
 							" do echo [dataplane] connector says $SEND_STRING; done" +
 							" | for i in $(seq $CONN_TRIES);" +
-							" do if nc -v $REMOTE_IP $REMOTE_PORT -w $CONN_TIMEOUT;" +
+							" do if nc -v $REMOTE_IP $REMOTE_PORT -w ${CONN_TIMEOUT}s;" +
```
skitt, line 331:
```suggestion
							" do if ncat -v $REMOTE_IP $REMOTE_PORT -w ${CONN_TIMEOUT}s;" +
```

skitt (changes requested):
Since we’re switching to `ncat` syntax, it’s safer to invoke `ncat` directly — otherwise in the next PR after this, once images are rebuilt with the `nc` compatibility wrapper first on the path, this will fail again.

test/e2e/framework/network_pods.go
vthapar: I want to see if it fails or not. If it works, I'd revert to `nc` syntax so it works with older images, as you pointed out earlier. It will also be a good way to see if script works or not.

PR #1805, opened 2025-01-06
Revert to busybox nc for metricsproxy

skitt:
nmap-ncat is causing issues in metricsproxy with globalnet:

Ncat: Version 7.92 ( https://nmap.org/ncat )
Ncat: Listening on :::8081
Ncat: Listening on 0.0.0.0:8081
Ncat: Connection from 10.129.0.19.
Ncat: Connection from 10.129.0.19:45064.
Ncat: assertion failed: count <= INT_MAX QUITTING.

It's not clear how to fix that, so this installs busybox and reverts to busybox nc instead. To avoid having too many changes in nettest, nmap-ncat is preserved for other uses.

### DIFF
--- a/package/Dockerfile.nettest
+++ b/package/Dockerfile.nettest
@@ -9,7 +9,7 @@ COPY scripts/shared/dnf_install /
 
 RUN /dnf_install -a ${TARGETPLATFORM} -v ${FEDORA_VERSION} -r /output/nettest \
     glibc bash glibc-minimal-langpack coreutils-single libcurl-minimal \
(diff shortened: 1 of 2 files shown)


tpantelis (approved):

maayanf24 (approved):

dfarrell07:
@skitt You wanted to backport this I see with the label. Is that still needed? Which branches?

skitt:
> @skitt You wanted to backport this I see with the label. Is that still needed? Which branches?

It’s been taken care of on all branches.

PR #1809, opened 2025-01-13
Check the deployed version of local components

### LINKED ISSUE #479: Check that the images we deploy are those we built

skitt:
In projects building deployed components, declaring LOCAL_COMPONENTS allows deploy.sh to check that the deployed version used in tests matches the expected version.

Fixes: #479

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,11 +1,14 @@
 BASE_BRANCH ?= devel
 OCM_BASE_BRANCH ?= main
 IMAGES ?= shipyard-dapper-base shipyard-linting nettest
+LOCAL_COMPONENTS := submariner-metrics-proxy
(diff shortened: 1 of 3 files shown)


skitt:
This is expected to fail, that proves that it works. Merging #1808 will fix the failure.

tpantelis (approved):

dfarrell07 (approved):

skitt:
What should we do here — disable the checks on Helm, or fix Helm to deploy the newly-built artifacts?

PR #1863, opened 2025-03-06
Create configmap to control Submariner packetfilter driver

yboaron:
We want to start testing Submariner with nftables, to do so we need to change RouteAgent and GlobalNet to dynamically configure the required packet filter driver (iptables or nftables), as it is currently hardcoded into iptables.

This PR creates a configmap with the desired use-nftables value.

### DIFF
--- a/scripts/shared/deploy.sh
+++ b/scripts/shared/deploy.sh
@@ -136,6 +136,18 @@ function schedule_dummy_pod() {
     deploy_resource "${RESOURCES_DIR}"/dummypod.yaml "$ns"
 }
 
+function deploy_subm_global_cm() {
+    local ns="submariner-operator"
+    source "${SCRIPTS_DIR}"/lib/deploy_funcs
+
+    echo "Creating the ${ns} namespace..."
+    kubectl create namespace "${ns}" || :
+    echo Setting up submariner global configmap...
+    NFTABLES="${NFTABLES:-false}"
+    deploy_resource "${RESOURCES_DIR}"/sm-global-cm.yaml "$ns"
+}
+
+
 ### Main ###
 
 load_settings
@@ -153,7 +165,7 @@ declare_kubeconfig
 load_library deploy DEPLOYTOOL
 deploytool_prereqs
 [[ "$PROVIDER" != kind ]] || run_all_clusters schedule_dummy_pod
-
+run_all_clusters deploy_subm_global_cm
 run_if_defined pre_deploy
 
 with_context "$broker" setup_broker
--- a/scripts/shared/resources/sm-global-cm.yaml
+++ b/scripts/shared/resources/sm-global-cm.yaml
@@ -1,0 +1,8 @@
+---
+apiVersion: v1
+data:
+  use-nftables: \"${NFTABLES}\"
+kind: ConfigMap
+metadata:
+  name: submariner-global
+  namespace: ${ns}


scripts/shared/deploy.sh
```diff
@@ -136,6 +136,23 @@ function schedule_dummy_pod() {
...
+
+    echo "Creating the ${ns} namespace..."
+
+    if [[ -z "${NFTABLES}" ]]; then
+        export NFTABLES="\"false\""
+    else
+        export NFTABLES="\"${NFTABLES}\""
+    fi
+    kubectl create namespace "${ns}" || :
+    deploy_resource "${RESOURCES_DIR}"/nftables-cm.yaml "$ns"
```
skitt: Let’s keep logs next to the event they’re describing. There’s also no need to re-set `NFTABLES` if it’s already set, or to export it for `deploy_resource`:

```suggestion
    echo "Creating the ${ns} namespace..."
    kubectl create namespace "${ns}" || :

    echo Setting up the packet filter configmap...
    if [[ -z "${NFTABLES}" ]]; then
        NFTABLES=false
    fi
    deploy_resource "${RESOURCES_DIR}"/nftables-cm.yaml "$ns"
```

scripts/shared/resources/nftables-cm.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
```
skitt: If this is for packet filter configuration, let’s name the file accordingly; it’s used for nftables currently but doesn’t seem to be limited to that.

skitt (changes requested):

scripts/shared/resources/nftables-cm.yaml
yboaron: Should we have a generic submariner customization configmap instead of a specific cm for packetfilter?  maybe other submariner components can extend it if needed

tpantelis: I was actually thinking we would create separate general `ConfigMaps` for **route-agent** and **globalnet** that would be used for other config as well in the future. 

But in this case, the setting applies to both so having one `ConfigMap` would be easier. I suggest we generalize it though and name it "_global_"  so it could be used for general settings (eg leader election config) that apply to some or all components.

scripts/shared/resources/nftables-cm.yaml
```diff
@@ -0,0 +1,7 @@
...
+data:
+  nftablesEnabled: "${NFTABLES}"
```
tpantelis:
```suggestion
  use-nftables: "${NFTABLES}"
```

tpantelis (approved):

scripts/shared/resources/sm-global-cm.yaml
```diff
@@ -0,0 +1,8 @@
+---
+apiVersion: v1
+data:
+  use-nftables: "${NFTABLES}"
```
skitt: Since the quotes cause issues (and are required in the final YAML), I think it’s best to enforce them here:

```suggestion
  use-nftables: \"${NFTABLES}\"
```

This has the benefit of working without requiring specific quoting in the `NFTABLES` variable itself, and avoids misleading readers with unescaped quotes in this file (which seems to address the problem but doesn’t actually work).

scripts/shared/deploy.sh
```diff
@@ -136,6 +136,18 @@ function schedule_dummy_pod() {
...
 }
 
+function deploy_subm_global_cm() {
+    local ns="submariner-operator"
+    source "${SCRIPTS_DIR}"/lib/deploy_funcs
+
+    echo "Creating the ${ns} namespace..."
+    kubectl create namespace "${ns}" || :
+    echo Setting up submariner global configmap...
+    export NFTABLES="\"${NFTABLES:-false}\""
```
skitt:
```suggestion
    NFTABLES="${NFTABLES:-false}"
```

in line with the fix in the YAML file. Exporting the variable isn’t useful.

skitt (changes requested):

skitt (approved):

skitt:
Thanks @yboaron! YAML quoting rules are a pain

PR #1891, opened 2025-03-19
Add a demo mode for deployments

skitt:
This runs a standard deployment but leaves the clusters running with netshoot and "nginx-demo" pods deployed everywhere. This makes it easier to deploy netshoot and the server pods with the correct image overrides.


tpantelis:
What is the purpose of this? Also I only see "nginx-demo" deployed with demo, not netshoot.

skitt:
> What is the purpose of this? Also I only see "nginx-demo" deployed with demo, not netshoot.

I’ve updated the commit message — I’ve been working on some NetworkPolicy examples, and got tired of manually deploying the various pods. Having `make demo` will make it easier to write the corresponding docs too.

The netshoot pod is deployed here: https://github.com/submariner-io/shipyard/pull/1891/files#diff-12a4268ae92fef1d11c835409a46848a6484afdb41ebfc507a5e91e4cea02a91R69

tpantelis:
> The netshoot pod is deployed here: https://github.com/submariner-io/shipyard/pull/1891/files#diff-12a4268ae92fef1d11c835409a46848a6484afdb41ebfc507a5e91e4cea02a91R69

ahh yeah.  Sorry, missed that.

tpantelis (approved):

vthapar (approved):

PR #1929, opened 2025-04-24
Bump helm to 3.17.3 in /tools for CVE-2025-32387

dfarrell07:
Change generated by:

    # in /tools
    go get helm.sh/helm/v3@v3.17.3
    go mod tidy


dfarrell07:
After discussions (https://github.com/submariner-io/shipyard/pull/1932#issuecomment-2828960171), it seems like we don't want to do these Helm CVE bumps on older release branches since they require updating everything, including all the core K8s components, to the latest versions.

It's wild that Helm doesn't do their versioning correctly while still being such an in-demand project.

Closing all the bump older than 0.20, as that's the last version that only requires a K8s minor version update for the bump.

PR #1932, opened 2025-04-24
Bump helm to 3.17.3 in /tools for CVE-2025-32387

dfarrell07:
Change generated by:

    # in /tools
    go get helm.sh/helm/v3@v3.17.3
    go mod tidy


tpantelis:
This bumps K8s to v0.32.2 and bumps kustomize. I don't even know how helm is used. Maybe some build tooling does. It's probably not necessary to address this CVE.

dfarrell07:
> I know we typically don't want to do bumps like this on older release branches, especially for a likely-irrelevant Helm CVE, but I also know there's some complexity around needing to maintain some of these old branches for a long time. In some sense, I wonder if we should be doing more of these big dep updates on older branches where possible. It would be good to clarify what's too much of a bump vs what we should be doing for long-term support.

I don't have a clear answer there - largely depends on what the dependency is.

Helm is used in E2E when helm deploy is requested, which is only done by the submariner-charts repo. So the CVE is of no concern. We can bump it in devel and release-0.20 where the transitive dependencies are compatible anyway but I don't think we should bump helm in older branches.

dfarrell07:
If we don't want this bump on the older branches, which branch should we stop at?

tpantelis:
> If we don't want this bump on the older branches, which branch should we stop at?

I'd say release-0.20. Looking at the PR https://github.com/submariner-io/shipyard/pull/1928/files, it only bumps the K8s minor version.

dfarrell07:
> > If we don't want this bump on the older branches, which branch should we stop at?
> 
> I'd say release-0.20. Looking at the PR https://github.com/submariner-io/shipyard/pull/1928/files, it only bumps the K8s minor version.

That makes sense, and also makes sense more generally (try to keep significant dep version bumps to minor versions). Will do, thanks!

dfarrell07:
After discussions (https://github.com/submariner-io/shipyard/pull/1932#issuecomment-2828960171), it seems like we don't want to do these Helm CVE bumps on older release branches since they require updating everything, including all the core K8s components, to the latest versions.

It's wild that Helm doesn't do their versioning correctly while still being such an in-demand project.

Closing all the bump older than 0.20, as that's the last version that only requires a K8s minor version update for the bump.

PR #1989, opened 2025-06-02
Switch to Fedora packages of buildx and helm

skitt:
docker-buildx and helm are now packaged in Fedora, rely on that instead of installing from upstream using dependencies in the tools modules. This should reduce the churn in tools significantly, and avoid headaches involving Helm Go dependencies.


skitt:
A little present for you @dfarrell07!

dfarrell07:
Wow, that's huge! Thanks @skitt! 

BTW, for my learning, how did you discover this is now an option?

dfarrell07 (approved):

skitt:
> BTW, for my learning, how did you discover this is now an option?

I upgraded to Fedora 42 on my laptop and noticed that `docker-buildx` was installed during the upgrade :wink:. Then I checked the various tools used in Shipyard to see which ones had been packaged since I last checked!

maayanf24 (approved):

tpantelis (approved):

PR #1990, opened 2025-06-04
Configure IPv6 addresses for kind OVNK

yboaron:
This commit configures IPv6 addressing support for kind OVNK deployments
in ipv6,dual-stack mode.

Note: for local testing, it may be necessary to add
`--cap-add=NET_ADMIN --privileged=true` to the Dockerfile at [1].
This is not part of this PR, but may be useful for local testings.

[1]
https://github.com/submariner-io/shipyard/blob/devel/package/Dockerfile.shipyard-dapper-base#L7

### DIFF
--- a/scripts/shared/lib/clusters_kind
+++ b/scripts/shared/lib/clusters_kind
@@ -199,12 +199,32 @@ EOF
 
 function deploy_kind_ovn(){
     export K8S_VERSION
-    export NET_CIDR_IPV4="${cluster_CIDRs[${cluster}]}"
-    export SVC_CIDR_IPV4="${service_CIDRs[${cluster}]}"
+    export SVC_CIDR_IPV4
+    export NET_CIDR_IPV4
+    export SVC_CIDR_IPV6
+    export NET_CIDR_IPV6
     export KIND_CLUSTER_NAME="${cluster}"
 
     local ovn_flags=()
     [[ "$OVN_IC" != true ]] || ovn_flags=( -ic -npz 1 -wk 3 )
+
+    if [[ "$IPV6_STACK" ]]; then
+        ovn_flags+=( -n4 -i6 -sw)
+        SVC_CIDR_IPV6="${service_IPv6_CIDRs[${cluster}]}"
+        NET_CIDR_IPV6="${cluster_IPv6_CIDRs[${cluster}]}"
+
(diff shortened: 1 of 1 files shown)


skitt (changes requested):
The commit message suggests that the extra privilege arguments should be part of this PR; I take it that’s not the case though, and that the note is instructions for people wanting to do the same thing locally. I suggest

> Note: to enable local deployment of kind with OVNK in dual-stack mode, you need to add --cap-add=NET_ADMIN --privileged=true to [1].

How will this work in CI?

yboaron:
> The commit message suggests that the extra privilege arguments should be part of this PR; I take it that’s not the case though, and that the note is instructions for people wanting to do the same thing locally. I suggest

Correct, do you think that commit message should be updated ?

> 
> > Note: to enable local deployment of kind with OVNK in dual-stack mode, you need to add --cap-add=NET_ADMIN --privileged=true to [1].
> 
> How will this work in CI?

In case we need to enable IPv6 (it worked out of the box for kind+kindnet for intra-cluster and inter-cluster traffic) in CI configuration I thought maybe to adopt your [idea](https://github.com/submariner-io/shipyard/pull/1813/files#diff-7dc87d4394e1756c519dbfd0b80d3b31377f643f0bc25d3ed807ce8a3794023dR17)

skitt:
> > The commit message suggests that the extra privilege arguments should be part of this PR; I take it that’s not the case though, and that the note is instructions for people wanting to do the same thing locally. I suggest
> 
> Correct, do you think that commit message should be updated ?

Yes please, that was the purpose of my suggestion :wink::

> Note: to enable local deployment of kind with OVNK in dual-stack mode, you need to add --cap-add=NET_ADMIN --privileged=true to [1].

> > How will this work in CI?
> 
> In case we need to enable IPv6 (it worked out of the box for kind+kindnet for intra-cluster and inter-cluster traffic) in CI configuration I thought maybe to adopt your [idea](https://github.com/submariner-io/shipyard/pull/1813/files#diff-7dc87d4394e1756c519dbfd0b80d3b31377f643f0bc25d3ed807ce8a3794023dR17)

OK, thanks!

skitt (approved):

skitt:
Thanks!

aswinsuryan (approved):

PR #1991, opened 2025-06-06
Set IPv6 forwarding to enable kind OVNK dual-stack deployment

yboaron:
Kind OVNK dual-stack and IPv6 deployments require enabling IPv6 forwarding by setting `net.ipv6.conf.all.forwarding=1`.

Currently this setting cannot be applied and results in a `sysctl: permission denied on key "net.ipv6.conf.all.forwarding"` error.

This PR updates .dapper file to set IPv6 sysctl on the container at startup.


skitt:
Adding the `--sysctl` line to `.dapper` directly was just a PoC :wink:. I think the appropriate fix is #1992.

PR #1992, opened 2025-06-06
Enable IPv6 forwarding by default

skitt:
This is required for OVN-K using IPv6. The option must be set using sysctl inside the Dapper container's namespace; to do this without requiring privileged container invocations, this uses Docker's --sysctl option.


dfarrell07 (approved):

yboaron (approved):

yboaron:
cc @vthapar @tpantelis @sridhargaddam

tpantelis (approved):

PR #2010, opened 2025-06-16
Use a single ncat for IPv4 and IPv6, or IPv6 only

skitt:
ncat without IP limitations (-4 or -6) uses both v4 and v6, so ncat
listening on port 8080 binds to the port on IPv4 and IPv6 addresses.
Starting a second ncat for IPv6 only then fails because the port is
already taken.

This tries an unspecified connection first, listening on IPv4 and, if
present, IPv6; if that fails, it uses an IPv6-only connection. It also
drops --keep-open, the tests rely on ncat restarting after every
connection.

### DIFF
--- a/scripts/nettest/simpleserver
+++ b/scripts/nettest/simpleserver
@@ -1,8 +1,9 @@
 #!/bin/bash
 set -e
 
+# By default, ncat fails if it can't use an IPv4 connection; this tries an unspecified connection first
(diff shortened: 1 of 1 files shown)


skitt:
Does this seem OK to you @yboaron?

skitt:
Ah, this fails in IPv6-only deployments.

vthapar (approved):

yboaron (approved):

PR #2102, opened 2025-09-01
Add Skip NonGW dataplane tests flag

yboaron:
With the support for decoupling intra-cluster and inter-cluster routing components, we want to give users control over which dataplane tests are executed.

By default, all tests will run, but users should be able to skip NonGW dataplane connectivity tests by setting flag in subctl verify.


test/e2e/framework/test_context.go
```diff
@@ -45,6 +45,7 @@ type TestContextType struct {
 	OperationTimeout        uint
 	PacketSize              uint
 	SkipConnectorSrcIPCheck bool
+	SkipNonGWDataplaneTests bool
```
tpantelis: How would this be used? Also the naming seems specific to the submariner project. Can we generalize it to `IntraClusterRoutingDisabled` for consistency consistent with other components?

tpantelis (changes requested):

test/e2e/framework/test_context.go
tpantelis: I was thinking we would dynamically determine this by retrieving the RA daemonset and inspecting the `SUBMARINER_INTRAROUTINGDISABLED` env var. That way when users run `subctl verify` we would skip the appropriate tests.

In addition, more broadly, I was thinking we'd add a `--disable-intra-cluster-routing` CLI flag to `subctl join` and a corresponding shipyard `Makefile` flag. Then add more jobs to the GHA E2E test matrix.

yboaron: @tpantelis  

I was thinking of controlling the E2E datapath tests via a flag on the **subctl verify** command.

The reason is that even when IntraClusterRoutingDisabled is set, some other SDN component may still handle routing traffic towards the gateway. 
In that case, it could still be valuable to run all datapath tests, including those involving non-GW nodes.

If a user isn’t interested in verifying the non-GW datapath, they can explicitly set the flag in subctl verify to skip those tests. 

This approach gives users flexibility while still allowing us to catch issues in cases where intra-cluster routing is handled outside of Submariner.

Regarding the naming, do you think **IntraClusterRoutingDisabled** would be more appropriate than **SkipNonGWDataplaneTests**? in that case we can add **--disable-intra-cluster-routing** CLI flag to subctl join and subctl verify

tpantelis: In that case, I think `--disable-intra-cluster-routing` CLI flag for `subctl join`. For `subctl verify`, how about `--skip-intra-cluster-routing-tests` ...

yboaron: +1 for **--skip-intra-cluster-routing-tests**

dfarrell07 (approved):

tpantelis (approved):

PR #2118, opened 2025-09-03
Add intra-cluster disablement support for deployment

tpantelis:
Add `intra-cluster-disabled` Makefile `using` flag for deploying Submariner and for skipping intra-cluster connectivity E2E tests.


.github/workflows/testing.yml
```diff
@@ -125,6 +125,7 @@ jobs:
             globalnet: globalnet
           - extra-toggles: ipv6-stack
           - extra-toggles: ovn
+          - extra-toggles: 'ovn, intra-cluster-disabled'
```
yboaron, line 128: Since Lighthouse shouldn’t be impacted by intra-cluster routing, perhaps we should include it here as well?

yboaron (approved):

.github/workflows/testing.yml
tpantelis: I don't think it's necessary here.

aswinsuryan (approved):

PR #2147, opened 2025-09-30
Automated backport of #2130: Build golangci-lint with the installed Go compiler

skitt:
Backport of #2130 on release-0.19.

#2130: Build golangci-lint with the installed Go compiler

For details on the backport process, see the [backport requests](https://submariner.io/development/backports/) page.

### DIFF
--- a/package/Dockerfile.shipyard-dapper-base
+++ b/package/Dockerfile.shipyard-dapper-base
@@ -57,9 +57,12 @@ RUN dnf -y install --nodocs --setopt=install_weak_deps=False \
 
 COPY tools/go.mod /tools.mod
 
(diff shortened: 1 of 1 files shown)


tpantelis:
Do we really need this backported all the way down?

skitt:
> Do we really need this backported all the way down?

I reckoned it might be worth it because currently the job fails by running out of memory (see https://github.com/submariner-io/shipyard/actions/runs/17884436578/job/50856290443) which takes a while. Backporting this everywhere reduces the amount of GHA minutes consumed and might even allow the linting jobs to run (although we’re seeing the old ginkgo failures now).

tpantelis (approved):

PR #2280, opened 2026-02-04
Run tests in parallel

skitt:
This uses the Ginkgo driver directly to allow running the tests in parallel.

## Summary by CodeRabbit

* **Chores**
  * Upgraded testing infrastructure to use Ginkgo v2 runner for more efficient test execution and improved coverage profiling.


tpantelis:
What is the motivation for running unit tests is parallel, especially at this point? This will likely cause issues in some test suites that modify shared globals.

skitt:
> What is the motivation for running unit tests is parallel, especially at this point? This will likely cause issues in some test suites that modify shared globals.

This is just a draft to measure the impact. The motivation is to reduce the time spent waiting for tests (but it would be more useful on e2e).

PR #2487, opened 2026-07-09
Fix CVEs for release-0.22

maayanf24:
## Summary by CodeRabbit

* **Chores**
  * Updated several Go module dependencies to newer versions.
  * Brought the tooling dependency set in line with the latest compatible releases.

* **Bug Fixes**
  * Added a vulnerability exemption for a low-severity Go package issue to avoid unnecessary breaking changes while keeping builds moving smoothly.


skitt (approved):
GO-2026-5024 is indeed insignificant for Submariner, but stdlib vulnerabilities may require bumping the Go compiler (if not the `go.mod` Go language level) at some point.

PR #2512, opened 2026-08-19
Harden Dockerfiles (FIND-005/008/012)

dfarrell07:
Dockerfile hardening covering three findings:

- **FIND-005** — digest-pin `package/Dockerfile.shipyard-dapper-base` (fedora:43), `package/Dockerfile.shipyard-linting` (alpine:3.24), and `package/Dockerfile.nettest.konflux` (ubi9 multi-arch index digest; `--platform` kept, `:latest` dropped). `Dockerfile.nettest` is intentionally not pinned — its `FEDORA_VERSION` ARG also drives `dnf_install` repo selection.
- **FIND-012** — add `USER 10000` to `package/Dockerfile.nettest` and `package/Dockerfile.nettest.konflux` (no `adduser` needed).
- **FIND-008** — switch `scripts/nettest/metricsproxy.konflux` ncat from `-c` (`--sh-exec`) to `-e` (`--exec`), removing the second-shell re-parse.

## Summary by CodeRabbit

* **Security**
  * Network testing containers now run as an unprivileged user rather than root.
  * Metrics proxy commands execute directly without shell interpretation.

* **Reliability**
  * Container base images are pinned to specific versions and immutable digests, providing more consistent and reproducible environments.


package/Dockerfile.nettest.konflux
```diff
@@ -1,6 +1,6 @@
 ARG BASE_BRANCH=release-0.22
 
-FROM --platform=${TARGETPLATFORM} registry.redhat.io/ubi9:latest
+FROM --platform=${TARGETPLATFORM} registry.redhat.io/ubi9@sha256:5426a8f45e80a07168a30ea24d84f266094b3756624a5508cc53927e6ee39e09
```
skitt, line 3: The justification given for this is reproducibility, but our images aren’t reproducible anyway since they run `dnf update` and install additional packages (which will be whatever package is available at the time the image is built). There _is_ however a supply-chain benefit to pinning images: it ensures the repository configuration in the base image remains constant and therefore reliable (assuming it’s been verified…). This means that our supply-chain exposure is limited to the already-known repositories; without the pin an attacker with access to the container image repository could replace the repository configuration in the base image to point to an attacker-controlled repository. (Although if an attacker controls base images the game is lost already.)

Would you mind updating the commit message accordingly?

skitt (changes requested):

package/Dockerfile.nettest.konflux
dfarrell07: Reworded. The reproducibility phrase you flagged was actually in the FIND-005 commit (Pin base image digests...), not this one, since that's where the framing originally came from. Dropped the reproducibility justification there and credited the real reason instead: pinning locks the base image's repo config, so an attacker who compromises the registry can't swap in a malicious repo pointing elsewhere. Pushed.

skitt (approved):

PR #2516, opened 2026-08-19
Harden Dockerfiles (FIND-005/012)

dfarrell07:
Dockerfile hardening covering two findings:

- **FIND-005** — digest-pin `package/Dockerfile.shipyard-dapper-base` (fedora:41) and `package/Dockerfile.shipyard-linting` (alpine:3.24). `Dockerfile.nettest` is intentionally not pinned — its `FEDORA_VERSION` ARG also drives `dnf_install` repo selection.
- **FIND-012** — add `USER 10000` to `package/Dockerfile.nettest` (after the version `RUN`, before `CMD`) so the community image defaults to non-root off-OpenShift.

Konflux files do not exist on release-0.18; FIND-019 Dependabot docker is devel-only.

### DIFF
--- a/package/Dockerfile.nettest
+++ b/package/Dockerfile.nettest
@@ -22,4 +22,6 @@ COPY scripts/nettest/* /app/
 
 RUN echo ${VERSION} >> /app/version
 
+USER 10000
+
 CMD ["/bin/bash","-l"]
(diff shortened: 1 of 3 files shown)


skitt (changes requested):
See #2512

dfarrell07:
Same fix as #2512 (commit message reworded to drop the reproducibility framing, credit the actual supply-chain rationale instead). See that PR for the full discussion. Pushed here too.

skitt (approved):

PR #2534, opened 2026-08-19
Harden e2e test-pod security posture (FIND-009/010/011/017)

dfarrell07:
Test-pod security hardening covering four findings:

- **FIND-017** — add pod/container `securityContext`, `automountServiceAccountToken: false`, and CPU/memory requests+limits to the `dummypod`, `netshoot`, and `nginx-demo` CI helper manifests.
- **FIND-009** — disable service-account token automount on all 7 `NetworkPod` PodSpecs in `test/e2e/framework/network_pods.go`.
- **FIND-010** — add `readOnlyRootFilesystem: true` to the shared container `SecurityContext`; `CustomPod` callers that need a writable root can set `WritableRootFS: true`.
- **FIND-011** — differentiated resource profiles: hard-capped for TCP listener/connector/CustomPod; requests-only (no CPU limit) for throughput/latency pods where a hard CPU cap corrupts benchmark numbers.

**Note:** apply the `e2e-projects` label so the consuming-repo e2e gate exercises the framework changes.

## Summary by CodeRabbit

* **Security Enhancements**
  * Hardened demo and test workloads with non-root execution, restricted privileges, read-only filesystems, dropped capabilities, RuntimeDefault seccomp, and disabled service-account token mounting.
  * Added default CPU and memory requests and limits.

* **Testing**
  * Added resource profiles for lightweight and throughput-focused network tests.
  * Enabled custom test pods to use writable filesystems and override resource requirements.
  * Updated Kubernetes test configuration handling for compatibility.


skitt (changes requested):
See #2535 (but using `new` instead of `ptr.To` since we have Go 1.26 here).

dfarrell07:
Applied in af42da0b, using `new(...)` per Go 1.26 instead of `ptr.To(...)` since this branch never had the `k8s.io/utils/ptr` import to begin with.

skitt (approved):

PR #2542, opened 2026-08-19

dfarrell07:

## Summary by CodeRabbit

* **Security**
  * Release images can now be signed and verified using Cosign and Sigstore.
  * Image signing supports registry authentication and records the published image digest.
  * Signing failures can be handled without blocking the release process.
  * Release workflows now support secure OIDC-based authentication for signing operations, improving provenance and verification of published images.


dfarrell07:
This is a known caveat documented in the finding: co-located checksums from the same GitHub release don't protect against a compromised release origin — an attacker with `contents: write` on the repo controls both the binary and the checksum file. This is explicitly acknowledged in the FIND-015 caveat section.

`cosign verify-blob` with the `.sig`/`.pem` files would provide an independent Sigstore/Rekor trust root, but requires cosign to already be present on the system — a chicken-and-egg problem for a bootstrap install step. The `sha256sum` approach is standard practice for cosign bootstrap installs and is what the cosign project itself recommends for this scenario.

The current approach eliminates `get.submariner.io` as a vector and adds forensic value (Rekor transparency entries on signed images) while accepting the residual GitHub release origin trust. Hardening the bootstrap itself (e.g. a pre-installed cosign via a trusted action on the runner, outside Dapper) can be tracked as a follow-up if desired.

skitt (dismissed):
See #2543

dfarrell07:
Same fix as #2543 (TUF-based cosign verification, replacing the same-source checksum). See that PR for the full discussion. Pushed here too.

Also addresses the CodeRabbit finding on this PR (same underlying issue) - it already self-marked as addressed against the pushed commit range.

tpantelis:
@coderabbitai approve

tpantelis (approved):

skitt (approved):

PR #2543, opened 2026-08-19

dfarrell07:

## Summary by CodeRabbit

- **Security**
  - Release container images are now signed using Sigstore Cosign, enabling signature verification after publication.
  - Cosign is securely verified and installed during the image-building process.
  - Release workflows now support the permissions and credentials required for keyless signing.
  - Checkout credentials are no longer retained after use.

- **Reliability**
  - Image signing and temporary release resources are cleaned up automatically.
  - Signing or authentication issues produce warnings without preventing the broader release process from completing.


package/Dockerfile.shipyard-dapper-base
```diff
@@ -95,6 +95,14 @@ RUN go -C /tools install github.com/golangci/golangci-lint/v2/cmd/golangci-lint
     find /go/bin -type f -executable -newercm /go \( -execdir upx ${UPX_LEVEL} {} \; -o -true \) && \
     go clean -cache -modcache && rm -rf /tools
 
+# cosign for signing released community images (not packaged in Fedora dnf).
+ARG COSIGN_VERSION=v3.1.3
+RUN curl -sSfL "https://github.com/sigstore/cosign/releases/download/${COSIGN_VERSION}/cosign-linux-${ARCH}" -o /usr/local/bin/cosign && \
+    curl -sSfL "https://github.com/sigstore/cosign/releases/download/${COSIGN_VERSION}/cosign_checksums.txt" -o /tmp/cosign_checksums.txt && \
+    grep "cosign-linux-${ARCH}$" /tmp/cosign_checksums.txt | sed "s#cosign-linux-${ARCH}#/usr/local/bin/cosign#" | sha256sum -c - && \
```
skitt, line 102: This downloads the checksum and binary from the same source, which means the checksum only protects against undetected partial downloads, and those should result in a `curl` error anyway.

The correct validation process is described [in the cosign installation guide](https://docs.sigstore.dev/cosign/system_config/installation/#verifying-cosign-releases).

Note that if we trust the Go compiler, then adding `cosign` as a Go dependency would provide source-level authentication (using `go.sum`) and allow any fixes in the Go compiler to be applied without waiting for a new `cosign` release.

skitt (changes requested):

package/Dockerfile.shipyard-dapper-base
dfarrell07: Went with the first option (cosign's documented TUF-based artifact-key verification), not the go.sum route. Prototyped that one first, but cosign pulls a huge dependency tree (AWS/Azure/GCP/Alibaba KMS clients) and forces this branch's Go toolchain from 1.25 to 1.26, which isn't safe on a stable release branch.

The new step: fetches Sigstore's TUF root (pinned by hash, since it's an immutable root_history/ entry), uses it to get the real artifact public key from tuf-repo-cdn.sigstore.dev independent of the GitHub release, verifies the release's -kms.sigstore.json bundle against the binary with openssl dgst -verify, then runs cosign verify-blob --bundle per the docs to check Rekor log-inclusion. Diffed it line-by-line against the current docs page and it's a faithful match. Tested with a real docker build --no-cache on all 8 branches, not just this one. Pushed.

skitt (approved):

PR #2544, opened 2026-08-19

dfarrell07:


skitt (changes requested):
See #2543

dfarrell07:
Same fix as #2543 (TUF-based cosign verification, replacing the same-source checksum). See that PR for the full discussion. Pushed here too.

skitt (approved):

PR #2550, opened 2026-08-19
Print debug trace commands without eval

dfarrell07:
debug_functions ran `eval echo "$BASH_COMMAND"` to expand each traced
command before printing it. With DEBUG_PRINT=true that expanded any
variable in the command, leaking $QUAY_USERNAME (and any other inherited
secret) into the release trace, and re-executing command substitutions.

Print $BASH_COMMAND verbatim instead: no expansion, no re-execution.
Ordinary commands are still traced; only literal $VAR tokens differ.

Follow-up to FIND-007 (the render_template eval, PR 7); different file
and mechanism, tracked separately.

## Summary by CodeRabbit

- **Bug Fixes**
  - Debug traces now print commands verbatim without expanding embedded variables.
  - Sensitive values stored in shell variables are no longer exposed in debug output.


skitt (changes requested):
The change is OK but the explanation mentions command substitutions, which were already avoided (at least, those using `$()` and not backticks, but our shell linting is supposed to prevent the latter).

Please fix the commit message.

dfarrell07:
Fair point, the commit message overstated it. Reworded to drop the re-executing command substitutions framing. The actual justification is the eval echo "$BASH_COMMAND" leaking secrets like $QUAY_USERNAME via variable expansion, which is what's fixed. Pushed.

skitt (approved):

PR #2551, opened 2026-08-19
Print debug trace commands without eval

dfarrell07:
debug_functions ran `eval echo "$BASH_COMMAND"` to expand each traced
command before printing it. With DEBUG_PRINT=true that expanded any
variable in the command, leaking $QUAY_USERNAME (and any other inherited
secret) into the release trace, and re-executing command substitutions.

Print $BASH_COMMAND verbatim instead: no expansion, no re-execution.
Ordinary commands are still traced; only literal $VAR tokens differ.

Follow-up to FIND-007 (the render_template eval, PR 7); different file
and mechanism, tracked separately.


skitt (changes requested):
See #2550

dfarrell07:
Same fix as #2550 (commit message reworded). See that PR for the full discussion. Pushed here too.

skitt (approved):

PR #2558, opened 2026-08-19
Replace eval-based rendering with envsubst

dfarrell07:
render_template ran `eval` on template file contents, executing any shell in a
template or in a substituted value (CWE-95). Replace it with envsubst restricted
to an explicit allow-list, run in a subshell so only those vars are exported —
inherited Dapper secrets can never reach rendered output.

envsubst substitutes only scalar env vars, so install_subm pre-resolves CIDR
associative-array subscripts to scalars and bundle/submariner.yaml is updated to
match. Other rendered templates (bundle/broker.yaml, the kind configs, and where
present sm-global-cm.yaml) also drop eval-only quote escaping so envsubst output
stays byte/YAML-equivalent to the old eval rendering for all templates.

**Note:** render_template is baked into shipyard-dapper-base; effective only after
this branch's image is rebuilt and republished (gettext-envsubst provides envsubst).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved deployment template rendering so network and cluster settings are populated reliably.
  * Prevented unrelated environment secrets from being unintentionally included in generated configuration.
  * Corrected registry and YAML quoting across cluster configurations.
  * Fixed configuration substitutions for broker, Submariner, namespace, image, and version settings.

* **Chores**
  * Added environment-substitution support to the deployment image.


skitt (changes requested):
See #2565

dfarrell07:
Same fix as #2565 (commit message reworded, envsubst line simplified). See that PR for the full discussion. Pushed here too.

tpantelis (approved):

tpantelis:
@coderabbitai approve

skitt (approved):

PR #2565, opened 2026-08-19
Replace eval-based rendering with envsubst

dfarrell07:
render_template ran `eval` on template file contents, executing any shell in a
template or in a substituted value (CWE-95). Replace it with envsubst restricted
to an explicit allow-list, run in a subshell so only those vars are exported —
inherited Dapper secrets can never reach rendered output.

envsubst substitutes only scalar env vars, so install_subm pre-resolves CIDR
associative-array subscripts to scalars and bundle/submariner.yaml is updated to
match. Other rendered templates (bundle/broker.yaml, the kind configs) also drop
eval-only quote escaping so envsubst output stays byte/YAML-equivalent to the old
eval rendering for all templates.

**Note:** render_template is baked into shipyard-dapper-base; effective only after
this branch's image is rebuilt and republished (gettext-envsubst provides envsubst).

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected CIDR values in generated Submariner configurations.
  - Fixed rendering of broker component lists and registry settings.
  - Improved template substitution so required configuration values expand reliably.
  - Prevented sensitive environment variables from appearing in generated configuration files.

- **Improvements**
  - Added environment-variable substitution support to the deployment image.
  - Simplified generated YAML and TOML formatting without changing intended behavior.


scripts/shared/lib/utils
```diff
@@ -52,7 +52,26 @@ function render_template() {
...
+        cluster components cs iib ns og sub bundle mc master_ip
+        pod_cidr service_cidr pod_cidr_ipv6 service_cidr_ipv6
+        dns_domain disable_cni kube_proxy_mode nodes
+        service_CIDR cluster_CIDR global_CIDR clusterset_ip_CIDR
+    )
+    (   # subshell: export without polluting parent; envsubst only reads env vars
+        export "${NAMES[@]}" 2>/dev/null
+        # shellcheck disable=SC2016  # literal $NAME tokens are envsubst's SHELL-FORMAT, not shell vars
+        envsubst "$(printf '$%s ' "${NAMES[@]}")" < "$1"
+    )
```
skitt, line 74: The commit message says “run in a subshell so only those vars are exported” but this doesn’t achieve that; anything exported in the current shell is visible in the subshell.

The usual way of handling this is `env -i` but that then requires listing all the variables with their values; `env -i "SUBM_NS=$SUBM_NS"` etc. but that quickly gets convoluted in this sort of use case. If the reliance on `envsubst` is sufficient, then removing the claim from the commit message would be acceptable (the subshell is still useful then to avoid polluting the current shell).

scripts/shared/lib/utils
```diff
@@ -52,7 +52,26 @@ function render_template() {
...
+        LIGHTHOUSE OVERLAPPING MARKETPLACE_NAMESPACE USE_CLUSTERSET_IP NFTABLES
+        cluster components cs iib ns og sub bundle mc master_ip
+        pod_cidr service_cidr pod_cidr_ipv6 service_cidr_ipv6
+        dns_domain disable_cni kube_proxy_mode nodes
+        service_CIDR cluster_CIDR global_CIDR clusterset_ip_CIDR
+    )
+    (   # subshell: export without polluting parent; envsubst only reads env vars
+        export "${NAMES[@]}" 2>/dev/null
+        # shellcheck disable=SC2016  # literal $NAME tokens are envsubst's SHELL-FORMAT, not shell vars
+        envsubst "$(printf '$%s ' "${NAMES[@]}")" < "$1"
```
skitt, line 73: The command substitution isn’t necessary:

```suggestion
        envsubst "${NAMES[@]/#/$}" < "$1"
```

skitt (changes requested):

scripts/shared/lib/utils
dfarrell07: Applied as suggested. Also dropped the now-unneeded shellcheck disable=SC2016 comment above it, since the new form doesn't trigger that warning. Pushed.

scripts/shared/lib/utils
dfarrell07: Good catch. Reworded the commit message to drop the claim that the subshell scopes what's exported. The safety property really comes from envsubst's SHELL-FORMAT allow-list restricting substitution to the listed names, not the subshell. The subshell is now described only as avoiding polluting the parent shell. Pushed.

scripts/shared/lib/utils
skitt: Sorry, the correct fix is

```
envsubst "${NAMES[*]/#/$}" < "$1"
```

tpantelis:
@coderabbitai approve

tpantelis (approved):

skitt (approved):

PR #2511, opened 2026-08-19
Harden Dockerfiles (FIND-005/008/012)

dfarrell07:
Dockerfile hardening covering three findings:

- **FIND-005** — digest-pin `package/Dockerfile.shipyard-dapper-base` (fedora:43), `package/Dockerfile.shipyard-linting` (alpine:3.24), and `package/Dockerfile.nettest.konflux` (ubi9 multi-arch index digest; `--platform` kept, `:latest` dropped). `Dockerfile.nettest` is intentionally not pinned — its `FEDORA_VERSION` ARG also drives `dnf_install` repo selection.
- **FIND-012** — add `USER 10000` to `package/Dockerfile.nettest` and `package/Dockerfile.nettest.konflux` (no `adduser` needed).
- **FIND-008** — switch `scripts/nettest/metricsproxy.konflux` ncat from `-c` (`--sh-exec`) to `-e` (`--exec`), removing the second-shell re-parse.

See commit messages for details.

## Summary by CodeRabbit

* **Security**
  * Containers now run with a non-root user, reducing potential security risks.
  * Container base images are pinned to verified versions for improved reproducibility and supply-chain security.

* **Bug Fixes**
  * Updated network test proxy command execution to improve connection handling.

PR #2619, opened 2026-09-16
Fix CI helper pod manifests to work on OCP

dfarrell07:
## Problem

The glasswing FIND-017 hardening added a pod-level `securityContext` with
`runAsUser: 10000` to `netshoot.yaml`, `nginx-demo.yaml`, and `dummypod.yaml`.
This breaks pod admission on OCP with two SCC rejections:

```
provider "restricted-v2": .containers[0].runAsUser: Invalid value: 10000:
  must be in the ranges: [1001530000, 1001539999]
provider "restricted-v3": .spec.securityContext.hostUsers: Invalid value: "null":
  Host Users must be set to false
```

OCP's `restricted-v2` SCC uses `MustRunAsRange` — any explicit `runAsUser`
must fall within the namespace's allocated UID range. 10000 is never in that
range. OCP 4.20+ `restricted-v3` requires `hostUsers: false` whenever a
pod-level `securityContext` block is present.

Before FIND-017, these manifests had no pod-level `securityContext`, so OCP
injected a valid namespace-range UID automatically. Upstream CI uses kind
only and never caught this.

## Fix

Remove `runAsUser: 10000` and add `hostUsers: false`.

Without an explicit `runAsUser`:
- **OCP**: injects a UID from `openshift.io/sa.scc.uid-range` (always in range, always non-zero → satisfies `runAsNonRoot`)
- **kind**: `Dockerfile USER 10000` (from FIND-012/PR #2511) applies — still runs as non-root 10000

`hostUsers: false` satisfies `restricted-v3` on OCP 4.20+. All other glasswing hardening (`runAsNonRoot`, `seccompProfile`, `automountServiceAccountToken: false`, `readOnlyRootFilesystem`, capability drops) is preserved unchanged.

Backport of #2618.

https://claude.ai/code/session_01G5eahW1Ke5qZ3LMFTAPfJf


skitt (approved):

scripts/shared/resources/dummypod.yaml
```diff
@@ -16,7 +16,7 @@ spec:
       automountServiceAccountToken: false
       securityContext:
         runAsNonRoot: true
-        runAsUser: 10000
+        hostUsers: false
```
skitt: This needs to go in the `PodSpec`, not the security context.

